PR #232, opened 2016-03-17
Updated help navigation #214

### LINKED ISSUE #214: optimize help navigation integration
Help is accessible through the sidebar menu. 
but Help is a new activity which closes the main application with no history back. 

basically when the user is in the help section, using the device back buttons should go back to the main application an using the home button should toggle the sidebar.

aferditamuriqi:
fixes #214


simplified-app-shared/src/main/res/values/strings.xml
```diff
@@ -55,7 +56,7 @@
   <string name="catalog_book_sync">Sync</string>
   <string name="catalog_book_download">Download</string>
   <string name="catalog_book_delete">Delete</string>
-  <string name="catalog_book_revoke_loan">Return</string>
+  <string name="catalog_book_revoke_loan">Return now</string>
```
winniequinn: @aferditamuriqi: Should these strings be titlecase (e.g. "Return Now")?

simplified-app-shared/src/main/res/values/strings.xml
```diff
@@ -43,6 +43,7 @@
   <string name="catalog_refresh">Refresh Feed</string>
   <string name="catalog_search">Search</string>
   <string name="catalog_unknown_feed">Unknown</string>
+  <string name="catalog_more_feed">More...</string>
```
winniequinn: @aferditamuriqi: We should use the actual ellipsis character in such strings ("…"), not three periods ("...").

PR #278, opened 2016-06-07
improved search activity

### LINKED ISSUE #204: Limit search Activity to one instance
Currently every search adds an activity to the stack. 
Search Activity should be reused.

aferditamuriqi:
- fixed search activity, to not keep multiple instances.
- clear search view when back pressed
- keep search view always on action bar

fixes #204 

@winniequinn

### DIFF
--- a/simplified-app-shared/src/main/java/org/nypl/simplified/app/catalog/CatalogFeedActivity.java
+++ b/simplified-app-shared/src/main/java/org/nypl/simplified/app/catalog/CatalogFeedActivity.java
@@ -116,6 +116,7 @@ public abstract class CatalogFeedActivity extends CatalogActivity implements
   private @Nullable ViewGroup    progress_layout;
   private           int          saved_scroll_pos;
   private           boolean      previously_paused;
+  private           SearchView   search_view;
 
   /**
    * Construct an activity.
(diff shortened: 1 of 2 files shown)


simplified-app-shared/src/main/java/org/nypl/simplified/app/catalog/CatalogFeedActivity.java
```diff
@@ -678,18 +687,21 @@ private void onCreateOptionsMenuSearchItem(
...
       /**
        * Set some placeholder text
        */
 
       final CatalogFeedArgumentsType args = this.getArguments();
-      if (this.getUpStack().isEmpty()) {
-        sv.setQueryHint("Search");
-      } else {
-        sv.setQueryHint("Search " + args.getTitle());
+      this.search_view.setQueryHint("Search " + this.feed.getFeedTitle());
```
winniequinn: Should this string be localized?

winniequinn:
@aferditamuriqi Okay except for the one line comment! 👍

simplified-app-shared/src/main/java/org/nypl/simplified/app/catalog/CatalogFeedActivity.java
aferditamuriqi: yes, it could, will add this as part of the next branch where i have added "all" the missing localized strings

Issue #350, opened 2016-10-28
Add missing localized strings


aferditamuriqi:

mattniehoff:
Due to be addressed in iOS as well:  NYPL-Simplified/Simplified-iOS#482

aferditamuriqi:
closed in PR #412

PR #383, opened 2017-03-13
Clarifies the age verification message

### LINKED ISSUE #358: confusing age verification msg
I download new SimplyE, and open the InstantClassics catalog.  I get a dialog that says: "Age Verification
You must be 13 years of age or older to download a book form the collection.  How old are you?"
And then I can chose "under 13" or "13 or older".  Both options are valid, and would show me each its own catalog, so the "must be 13" message is confusing.  Let's rephrase to something like: "You must be 13 or older to download some of the books in this collection.  How old are you?"

mattniehoff:
This states that 'some' of the books aren't able to be downloaded if you're under 13 instead of all.

Fixes #358

### DIFF
--- a/simplified-app-shared/src/main/java/org/nypl/simplified/app/catalog/CatalogFeedActivity.java
+++ b/simplified-app-shared/src/main/java/org/nypl/simplified/app/catalog/CatalogFeedActivity.java
@@ -965,7 +965,7 @@ public abstract class CatalogFeedActivity extends CatalogActivity implements
       alert.setTitle("Age Verification");
 
       // Setting Dialog Message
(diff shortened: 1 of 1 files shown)


winniequinn:
How do you feel about swapping the phrase "certain books" for "some of the books"?

Also, is localization happening in a separate pass or should this string be localized as part of this pull request?

mattniehoff:
I have no objection with the language change. I'm imagining we'd want it to be consistent with what we do on NYPL-Simplified/Simplified-iOS#563 though.

mattniehoff:
There's more localization that needs to be done as part of #350. 

Could tackle it then.

PR #402, opened 2017-04-19
Multi library develop

bluefirepatrick:
This is our pull request for URMS implementation in Simplified for Android.  Inorder to have this work with Adept the ContentModule approach will have to be taken for both Adept and Marlin ContentFilters.  We have successfully created a Marlin ContentModule and I sent the examples of that and the LCP ContentModule earlier so that one can be attempted for Adept.


aferditamuriqi:
referencing https://github.com/NYPL-Simplified/android/issues/164

PR #421, opened 2017-06-13
Add location saving on activity pause

winniequinn:
This fixes an issue where the reader's last location would not be saved if the app was killed while in the background.


winniequinn:
Created against wrong base.

PR #430, opened 2017-07-11
Add Howard County to library list

datalogics-cgreen:
This adds Howard County's library to the library list for the Android app, plus the relevant styles and assets.

As with our other library addition(s), we would like to coordinate the release, so _please do not merge just yet_.

Links such as privacy policy and the license are incoming.


simplified-app-simplye/src/main/assets/Accounts.json
```diff
@@ -58,5 +58,23 @@
     "catalogUrl13": "https://instantclassics.librarysimplified.org/all-books.xml",
     "licenseUrl": "http://www.librarysimplified.org/iclicenses.html",
     "mainColor": "#497049"
+  },
+  {
+    "id": 4,
```
leonardr: Library ID 4 is already taken by the Canadian Electronic Library, in a branch that has not been merged yet.

To avoid future confusion of this type I've established a list of library IDs here: https://github.com/NYPL-Simplified/Simplified/wiki/UniqueLibraryIDs

I've assigned the ID `8` to Howard County library.

datalogics-cgreen: Many thanks @leonardr !

simplified-app-shared/src/main/res/values/styles.xml
```diff
@@ -88,6 +99,18 @@
 
   </style>
 
+  <style
+    name="SimplifiedThemeNoActionBar_hcls"
```
datalogics-cgreen: 'hcls' is not capitalized here. Is this a concern?

gioneill:
Branch assets have been merged in as part of 

PR #431, opened 2017-07-13
Add Montgomery to library list

datalogics-tsmith:
This pull request adds Montgomery County Library to the library list.
### Please do not merge this just yet, we would like to coordinate the release with Montgomery


gioneill:
This branch's assets have been merged in as part of 

PR #433, opened 2017-07-17
Add Frederick to library list

datalogics-cgreen:
This adds Frederick County Public Library to the library list in the Android client, plus its styles and assets.

As with our other additions, we would like to coordinate the releases of these libraries, so **please do not merge yet**.

Links such as the privacy policy and license are incoming.


gioneill:
This branch's assets have been merged in as part of 

PR #445, opened 2017-09-11
New Book Detail UX

gioneill:
Mostly layout changes to promote better text layout, legibility, and image sizes.

develop and gradle-conversion should be merged first.


gioneill:
Wrong branch. Going to re-create PR with the newer book detail branch based on the gradle updated branch.

PR #477, opened 2017-11-02
Add all libraries

datalogics-tarar:


simplified-app-shared/src/main/res/layout/login_dialog.xml
```diff
@@ -76,7 +76,6 @@
         android:layout_weight="1"
         android:inputType="textNoSuggestions"
         android:minWidth="160dp"
-        android:maxLength="4"
```
gioneill: There are new keys in Accounts.json for a library that dictate whether or not a library should be allowing alphanumeric passwords as well as the maximum allowed length (with 0 being unlimited). Refer to the iOS code [here](https://github.com/NYPL-Simplified/Simplified-iOS/blob/340df4266f82fc9ba643034a334db467c2aa4dd0/Simplified/NYPLSettingsAccountDetailViewController.m#L171) and [here](https://github.com/NYPL-Simplified/Simplified-iOS/blob/340df4266f82fc9ba643034a334db467c2aa4dd0/Simplified/NYPLSettingsAccountDetailViewController.m#L1275) and let me know if you need assistance implementing.

simplified-app-simplye/src/main/assets/Accounts.json
```diff
@@ -60,6 +60,130 @@
     "mainColor": "#497049"
   },
   {
+    "id": 7,
```
gioneill: There should be no changes to Accounts.json in this PR. Go ahead and reset the file it to the master branch. Before doing that, ensure that Accounts.json residing in the Simplified-iOS repo is up-to-date in regards to anything that may have been changing. This is to ensure predictability between platforms in our testing. Next time I'll get out in front of this a little better!

gioneill: Sorry I take this comment back. It looks like the branch was just really out of date. Still be sure that Accounts.json on iOS is the way you want it, however. And double check the newer keys for password attributes that I mentioned above.

simplified-books-core/src/main/java/org/nypl/simplified/books/core/BooksControllerRevokeBookTask.java
```diff
@@ -539,6 +558,10 @@ private void revokeLoanedWithDRM(final URI revoke_uri)
         if (in_error.startsWith("E_ACT_NOT_READY")) {
           this.error = Option.some((Throwable) new AccountNotReadyException(in_error));
         }
+        else if (in_error.startsWith("E_STREAM_ERROR")) {
```
gioneill: This has not yet been approved in a separate PR I have in to Master, so remove this change for now.

simplified-app-shared/src/main/java/org/nypl/simplified/app/SimplifiedActivity.java
```diff
@@ -251,12 +251,36 @@ private void startSideBarActivity() {
       setTheme(R.style.SimplifiedTheme_NYPL);
     } else if (id == 1) {
       setTheme(R.style.SimplifiedTheme_BPL);
+    } else if (id == 10) {
+      setTheme(R.style.SimplifiedTheme_FCPL);
```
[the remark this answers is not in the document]
datalogics-tarar: Ah. The missing numbers are for libraries we aren't working on yet. We'll add them in order as we deploy them.

datalogics-tarar:
Since most of the changes you requested are outside the realm of the library we're adding, I'll create a separate PR with just those changes.

datalogics-dans:
It looks like v21/styles.xml is incomplete, by my counting. And values/styles.xml is _way_ incomplete. I haven't noticed any issues, but maybe I haven't been looking closely enough.

PR #478, opened 2017-11-07
Allow the PIN field's length and type to be configurable

datalogics-dans:
Two new fields can be parsed from Accounts.json - pin length and pin
type. By default, it will be unlimited and numbers only.

This commit does not fully implement this feature. Reading the config
doesn't work right when done from the accounts screen.

### DIFF
--- a/simplified-app-shared/src/main/java/org/nypl/simplified/app/LoginDialog.java
+++ b/simplified-app-shared/src/main/java/org/nypl/simplified/app/LoginDialog.java
@@ -61,6 +61,8 @@ public final class LoginDialog extends DialogFragment
   private static final String PIN_ID;
   private static final String TEXT_ID;
   private static final String ACCOUNT_ID;
+  private static final String PIN_ALLOWS_LETTERS;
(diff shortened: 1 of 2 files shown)


datalogics-dans:
To clarify the commit message, the code currently doesn't know what account you're looking at if you try logging in through the accounts screen. It only knows which account is active (i.e. which account's colors are visible in the view).

I'd like some advice on how I should be figuring that out.

gioneill:
You laid out the groundwork for me, thanks. Login's can happen in Settings, OR when a user tries to check out a book on the fly. In the latter case, we can use the current account properties and serialize them through to the Dialog. Otherwise in Settings, we already have an instance of Account, so it was just a matter of serializing the two new properties into the Bundle that gets passed through.

Take a look at what we have, make sure to test on your branch, then let me know if you approve and I can merge it in!

Thanks again Dan!

datalogics-dans:
I built it and tested it out and your changes look great! Feel free to merge anytime.
Thanks for making my changes work right!

PR #484, opened 2017-11-10
Fix color for SimplyE collection

winniequinn:
This is a hack, but the whole approach to branding and colors needs to be redone anyway. There's no nice way to fix this for now. Refactoring the branding will be done as part of the library registry work.


gioneill:
So I think SimplyE Colletion used to be the final “else” case in all of the various areas that we’re now hardcoding Multi-Library for. I doubt this “main color” existed for anything other than to prevent unexpected crashes in other areas..

PR #488, opened 2017-11-14
Fix calculation of progress within the entire book

### LINKED ISSUE #469: Progress indicator goes back when advancing chapters
It seems the algorithm for the progress indicator is not correct. It moves up each time you flip a page, but then it jumps back a bit when you hit the next chapter. Presumably someone tries to take into account the current chapter as well as the position within the current chapter and just messed up the math. 

This affects both SimplyE and Open eBooks.

winniequinn:
The previous approach was based on a heuristic that did not make much sense and caused the progress bar to jump backwards when moving into a new spine item. The approach in this commit does not give us a totally accurate indication of progress but it will do quite well overall and avoids having to scan through the entire book. The only downside is the bar will move slightly more per page turn when in shorter spine items and slightly less per page turn when in longer spine items.

@gioneill, if you like this approach, let me know and we can roll out the same thing on iOS.

Closes #469.


winniequinn:
@gioneill Yes, it is. The spine items are available as part of a book's metadata.

PR #491, opened 2017-11-17
ZXing Barcode Scanner and Display

mattniehoff:
Adds a barcode scanner option to the LoginDialog and for libraries set for supportsBarcodeDisplay, will expose a "Show Barcode" toggle.

For this PR, I've set NYPL as supporting it, but left Brooklyn as not.


simplified-app-shared/src/main/java/org/nypl/simplified/app/LoginDialog.java
```diff
@@ -405,6 +418,15 @@ public void onAccountLoginFailureDeviceActivationError(final String message)
       final Account account = new AccountsRegistry(getActivity()).getAccount(Integer.valueOf(account_id));
       books = Simplified.getBooks(account, getActivity(), Simplified.getCatalogAppServices().getAdobeDRMExecutor());
       docs = Simplified.getDocumentStore(account, getActivity().getResources());
+
+      if (account.supportsBarcodeScanner()) {
```
gioneill: One thing I'm wondering: if these two key/values are missing in Accounts.json, would trying to access this variable from `account` throw a null exception? I haven't tried and don't know enough about Java to know if it falls back gracefully.

If it does crash, just for these two variables, we should provide defaults of `false` for scan and display in `Accounts.java`. The whole class is getting refactored soon with the addition of the account registry.

Since outside contributors haven't been accustomed to adding these properties for other libraries, it's a likely point of failure.

simplified-app-shared/src/main/java/org/nypl/simplified/app/LoginDialog.java
```diff
@@ -451,6 +474,23 @@ public void onAccountLoginFailureDeviceActivationError(final String message)
...
+          public void onClick(
+              final @Nullable View v) {
+
+            // IntentIntegrator exit will fire on Scan or Back and hit the onActivityResult method.
+            IntentIntegrator
+                .forFragment(myFragment)
+                .setPrompt("Scan your library barcode.")
+                .setBeepEnabled(false)
+                //.setDesiredBarcodeFormats(IntentIntegrator.ALL_CODE_TYPES)
+                //.setOrientationLocked(false)
```
gioneill: Parameters you've set look good. Just needs comment cleanup.

simplified-app-shared/src/main/java/org/nypl/simplified/app/MainSettingsAccountActivity.java
```diff
@@ -941,6 +981,26 @@ public void run() {
             in_pin_text.setText(creds.getPin().toString());
             in_pin_text.setContentDescription(creds.getPin().toString().replaceAll(".(?=.)", "$0,"));
 
+            if (account.supportsBarcodeDisplay()) {
```
gioneill: I would say to add the conditional that generateBarcodeImage doesn't fail. We don't need any of the objects visible or setting onClick listeners if we know there's not going to be a valid image created.

This will be fairly common trying to use an alphanumeric "barcode" which is probably a good percentage of users in certain library systems where usernames have been in common use for awhile. The variable `barcode` should really be updated to `username` but I'll get to that another time.

simplified-app-shared/src/main/java/org/nypl/simplified/app/LoginDialog.java
mattniehoff: If the pair is missing, it is defaulting to false on these account methods.

simplified-app-shared/src/main/java/org/nypl/simplified/app/LoginDialog.java
mattniehoff: See c85d987

simplified-app-shared/src/main/java/org/nypl/simplified/app/MainSettingsAccountActivity.java
mattniehoff: See a5f2ceb

PR #503, opened 2017-12-16
Upgrade Gradle Android tools to 3.0.1

io7m:
This upgrades the Android Gradle tools to 3.0.1. Android Studio doesn't stop complaining unless you do this!

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,10 +1,11 @@
 buildscript {
   repositories {
+    google()
     jcenter()
   }
 
   dependencies {
-    classpath 'com.android.tools.build:gradle:2.3.3'
+    classpath 'com.android.tools.build:gradle:3.0.1'
     classpath 'com.netflix.nebula:gradle-lint-plugin:7.4.0'
     classpath 'com.nabilhachicha:android-native-dependencies:0.1.2'
   }


winniequinn:
@io7m This PR also requires updating `gradle-wrapper.properties` because Gradle 3.5 does not have the `google` function. Attempting to build will get you this:

```
> Could not find method google() for arguments [] on repository container of type 
org.gradle.api.internal.artifacts.dsl.DefaultRepositoryHandler.
```

Simply updating to Gradle 4.4 causes the build to emit a number of warnings though because `compile` is deprecated: You're supposed to use `implementation` now. As such, I'll go ahead and make those changes too.

winniequinn:
@io7m It seems I spoke too soon. Everything worked with 4.4, but it only worked until I did `--refresh-dependencies`. It now fails due to what seems like a large number of dependencies that _should_ have been declared but never were. I'll see if I can't go ahead and add all of them.

winniequinn:
@io7m It's too much to do in this PR. It seems `implementation` no longer pulls in transitive dependencies the way `compile` did. I'll take this with `compile` for now and make a new issue for moving to `implementation`.

winniequinn:
@io7m This will be merged via #506 instead which has the changes to `gradle-wrapper.properties`.

winniequinn:
The move from `compile` to `implementation` is tracked via #507.

io7m:
I didn't realize `gradle-wrapper.properties`. The very first time I ran Gradle, it modified files that were in version control...

PR #504, opened 2017-12-16
Centralize tests

io7m:
This one cleans up the entire test suite.

When the application was first written, Android had no support for JUnit 4 and required all tests to extend proprietary Android-specific classes. Because running tests on real devices is very slow, it was desirable to be able to run non-Android-specific tests locally during development. Unfortunately, this meant that the test suite had to be split into a set of classes that implemented tests without making any references to `junit`, a `junit4` frontend that ran the tests locally by delegating to the `junit`-ignorant implementation, and an `android` frontend that ran the tests on a real device by delegating to the `junit`-ignorant implementation.

Nowadays, Android does support JUnit 4, so the entire test suite can be drastically simplified. It's still necessary to have the two frontends, but the burden of maintenance is severely reduced as the frontends are now empty classes that extend abstract classes, possibly with a couple of annotations added. This eliminates a few hundred lines of boilerplate. All of the tests are now moved into two modules, and the `test-utilities` module is gone entirely as it's no longer needed.

There are still test failures, but this is due to the failure of the previous maintainers to update the tests when they changed functionality. 🤕 

This also includes the commit from #503. I'm not entirely sure how to prune that out, but it should merge cleanly anyway.


winniequinn:
This is working for me on top of the latest master. Merging! Thanks, @io7m!

io7m:
👍 

I've fixed all of the unit tests in my fork, but I'm not sure how much of this work can be merged back in without bringing in other changes yet.

PR #511, opened 2017-12-28
Document how page turns are animated

winniequinn:
As per @daryanypl's request.


winniequinn:
@gioneill It might. I've added NYPL-Simplified/Simplified-iOS#886 so we can get to it eventually.

PR #527, opened 2018-02-05
Add Simplified bearer token support

### LINKED ISSUE #493: Implement bearer token flow for book downloads

winniequinn:
Closes #493.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,4 +1,5 @@
 buildscript {
+  ext.kotlin_version = '1.2.21'
   repositories {
     google()
     jcenter()
@@ -8,6 +9,7 @@ buildscript {
     classpath 'com.android.tools.build:gradle:3.0.1'
     classpath 'com.netflix.nebula:gradle-lint-plugin:7.4.0'
     classpath 'com.nabilhachicha:android-native-dependencies:0.1.2'
+    classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
   }
 }
 
--- a/simplified-books-core/src/main/java/org/nypl/simplified/books/core/SimplifiedBearerToken.kt
+++ b/simplified-books-core/src/main/java/org/nypl/simplified/books/core/SimplifiedBearerToken.kt
@@ -1,0 +1,38 @@
+package org.nypl.simplified.books.core
+
+import org.json.JSONException
+import org.json.JSONObject
+import java.net.URI
+import java.net.URISyntaxException
+import java.util.Calendar
+import java.util.Date
+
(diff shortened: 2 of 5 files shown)


build.gradle
```diff
@@ -8,6 +9,7 @@ buildscript {
     classpath 'com.android.tools.build:gradle:3.0.1'
     classpath 'com.netflix.nebula:gradle-lint-plugin:7.4.0'
     classpath 'com.nabilhachicha:android-native-dependencies:0.1.2'
+    classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
```
gioneill: Kotlin could provide a lot of neat language features and just a general pleasure to programming for our app. I'm glad to see you taking my lead and adopting it for your work.

simplified-books-core/src/main/java/org/nypl/simplified/books/core/SimplifiedBearerToken.kt
```diff
@@ -0,0 +1,38 @@
...
+
+internal data class SimplifiedBearerToken(
+    val accessToken: String,
+    val expiration: Date,
+    val location: URI) {
+
+  companion object Factory {
+    fun withJSONObject(jsonObject: JSONObject): SimplifiedBearerToken? {
+      try {
+        if (jsonObject.getString("token_type") != "Bearer") {
```
gioneill: Should we add any kind of logging in the interest of any potential issues as a completely new process implemented in the app to potentially help the server team? Any warning-level issues are automatically uploaded to Bugsnag.
This was my yet-untested attempt:
```
import org.slf4j.LoggerFactory
...
private companion object {
    val LOG = LoggerFactory.getLogger(AnnotationsManager::class.java)!!
}
```

build.gradle
winniequinn: GitHub is no place for joking around and telling lies, @gioneill!

simplified-books-core/src/main/java/org/nypl/simplified/books/core/SimplifiedBearerToken.kt
winniequinn: I'll add logging for this one!

winniequinn: @gioneill Actually, I already have logging for this one. If the token fails to parse, the caller of `withJSONObject` throws an exception with a message indicating as such and that is then logged after the exception is caught. I prefer that to having the logging in the parsing code itself because it's not really the parsing that's the problem—it's the data the caller is passing to it. As such, I think we're all set!

PR #540, opened 2018-03-20
Annotation Reading Sync and Related Classes

gioneill:
Work is ongoing for Bookmarks-related features (some methods are just converted from Swift and not yet tested)

Main Classes:
`AnnotationsManager.kt` and `ReaderActivitySyncManager.kt`

Annotations Manager provides the main business logic and networking support for Web Annotations used in SimplyE: Bookmarks and Page Position Syncing.

Sync Manager has an interface to the Reader Activity to keep bloat out.

Settings has a Switch with UI logic based on server and network conditions.

### DIFF
--- a/simplified-app-shared/src/main/java/org/nypl/simplified/app/AnnotationsManager.kt
+++ b/simplified-app-shared/src/main/java/org/nypl/simplified/app/AnnotationsManager.kt
@@ -152,7 +152,7 @@ class AnnotationsManager(private val libraryAccount: Account,
                                           completion: (success: Boolean) -> Unit) {
 
     val jsonBody = JSONObject(parameters)
-    val additionalHeaders = mapOf<String,String>("vnd.librarysimplified/user-profile+json" to "Content-Type")
+    val additionalHeaders = mapOf<String,String>("Content-Type" to "vnd.librarysimplified/user-profile+json")
 
     val request = NYPLJsonObjectRequest(
         Request.Method.PUT,
--- a/simplified-app-shared/src/main/java/org/nypl/simplified/app/AnnotationsManager.kt
+++ b/simplified-app-shared/src/main/java/org/nypl/simplified/app/AnnotationsManager.kt
@@ -1,0 +1,647 @@
+package org.nypl.simplified.app
+
+import android.content.Context
+import android.net.Uri
+import com.android.volley.*
+import com.android.volley.Response.Listener
+import com.android.volley.toolbox.Volley
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.google.gson.Gson
+import com.google.gson.JsonSyntaxException
+import com.io7m.jfunctional.Some
+import org.json.JSONException
+import org.json.JSONObject
+import org.joda.time.Instant
+import org.nypl.drm.core.AdobeDeviceID
+import org.nypl.simplified.app.catalog.annotation.AnnotationResult
+import org.nypl.simplified.app.reader.ReaderBookLocation
+import org.nypl.simplified.books.core.AccountCredentials
+import org.nypl.simplified.books.core.AccountsControllerType
+import org.nypl.simplified.multilibrary.Account
+import org.nypl.simplified.volley.NYPLJsonObjectRequest
+import org.slf4j.LoggerFactory
+import java.io.IOException
+
+/**
+ * Handles any network requests relevant to OPDS Annotations
+ * for user settings or specific OPDS Entries for that user.
+ */
+
+class AnnotationsManager(private val libraryAccount: Account,
+                         private val credentials: AccountCredentials,
+                         context: Context) {
+
+  private val requestQueue = Volley.newRequestQueue(context)
+
+  private companion object {
+    val LOG = LoggerFactory.getLogger(AnnotationsManager::class.java)!!
+  }
+
+  /**
+   * Polls the server, and based on a set of conditions, decides whether
+   * or not to inform the user about "Syncing" as a feature via an Alert, where
+   * the user can choose to dismiss/ignore, or turn it on.
+   * @param account The particular library for the permission status request.
+   * @param completion Asynchronous handler for the user's decision, or error.
+   */
+  fun requestServerSyncPermissionStatus(account: AccountsControllerType,
+                                        completion: (enableSync: Boolean) -> Unit) {
+
+    if (!syncIsPossible(account)) {
+      LOG.debug("Account does not satisfy conditions for sync setting request.")
+      completion(false)
+      return
+    }
+
+    //TODO update these calls after updating Prefs.java to support multiple libraries
+    if (Simplified.getSharedPrefs().getBoolean("userHasSeenFirstTimeSyncMessage") == true &&
+        Simplified.getSharedPrefs().getBoolean("syncPermissionGranted") == false) {
+      completion(false)
+      return
+    }
+
+    syncPermissionUriRequest { initialized, syncIsPermitted ->
+      if (initialized && syncIsPermitted) {
+        Simplified.getSharedPrefs().putBoolean("userHasSeenFirstTimeSyncMessage", true)
+        LOG.debug("Sync has already been enabled on the server. Enable here as well.")
+        completion(true)
+      } else if (!initialized && !Simplified.getSharedPrefs().getBoolean("userHasSeenFirstTimeSyncMessage")) {
+        LOG.debug("Sync has never been initialized for the patron. Proceeding with alert.")
+        displayFirstTimeSyncMessage(completion)
+      } else {
+        completion(false)
+      }
+    }
+  }
+
+  private fun syncPermissionUriRequest(completion: (initialized: Boolean, syncIsPermitted: Boolean) -> Unit) {
+
+    val catalogUriString = libraryAccount.catalogUrl ?: null
+    val baseUri = if (catalogUriString != null) Uri.parse(catalogUriString) else null
+    val permissionRequestUri = if (baseUri != null) Uri.withAppendedPath(baseUri, "patrons/me/") else null
+
+    if (permissionRequestUri == null) {
+      LOG.debug("Could not create Annotations URL from Main Feed URL. Abandoning attempt to retrieve sync setting.")
+      return
+    }
+
+    val request = NYPLJsonObjectRequest(
+        Request.Method.GET,
+        permissionRequestUri.toString(),
+        credentials.barcode.toString(),
+        credentials.pin.toString(),
+        null,
+        Listener<JSONObject> { response ->
+          handleSyncPermission(response, completion)
+        },
+        Response.ErrorListener { error ->
+          LOG.error("GET request fail! Error: ${error.message}")
+        })
+
+    request.retryPolicy = DefaultRetryPolicy(
+        60 * 1000,
+        DefaultRetryPolicy.DEFAULT_MAX_RETRIES,
+        DefaultRetryPolicy.DEFAULT_BACKOFF_MULT)
+
+    requestQueue.add(request)
+  }
+
+  private fun handleSyncPermission(response: JSONObject,
+                                   completion: (initialized: Boolean, syncIsPermitted: Boolean) -> Unit) {
+
+    try {
+      val settings = response.getJSONObject("settings")
+      val syncSettingsBool = settings.getBoolean("simplified:synchronize_annotations")
+      completion(true, syncSettingsBool)
+    } catch (e: java.lang.Exception) {
+      LOG.error("Exception thrown during JSON deserialization: ${e.localizedMessage}")
+      completion(false, false)
+    }
+  }
+
+  /**
+   * Set the current user's permission setting on the server. If enabled is set to 'false',
+   * all annotations for the current user will be deleted as part of this request.
+   */
+  fun updateServerSyncPermissionStatus(enabled: Boolean,
+                                       completion: (successful: Boolean) -> Unit) {
+
+    val catalogUriString = libraryAccount.catalogUrl ?: null
+    val baseUri = if (catalogUriString != null) Uri.parse(catalogUriString) else null
+    val patronsUpdateUri = if (baseUri != null) Uri.withAppendedPath(baseUri, "patrons/me/") else null
+
+    if (patronsUpdateUri == null) {
+      LOG.debug("Could not get Annotations URI. Abandoning attempt to update sync setting.")
+      completion(false)
+      return
+    }
+
+    val parameters = mapOf<String,Any>("settings" to mapOf("simplified:synchronize_annotations" to enabled))
+
+    setSyncPermissionUriRequest(patronsUpdateUri.toString(), parameters, 20) { isSuccessful ->
+      if (!isSuccessful) {
+        presentAlertForSyncSettingError()
+      }
+      completion(isSuccessful)
+    }
+  }
+
+  private fun setSyncPermissionUriRequest(uri: String,
+                                          parameters: Map<String, Any>,
+                                          timeout: Int?,
+                                          completion: (success: Boolean) -> Unit) {
+
+    val jsonBody = JSONObject(parameters)
+    val additionalHeaders = mapOf<String,String>("vnd.librarysimplified/user-profile+json" to "Content-Type")
+
+    val request = NYPLJsonObjectRequest(
+        Request.Method.PUT,
+        uri,
+        credentials.barcode.toString(),
+        credentials.pin.toString(),
+        jsonBody,
+        additionalHeaders,
+        Listener<JSONObject> { _ ->
+          completion(true)
+        },
+        Response.ErrorListener { error ->
+          LOG.debug("PUT request fail! Error: ${error.message}")
+        })
+
+    if (timeout != null) {
+      request.retryPolicy = DefaultRetryPolicy(
+          timeout * 1000,
+          DefaultRetryPolicy.DEFAULT_MAX_RETRIES,
+          DefaultRetryPolicy.DEFAULT_BACKOFF_MULT)
+    }
+
+    requestQueue.add(request)
+  }
+
+  /**
+   * Get the current reading position for the given book and current user.
+   * @param bookID Identifier for the entry
+   * @param uri The Annotation ID, URI, for the OPDS Entry
+   * @param completion Called asynchronously by the network request to return the CFI
+   */
+  fun requestReadingPositionOnServer(bookID: String?, uri: String?,
+                                     completion: (location: ReaderBookLocation?) -> Unit ) {
+
+    if (!syncIsPossibleAndPermitted()) {
+      LOG.debug("Account does not support sync or sync is disabled.")
+      return
+    }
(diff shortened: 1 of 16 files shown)


simplified-app-shared/src/main/java/org/nypl/simplified/app/ReaderActivitySyncManager.kt
```diff
@@ -0,0 +1,173 @@
...
+
+    if (delayPageSync) {
+      LOG.debug("Post of last read position delayed. Initial sync still in progress.")
+      return
+    }
+
+    if (annotationsManager.syncIsPossibleAndPermitted()) {
+      queueTimer.cancel()
+      queueTimer = Timer()
+      queueTimer.schedule(3000) {
```
gioneill: This was a direct port of the current rate-limiting code to server requests. It will be updated to reflect a more robust implementation to match iOS.

simplified-app-shared/src/main/java/org/nypl/simplified/app/AnnotationsManager.kt
```diff
@@ -0,0 +1,600 @@
...
+      } else {
+        LOG.debug("Sync has been initialized and not permitted. Continuing with sync disabled.")
+        completion(false)
+      }
+    }
+  }
+
+  private fun syncPermissionStatusUriRequest(completion: (initialized: Boolean, syncIsPermitted: Boolean) -> Unit)
+  {
+    val catalogUriString = libraryAccount.catalogUrl ?: null
```
djds23: do we need this `?:`? I thought the Elvis operator would only return the right side of the expression if the left is `null` already

simplified-app-shared/src/main/java/org/nypl/simplified/app/AnnotationsManager.kt
```diff
@@ -0,0 +1,600 @@
...
+    val baseUri = if (catalogUriString != null) Uri.parse(catalogUriString) else null
+    val permissionRequestUri = if (baseUri != null) Uri.withAppendedPath(baseUri, "patrons/me/") else null
+
+    if (permissionRequestUri == null) {
+      LOG.debug("Could not create Annotations URL from Main Feed URL. Abandoning attempt to retrieve sync setting.")
+      return
```
djds23: shouldn't we call the completion block here with an error? otherwise this would just cause the user to sit on a page.

simplified-app-shared/src/main/java/org/nypl/simplified/app/AnnotationsManager.kt
```diff
@@ -0,0 +1,600 @@
...
+        null,
+        Listener<JSONObject> { response ->
+          handleSyncPermission(response, completion)
+        },
+        Response.ErrorListener { error ->
+          LOG.error("GET request fail! Error: ${error.message}")
+        })
+
+    request.retryPolicy = DefaultRetryPolicy(
+        60 * 1000,
```
djds23: can we give these numbers names? `secondsInMinute * minutesUntilTimeout` or something

djds23: Maybe even make this value a `DEFAULT_RETRY_TIMEOUT` const on `DefaultRetryPolicy` or something of the like

simplified-app-shared/src/main/java/org/nypl/simplified/app/AnnotationsManager.kt
```diff
@@ -0,0 +1,600 @@
...
+    requestQueue.add(request)
+  }
+
+  private fun handleSyncPermission(response: JSONObject,
+                                   completion: (initialized: Boolean, syncIsPermitted: Boolean) -> Unit)
+  {
+    try {
+      val settings = response.getJSONObject("settings")
+      val syncSettingsBool = settings.getBoolean("simplified:synchronize_annotations")
+      completion(true, syncSettingsBool)
```
djds23: not to open up a can of worms, but are we guaranteed that this is called on a UI Thread? do we need such guarantees?

simplified-app-shared/src/main/java/org/nypl/simplified/app/AnnotationsManager.kt
```diff
@@ -0,0 +1,600 @@
...
+        Listener<JSONObject> { _ ->
+          completion(true)
+        },
+        Response.ErrorListener { error ->
+          LOG.debug("PUT request fail! Error: ${error.message}")
+        })
+
+    if (timeout != null) {
+      request.retryPolicy = DefaultRetryPolicy(
+          timeout * 1000,
```
djds23: same note about numbers, I'm not sure what the number on the right means

simplified-app-shared/src/main/java/org/nypl/simplified/app/AnnotationsManager.kt
gioneill: This is a good question and Winnie and I were discussing this as well. The compiler gives me the inferred type which is this "platform type" of "either nullable or nonnull"... which makes me kind of nervous. My first instinct as to simply failably cast "as?" but the compiler says "don't need that"... 

I will be reviewing these areas that I use the elvis for this purpose and I think the Idea that JetBrains is pushing to to just look downstream and protect for NPE's where appropriate.. I just don't have an intuitive feel for that yet.

simplified-app-shared/src/main/java/org/nypl/simplified/app/AnnotationsManager.kt
```diff
@@ -0,0 +1,600 @@
...
+        credentials.barcode.toString(),
+        credentials.pin.toString(),
+        null,
+        Listener<JSONObject> { response ->
+
+          try {
+            val first = response.getJSONObject("first")
+            val items = first?.getJSONArray("items")
+            if (items == null) {
+              LOG.error("")  //TODO
```
djds23: what's there TODO?

simplified-app-shared/src/main/java/org/nypl/simplified/app/AnnotationsManager.kt
```diff
@@ -0,0 +1,600 @@
...
+
+    NYPLJsonObjectRequest(
+        Request.Method.GET,
+        uri.toString(),
+        credentials.barcode.toString(),
+        credentials.pin.toString(),
+        null,
+        Listener<JSONObject> { response ->
+
+          try {
```
djds23: This callback lambda is quite large, for readability I recommend making it a new method.

simplified-app-shared/src/main/java/org/nypl/simplified/app/AnnotationsManager.kt
djds23: cool

simplified-app-shared/src/main/java/org/nypl/simplified/app/MainSettingsAccountActivity.java
```diff
@@ -890,97 +904,101 @@ protected void onResume() {
...
+
+    in_barcode_text.setText(creds.getBarcode().toString());
+    in_barcode_text.setContentDescription(creds.getBarcode().toString().replaceAll(".(?=.)", "$0,"));
+    in_pin_text.setText(creds.getPin().toString());
+    in_pin_text.setContentDescription(creds.getPin().toString().replaceAll(".(?=.)", "$0,"));
+
+    in_eula_checkbox.setEnabled(false);
+
+    in_login.setText(rr.getString(R.string.settings_log_out));
+    in_login.setOnClickListener( v -> {
```
djds23: what is `v`?

simplified-app-shared/src/main/java/org/nypl/simplified/app/MainSettingsAccountActivity.java
```diff
@@ -890,97 +904,101 @@ protected void onResume() {
...
+      this.sync_table_row.setVisibility(View.GONE);
+    }
+
+    in_table_with_code.setVisibility(View.VISIBLE);
+    in_table_signup.setVisibility(View.GONE);
+
+    in_barcode_text.setText(creds.getBarcode().toString());
+    in_barcode_text.setContentDescription(creds.getBarcode().toString().replaceAll(".(?=.)", "$0,"));
+    in_pin_text.setText(creds.getPin().toString());
+    in_pin_text.setContentDescription(creds.getPin().toString().replaceAll(".(?=.)", "$0,"));
```
djds23: what is this looking for explicitly? can we sanitize the string in another way?

simplified-app-shared/src/main/java/org/nypl/simplified/app/ReaderActivitySyncManager.kt
```diff
@@ -0,0 +1,173 @@
...
+  private val annotationsManager = AnnotationsManager(libraryAccount, credentials, delegate as Context)
+  private var queueTimer = Timer()
+
+  /**
+   * See if sync is enabled on the server before attempting to synchronize
+   * bookmarks or reading position. Set sharedPrefs accordingly.
+   * Execute the closure if it's already on or successfully enabled on the server.
+   */
+  fun serverSyncPermission(account: AccountsControllerType,
+                           completion: () -> Unit)
```
djds23: if two callers both hit this method, will we fire two requests? or will we store completion blocks and just call them when the outstanding request is completed?

simplified-app-shared/src/main/java/org/nypl/simplified/app/AnnotationsManager.kt
```diff
@@ -0,0 +1,600 @@
...
+
+  /**
+   * Queries the server, and based on a set of conditions, decides whether
+   * or not to inform the user about "Syncing" in an Alert Dialog, where
+   * the user can choose to dismiss/ignore, or turn it on for the first time.
+   * @param account The particular library for the permission status request.
+   * @param completion Asynchronous handler for the user's decision, or error.
+   */
+  fun requestServerSyncPermissionStatus(account: AccountsControllerType,
+                                        completion: (enableSync: Boolean) -> Unit)
```
djds23: if two callers both hit this method, will we fire two requests? is that okay for us todo? or should we store completion blocks and just call them when the outstanding request is completed?

simplified-app-shared/src/main/java/org/nypl/simplified/app/ReaderActivitySyncManager.kt
```diff
@@ -0,0 +1,173 @@
...
+      delayPageSync = false
+      LOG.error("No annotations uri or feed entry exists for this book. Abandoning sync attempt.")
+      return
+    }
+
+    val uriString = (feedEntry.annotations as Some<URI>).get().toString()
+
+    annotationsManager.requestReadingPositionOnServer(entryID, uriString, { serverLocation ->
+
+      delayPageSync = false
```
djds23: for readability, I would suggest making this its own method

simplified-app-shared/src/main/java/org/nypl/simplified/app/reader/ReaderActivity.java
```diff
@@ -585,27 +561,23 @@ else if (id == 1) {
...
-        @Override public void onClick(
-          final @Nullable View v)
-        {
-          final ReaderTOC sent_toc = ReaderTOC.fromPackage(p);
-          ReaderTOCActivity.startActivityForResult(
-            ReaderActivity.this, sent_toc);
-          ReaderActivity.this.overridePendingTransition(0, 0);
-        }
-      });
+    in_toc.setOnClickListener( v -> {
```
djds23: what is `v`?

simplified-app-shared/src/main/java/org/nypl/simplified/app/AnnotationsManager.kt
dshen6: I really like `TimeUnit.SECONDS.toMillis(60)` for these cases

simplified-app-shared/src/main/java/org/nypl/simplified/app/ReaderActivitySyncManager.kt
```diff
@@ -0,0 +1,173 @@
...
+import org.nypl.simplified.opds.core.OPDSAcquisitionFeedEntry
+import org.readium.sdk.android.Package
+import org.slf4j.LoggerFactory
+import java.net.URI
+import java.util.Timer
+import kotlin.concurrent.schedule
+
+
+interface ReaderSyncManagerDelegate {
+  fun navigateToLocation(location: ReaderBookLocation)
```
dshen6: this looks useful - unrelated but just curious, how might one translate a ReaderPaginationChangedEvent  -> ReaderBookLocation? Now that this function exists (and does what I assume), you can probably save and restore`ReaderPaginationChangedEvents` on `onReadiumFunctionSettingsApplied`

simplified-app-shared/src/main/java/org/nypl/simplified/app/ReaderActivitySyncManager.kt
```diff
@@ -0,0 +1,173 @@
...
+          alert.show()
+        }
+      }
+    })
+  }
+
+  private fun createAlertForSyncLocation(location: ReaderBookLocation, context: Context,
+                                       completion: (shouldMove: Boolean) -> Unit): AlertDialog
+  {
+    val builder= AlertDialog.Builder(context)
```
dshen6: you could use `with(builder)` here if you like

simplified-app-shared/src/main/java/org/nypl/simplified/app/ReaderActivitySyncManager.kt
```diff
@@ -0,0 +1,173 @@
...
+    builder.setTitle("Sync Reading Position")
```
dshen6: these strings probably belong in `strings.xml`?

simplified-app-shared/src/main/java/org/nypl/simplified/app/ReaderActivitySyncManager.kt
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  //TODO Work In Progress -
+  //TODO observe when set, then show jump to location dialog with appropriate messaging
+  val bookPackage: Package? = null
+
+  // Delay server posts until first page sync is complete
+  private var delayPageSync: Boolean = true
+  private val annotationsManager = AnnotationsManager(libraryAccount, credentials, delegate as Context)
+  private var queueTimer = Timer()
```
dshen6: hmm, do we want to cancel this timer when the app backgrounds?

dshen6:
Kotlin <3

simplified-app-shared/src/main/java/org/nypl/simplified/app/AnnotationsManager.kt
dshen6: I might do `val baseUri = libraryAccount.catalogUrl?.let { Uri.parse(it)}`, but just personal preference

gioneill: Thanks David. I read up on extension functions a little bit. I'll keep an eye out for more places to use this.

simplified-app-shared/src/main/java/org/nypl/simplified/app/AnnotationsManager.kt
gioneill: How about I take both of your advice?

simplified-app-shared/src/main/java/org/nypl/simplified/app/AnnotationsManager.kt
gioneill: Not gauranteed, and I don't imagine getting there very easily, but it can't hurt to check and throw an exception for done the road. `UIThread.checkIsUIThread()` is used in this way in many other spots of the app 👍

simplified-app-shared/src/main/java/org/nypl/simplified/app/MainSettingsAccountActivity.java
gioneill: In Java lambda syntax, I don't seem to see an equivalen "_" syntax to omit unused parameters. In this case the compile auto-completed the logout dialog "view" as just "v", but I don't mind editing those for clarity.

simplified-app-shared/src/main/java/org/nypl/simplified/app/MainSettingsAccountActivity.java
gioneill: This is doing specific work for accessibilty which I'm not sure of. Must have been lumped into this PR from some re-indenting on my part.

simplified-app-shared/src/main/java/org/nypl/simplified/app/ReaderActivitySyncManager.kt
gioneill: See below

simplified-app-shared/src/main/java/org/nypl/simplified/app/AnnotationsManager.kt
gioneill: Thanks for the discussion. It's a pretty rare possibility, but I've gone ahead and made some changes to queue up any possible caller's completions so that multiple network requests are not made. This will save requests, as an example, if someone begins to spam the switch in Settings.

simplified-app-shared/src/main/java/org/nypl/simplified/app/ReaderActivitySyncManager.kt
gioneill: Ideally the queued position will send off when the app backgrounds. This occurs on iOS and I will be bringing this functionality in with a newer PR. Thanks!

simplified-app-shared/src/main/java/org/nypl/simplified/app/ReaderActivitySyncManager.kt
gioneill: `ReaderPaginationChangedEvents` are fired off by the rendering engine each time the page is turned, and `navigateToLocation` will be utilizing a subset of that code to create an "openPage" event to take the user to another page in the book, based on the data that is syncronized with the server. Conceptually I'm kind of trying to keep these things separate in my head.

To address your question, "I don't know".. If you could let me know what you're thinking for why we'd want to change change the relationship of these rendering method/listeners you've listed.

dshen6: ah, so right now, when you change the text size or go to dusk mode/dark mode, it resets your location (you can confirm this by swiping a few pages in and changing the text size). It would good to save your location and restore it.

simplified-app-shared/src/main/java/org/nypl/simplified/app/AnnotationsManager.kt
gioneill: Yeah you're right. I've updated this. I've made a note to go back and refactor the two usages of this since one caller requires the completion for correct execution, and the other does not.

PR #544, opened 2018-03-30
Kotlin local bookmarks toc

gioneill:
Mainly Bookmarks related work:

- Updated TOC to support PagerView
- Ported networking methods into `AnnotationsManager.kt`
- GSON cleanup. Moved package. Renamed classes for clarity. `BookAnnotation.java` etc.
- Updated interfaces for bookmarks support in books database. `BookDatabase.java` and associated interfaces.
- Setup reading and writing to the database. *Some in-person discussion will be appreciated.


simplified-app-shared/src/main/java/org/nypl/simplified/app/AnnotationsManager.kt
```diff
@@ -360,6 +361,147 @@ class AnnotationsManager(private val libraryAccount: Account,
     }
   }
 
+  /**
+   * Get a list of any bookmark-type annotations created for the given book and the current user.
+   * @param bookID Identifier for the entry
+   * @param uri The Annotation ID/URI for the OPDS Entry
+   * @param completion Called asynchronously by the network request to return the bookmarks
+   */
+  fun requestBookmarksFromServer(bookID: String?, uri: Uri?,
```
gioneill: These have just been ported. Not actually tested yet.

simplified-app-shared/src/main/java/org/nypl/simplified/app/reader/ReaderActivity.java
```diff
@@ -490,15 +506,13 @@ else if (id == 1) {
...
     this.current_location = l;
 
     final SimplifiedReaderAppServicesType rs = Simplified.getReaderAppServices();
-    final ReaderBookmarksType bm = rs.getBookmarks();
-    //TODO could book_id actually be null?
+    final ReaderBookmarksSharedPrefsType bm = rs.getBookmarks();
     final BookID in_book_id = NullCheck.notNull(this.book_id);
 
-//    bm.setBookmark(in_book_id, this.current_location);
+    bm.saveReadingPosition(in_book_id, this.current_location);
```
gioneill: The app was using "bookmarks" to mean "last place you were reading" so I've updated that to add more separation and clarity between the two.

simplified-app-shared/src/main/java/org/nypl/simplified/app/reader/ReaderActivity.java
```diff
@@ -588,6 +612,50 @@ else if (id == 1) {
...
+    final String motivation = "http://www.w3.org/ns/oa#bookmarking";
+    final String bookID = this.book_id.toString();
+    final String selectorType = "oa:FragmentSelector";
+    final String value = bookmark.toJsonString();
+    final String timestamp = new Instant().toString();
+    final OptionType<String> deviceID = this.credentials.getAdobeDeviceID().map(
+        id -> id.toString()
+    );
+
+    //TODO GSON-generated classes are not aware of nullability, and are not correctly representing the "body" property
```
gioneill: I don't believe BookAnnotation was generated correctly, because the "body" node is, in its current implementation, supposed to just be set after the constructor as a JsonElement.

simplified-app-shared/src/main/java/org/nypl/simplified/app/reader/ReaderBookmarks.java
```diff
@@ -1,233 +0,0 @@
-package org.nypl.simplified.app.reader;
```
gioneill: Renamed to shared prefs, but git couldn't find it :(

simplified-app-shared/src/main/java/org/nypl/simplified/app/AnnotationsManager.kt
```diff
@@ -360,6 +361,147 @@ class AnnotationsManager(private val libraryAccount: Account,
...
+                           completion: (serverID: String?) -> Unit) {
+
+    if (!syncIsPossibleAndPermitted()) {
+      LOG.debug("Account does not support sync or sync is disabled.")
+      completion(null)
+      return
+    }
+
+    val timestamp = Instant().toString()
+    val deviceID = if (credentials.adobeDeviceID.isSome) {
```
djds23: I didn't know Kotlin had these! This is fantastic!

djds23: I love if expressions!

simplified-app-shared/src/main/java/org/nypl/simplified/app/AnnotationsManager.kt
```diff
@@ -360,6 +361,147 @@ class AnnotationsManager(private val libraryAccount: Account,
...
+        null,
+        Listener<JSONObject> { response ->
+          completion(true)
+        },
+        Response.ErrorListener { error ->
+          completion(false)
+        })
+
+    request.retryPolicy = DefaultRetryPolicy(
+        20 * 1000,
```
djds23: what do these numbers mean?

simplified-app-shared/src/main/java/org/nypl/simplified/app/reader/ReaderActivity.java
```diff
@@ -24,15 +24,15 @@
 import android.widget.ProgressBar;
 import android.widget.TextView;
 
-import com.io7m.jfunctional.FunctionType;
 import com.io7m.jfunctional.Option;
 import com.io7m.jfunctional.OptionType;
+import com.io7m.jnull.NonNull;
```
djds23: non-blocking/non-pertinate:

It looks like this `jnull` library is unmaintained, perhaps we should consider removing/replacing it.

https://github.com/io7m/jnull/issues/1

djds23: The other package referenced above `jfunctional` also appears to be unmaintained:

https://github.com/io7m/jfunctional

simplified-app-shared/src/main/java/org/nypl/simplified/app/reader/ReaderActivity.java
```diff
@@ -546,25 +560,18 @@ else if (id == 1) {
...
-      });
-
-
-
-    //TODO set package to sync manager.. this.syncManager.package = p;
-
+        new Runnable() {
+          @Override
+          public void run() {
+            in_title_text.setText(NullCheck.notNull(p.getTitle()));
```
djds23: it looks like `notNull` raises if null is passed, how sure are we that this `package` has a title? 

These `NullCheckException` do not appear to be handled anywhere in this method, is that okay?

https://github.com/io7m/jnull/blob/04808485662d04bcab4534c4c3c1da6fb55e42be/com.io7m.jnull.core/src/main/java/com/io7m/jnull/NullCheck.java#L68

simplified-app-shared/src/main/java/org/nypl/simplified/app/reader/ReaderBookmarksSharedPrefsType.kt
```diff
@@ -0,0 +1,39 @@
...
+  fun saveReadingPosition(id: BookID,
+                          bookmark: ReaderBookLocation)
+
+  /**
+   * Retrieve list of user-created bookmarks.
+   *
+   * @param id The ID of the book
+   * @param entry feed entry
+   * @return A list bookmark annotations, if any
+   */
```
djds23: documentation for a ghost method 👻

simplified-app-shared/src/main/java/org/nypl/simplified/app/reader/ReaderTOCBookmarksFragment.kt
```diff
@@ -1,18 +1,132 @@
...
 
   override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?,
                             savedInstanceState: Bundle?): View? {
 
-    return inflater.inflate(R.layout.reader_toc_bookmarks, null)
+    this.inflater = inflater
+    bookmarksTOCLayout = inflater.inflate(R.layout.reader_toc_bookmarks, null)
+    bookmarksTOCListView = bookmarksTOCLayout?.findViewById(R.id.bookmarks_list) as? ListView
+
+    val activity = activity as? ReaderTOCActivity
```
djds23: where does this activity variable come from?

simplified-app-shared/src/main/java/org/nypl/simplified/app/reader/ReaderTOCBookmarksFragment.kt
```diff
@@ -1,18 +1,132 @@
...
+  override fun getItem(position: Int): Any {
+    return adapter!!.getItem(position)
+  }
+
+  override fun getItemId(position: Int): Long {
+    return adapter!!.getItemId(position)
+  }
+
+  override fun getItemViewType(position: Int): Int {
+    return adapter!!.getItemViewType(position)
```
djds23: is there a way we can provide default values here instead of hitting null pointer exceptions? 

https://kotlinlang.org/docs/reference/null-safety.html#safe-calls

djds23: I am referring to all uses of `adapter!!` in this class

simplified-app-shared/src/main/java/org/nypl/simplified/app/reader/ReaderTOCBookmarksFragment.kt
djds23: property on fragment (superclass)

simplified-app-shared/src/main/java/org/nypl/simplified/app/reader/ReaderTOCBookmarksFragment.kt
```diff
@@ -1,18 +1,132 @@
...
+
+    val itemView = if (reuse != null) {
+      reuse as ViewGroup
+    } else {
+      inflater?.inflate(R.layout.reader_toc_element, parent, false) as ViewGroup
+    }
+
+    val layoutView = itemView.findViewById<ViewGroup>(R.id.toc_bookmark_element)
+    val textView = layoutView.findViewById<TextView>(R.id.toc_bookmark_element_title)
+    val element = adapter!!.getItem(position)
```
djds23: since `element` returns an optional anyway, I think this would also be a good place for a safe cast instead of a force unwrap

https://kotlinlang.org/docs/reference/null-safety.html#safe-calls

simplified-app-shared/src/main/java/org/nypl/simplified/app/AnnotationsManager.kt
gioneill: Missed this one, thanks!

simplified-app-shared/src/main/java/org/nypl/simplified/app/reader/ReaderActivity.java
gioneill: It's by design to raise an uncaught exception when using `notNull()`

simplified-app-shared/src/main/java/org/nypl/simplified/app/reader/ReaderBookmarksSharedPrefsType.kt
gioneill: 👻

simplified-app-shared/src/main/java/org/nypl/simplified/app/reader/ReaderTOCBookmarksFragment.kt
gioneill: I'll make an issue for this so I don't forget.

PR #547, opened 2018-04-05
Compile to implement

daryanypl:
Changed build.gradle calls from "compile" to "implementation" (compile is being deprecate by google).
Updated kotlin version to 1.2.30.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -6,7 +6,7 @@ buildscript {
   }
 
   dependencies {
-    classpath 'com.android.tools.build:gradle:3.0.1'
+    classpath 'com.android.tools.build:gradle:3.1.0'
     classpath 'com.netflix.nebula:gradle-lint-plugin:7.4.0'
     classpath 'com.nabilhachicha:android-native-dependencies:0.1.2'
     classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
--- a/build.gradle
+++ b/build.gradle
@@ -1,5 +1,5 @@
 buildscript {
-  ext.kotlin_version = '1.2.21'
+  ext.kotlin_version = '1.2.30'
   repositories {
     google()
     jcenter()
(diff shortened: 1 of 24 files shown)


gioneill:
Thanks Darya. Since this change to `implementation` changes the behavior of nested dependencies.. do we need to just ensure the project fully builds to feel like we're not breaking anything? I honestly don't know.

build.gradle
```diff
@@ -1,12 +1,12 @@
 buildscript {
-  ext.kotlin_version = '1.2.21'
+  ext.kotlin_version = '1.2.30'
```
gioneill: I'd just make the Kotlin update a separate PR. And it changes so frequently that maybe it's just worth it to wait until I get the updated Bookmarks feature in (since the version will no doubt change again in a week or two).

simplified-app-shared/build.gradle
```diff
@@ -15,33 +15,48 @@ android {
   }
 }
 
+repositories {
```
gioneill: What is jcenter?

build.gradle
daryanypl: Studio was complaining about it, but it was a warning, not an error.  It will still build with the old version, so I can revert.

simplified-app-shared/build.gradle
daryanypl: Some of the stuff I was pulling in comes from there.  See this:  https://stackoverflow.com/questions/45818613/trove4j-library-cannot-be-resolved .  jcenter used to get pulled in by some of the child libraries.  Not that "implementation" doesn't do recursive dependencies, you have to add it everywhere.

daryanypl:
Well, I tested by building to my pixel and reading an overdrive book.  It seems to work.  Given the amount of serious changes this represents, I would definitely test on multiple devices before releasing into production.

PR #565, opened 2018-04-11
Short screen layout

### LINKED ISSUE #525: Buttons cut off on short screens in Welcome Screen
related to #513 

Small screens are cutting off UI elements, and in this case the buttons on the welcome screen are not present.

Maybe also worth looking at analytics to see the overall audience size of small screen sizes. The following image is from an Alcatel OneTouch:

screenshot

daryanypl:
Changed the welcome screen to spread content more evenly on a large screen, and compact better on a short screen.
Ultra-short screens, such as android watches will still need their own layout.
The landscape layout files are made redundant by the new layout-sw123 system, and will eventually need to go.
Tablet screens need testing.

Took extra icons away and changed the button color as Ricardo suggested.  Also made the layout dynamically stretchy, but clustered somewhat towards the vertical middle, as Ricardo suggested.

Was not able to add a scroll bar as the final safety precaution for small screens I wanted.  I've tested with very small screens in the emulator, and feel the number of screens that would need a scroll bar is very small (like watches, which need to be handled separately anyway).

Added new copy for talkback.  

resolves #525


gioneill:
For something like this, instead of me having to build and check on several sizes as a review, I'd be happy to just look at a couple screenshots of the emulator at different sizes. Is that possible to add on?

>The landscape layout files are made redundant by the new layout-sw123 system, and will eventually need to go.

Are there actually landscape-specific layout files for certain Activities? To my knowledge, the app does not support landscape anywhere.. at least that's how it is on iOS.

>Tablet screens need testing.
I think it would suffice to run on a size in the emulator at around a 7" screen. I'd even probably accept just looking at the "preview" layout inside the IDE without even running it.

But on that last point: Once you merge it, everyone will consider it production-ready. If you want to coordinate testing with Joe or us let us know.

daryanypl:
There are landscape files for some activities.  My guess is they were made, but not used.  As we go redesigning the interface, we can clean it up,

daryanypl:
screenshot_1523980090
screenshot_1523980503
screenshot_1523980858
screenshot_1523980866

daryanypl:
10" Tablet, 2.7", and 3.7" devices.  Looks like the welcome screen is not locked into portrait orientation, as we'd thought.  I think it rather should be.  


screenshot_1523981055
screenshot_1523981115

daryanypl:
Sent new screens to UX for final approval.

PR #574, opened 2018-04-20
Dont search unsearchable feeds issue538

daryanypl:
Search widget is no longer active or showing by default.
Now it's only displayed and activated after the catalog feed has loaded is has been found to be searchable.

This means that the customer cannot start typing a search query until the feed has finished loading, and may need to wait a noticeable time period in slow network connection conditions.  However, we felt this was a good trade-off.  The alternative was to show the search widget, load the feed, then disappear the search widget -- a frustrating experience.

Before:
screenshot_1524088199

After:
screenshot_1524089648

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,6 @@ helpstack.conf
 # backup files
 *~
 
-/.gitignore
+
 
 
--- a/.gitignore
+++ b/.gitignore
@@ -27,5 +27,4 @@ helpstack.conf
 # backup files
 *~
 
-# all Android Studio cruft
-/.idea
+
--- a/.gitignore
+++ b/.gitignore
@@ -27,4 +27,6 @@ helpstack.conf
 # backup files
 *~
 
+/.gitignore
+
 
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,9 @@ ReaderClientCert.sig
 bugsnag.conf 
 helpstack.conf 
 *.jks
+
+# backup files
+*~
+
+# all Android Studio cruft
+/.idea
--- a/simplified-app-shared/src/main/java/org/nypl/simplified/app/catalog/CatalogFeedActivity.java
+++ b/simplified-app-shared/src/main/java/org/nypl/simplified/app/catalog/CatalogFeedActivity.java
@@ -688,44 +688,29 @@ public abstract class CatalogFeedActivity extends CatalogActivity implements
     return true;
   }
 
-  private void onCreateOptionsMenuSearchItem(
-    final Menu menu_nn)
-  {
+
+
+  /**
+   * If the feed actually has a search URI, then show the search field.
+   * Otherwise, disable and hide it.
+   */
+  private void onCreateOptionsMenuSearchItem(final Menu menu_nn) {
     final MenuItem search_item = menu_nn.findItem(R.id.catalog_action_search);
 
-    /**
-     * If the feed actually has a search URI, then show the search field.
-     * Otherwise, disable and hide it.
-     */
-
     final FeedType feed_actual = NullCheck.notNull(this.feed);
+
     final OptionType<FeedSearchType> search_opt = feed_actual.getFeedSearch();
+
+    // Set some placeholder text
+    final CatalogFeedArgumentsType args = this.getArguments();
+
     boolean search_ok = false;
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -23,3 +23,9 @@ ReaderClientCert.sig
 bugsnag.conf 
 helpstack.conf 
 *.jks
+
+# backup files
+*~
+
+# all Android Studio cruft
+/.idea
```
winniequinn: We do not want to ignore the `.idea` directory. Keeping this is essential to being able to load the project in Android Studio without importing.

simplified-app-shared/src/main/java/org/nypl/simplified/app/catalog/CatalogFeedActivity.java
```diff
@@ -748,14 +733,32 @@ public Boolean onFeedSearchLocal(
...
+      CatalogFeedActivity.LOG.debug("Feed has no search opts.");
     }
 
     if (search_ok) {
+      this.search_view.setSubmitButtonEnabled(true);
+      this.search_view.setIconifiedByDefault(false);
+      search_item.setShowAsActionFlags(MenuItem.SHOW_AS_ACTION_ALWAYS);
+      search_item.expandActionView();
+
+      this.search_view.setQueryHint("Search " + this.feed.getFeedTitle());
```
winniequinn: I believe this string should be localized. We should also avoid using string concatenation here (and use a format string instead) so that different languages can handle things appropriately.

simplified-app-shared/src/main/java/org/nypl/simplified/app/catalog/CatalogFeedActivity.java
```diff
@@ -748,14 +733,32 @@ public Boolean onFeedSearchLocal(
...
+      if (args.getTitle().startsWith("Search")) {
```
winniequinn: I think there is probably a better way to determine if we have search than examining the title. We should be deciding this based on a successful parse of an OpenSearch document.

simplified-app-shared/src/main/java/org/nypl/simplified/app/catalog/CatalogFeedActivity.java
daryanypl: sure

simplified-app-shared/src/main/java/org/nypl/simplified/app/catalog/CatalogFeedActivity.java
daryanypl: There probably is, but let's explore that in a future branch.  For now, this is pre-existing code that I just moved around a bit.

simplified-app-shared/src/main/java/org/nypl/simplified/app/catalog/CatalogFeedActivity.java
```diff
@@ -748,14 +734,33 @@ public Boolean onFeedSearchLocal(
...
     }
 
     if (search_ok) {
+      this.search_view.setSubmitButtonEnabled(true);
+      this.search_view.setIconifiedByDefault(false);
+      search_item.setShowAsActionFlags(MenuItem.SHOW_AS_ACTION_ALWAYS);
+      search_item.expandActionView();
+
+      // display either the category title or the previously searched keywords 
+      this.search_view.setQueryHint(String.format("%s %s", getString(R.string.search_hint_prefix), Objects.toString(this.feed.getFeedTitle(), "")));
```
gioneill: The token for the format string should itself be in the xml file, since not every language may structure the statement in the same way, so it'd be something like `String.format(R.string.search_hint,this.feed.getFeedTitle());` with the string being `"Search %s"` (in english).

I also believe that `getFeedTitle()` should be enough to create a String without the `Objects.toString()` call

daryanypl: default case:
"Search [feed title]".  If feed title is null, then hint should be: "Search feed".  If feed title is not null, and already contains the word "Search", then hint should be "Search [feed title]".

daryanypl:
Moved the string formatting into the internationalization file, and made a new issue to fix where the search hint is coming from (https://github.com/NYPL-Simplified/android/issues/587).

PR #581, opened 2018-04-30
ReaderActivity: Fix Warnings, refactor for Java 8 syntax, Replace nullness libraries for Android Studio

gioneill:
Update nullness module. Update nullcheck assertion module. Perform several linting hints with newer Java syntax. (lambdas, inline method names, re-indentation etc).

### DIFF
--- a/simplified-app-shared/src/main/java/org/nypl/simplified/app/reader/ReaderActivity.java
+++ b/simplified-app-shared/src/main/java/org/nypl/simplified/app/reader/ReaderActivity.java
@@ -11,9 +11,9 @@ import android.graphics.Color;
 import android.graphics.ColorMatrixColorFilter;
 import android.os.Build;
 import android.os.Bundle;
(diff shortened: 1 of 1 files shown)


gioneill:
@djds23 @daryanypl When you guys have a chance. Thanks!

simplified-app-shared/src/main/java/org/nypl/simplified/app/reader/ReaderActivity.java
```diff
@@ -27,9 +27,6 @@
 import com.io7m.jfunctional.Option;
 import com.io7m.jfunctional.OptionType;
 import com.io7m.jfunctional.Some;
-import com.io7m.jnull.NonNull;
-import com.io7m.jnull.NullCheck;
-import com.io7m.jnull.Nullable;
```
djds23: 🔥

PR #584, opened 2018-05-11
Use white accent color

winniequinn:
Fixes an issue where the cursor was the same color as the action bar.

It will now be white in all cases:

screenshot_1525880318


gioneill:
39944235-ebee079c-5533-11e8-91f7-3745989a9792

PR #614, opened 2018-08-23
Accounts.json updates

gioneill:
... for universal parsing between platforms by server generated file.

- Throw a crash on a failure to parse Accounts.json (required field).
- Handle production flag in entries.
- Handle base-64 string library logos.
- Handle migration of existing AccountsRegistry entries (saved "current accounts")


gioneill:
The updated `Accounts.json` is not in this yet until @winniequinn says whether this should be git-ignored or not..

PR #615, opened 2018-08-28
Address crash...

gioneill:
...if an Activity in the process of tearing down tries to present a dialog on the main thread. Can occur infrequently when testing and rapidly moving activities in and out.

### DIFF
--- a/simplified-app-shared/src/main/java/org/nypl/simplified/app/ReaderActivitySyncManager.kt
+++ b/simplified-app-shared/src/main/java/org/nypl/simplified/app/ReaderActivitySyncManager.kt
@@ -1,5 +1,6 @@
 package org.nypl.simplified.app
 
+import android.app.Activity
 import android.app.AlertDialog
 import android.content.Context
 import android.support.v4.content.ContextCompat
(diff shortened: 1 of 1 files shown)


simplified-app-shared/src/main/java/org/nypl/simplified/app/ReaderActivitySyncManager.kt
```diff
@@ -109,12 +110,14 @@ class ReaderSyncManager(private val feedEntry: OPDSAcquisitionFeedEntry,
     // 3 - The server mark came from the same device
     if (currentLocation.toString() != serverBookLocation.toString() &&
         device != serverLocation.body.device) {
-      UIThread.runOnUIThread {
-        alert.show()
-        val resID = ThemeMatcher.color(Simplified.getCurrentAccount().mainColor)
-        val mainTextColor = ContextCompat.getColor(context, resID)
-        alert.getButton(AlertDialog.BUTTON_NEGATIVE).setTextColor(mainTextColor)
-        alert.getButton(AlertDialog.BUTTON_POSITIVE).setTextColor(mainTextColor)
+      if ((context as? Activity)?.isFinishing == false) {
```
gioneill: Kotlin compares a "nullable boolean" with `true` `false` or `null`

PR #623, opened 2019-01-09
Moved in Accounts.json from private repository.

leonardr:
This branch adds a copy of Accounts.json as found in the private repository we used to keep it in. Keeping it in a public repository will allow partners to consult the most recent version of the file and to see a history of all changes made from this point forward.

This addresses https://jira.nypl.org/browse/SIMPLY-1612.


.gitignore
```diff
@@ -28,4 +28,4 @@ helpstack.conf
 
 # backup files
 *~
-Accounts.json
+
```
leonardr: If Accounts.json is supposed to live in a specific part of the path, I can just put it there, and you won't have to do anything special for android.

gioneill: @leonardr 
That sounds okay to me.
It would be in `/simplified-app-simplye/src/main/assets`

leonardr:
Great, I moved it.

PR #624, opened 2019-01-10
Implement automatic returns for audio books

io7m:
This change adds a dialog box that appears when the user listens to
the end of an audio book. If the user clicks the "return" button in
the dialog, the audio book loan is returned.

Fix: https://jira.nypl.org/browse/SIMPLY-391

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -19,7 +19,13 @@ ext {
    * The version of the NYPL Audiobook API used by this package.
    */
 
-  nypl_audiobook_api_version = "0.0.66"
+  nypl_audiobook_api_version = "0.0.67"
+
+  /*
+   * The version of the NYPL AudioEngine provider used by this package.
+   */
+
+  nypl_audiobook_findaway_version = "0.0.67"
 }
 
 allprojects {
--- a/simplified-app-shared/src/main/res/values/strings.xml
+++ b/simplified-app-shared/src/main/res/values/strings.xml
@@ -252,7 +252,8 @@
   <string name="audio_book_player">Audiobook Player</string>
   <string name="audio_book_player_error_engine_open">Unable to initialize audio engine</string>
   <string name="audio_book_player_error_book_open">Unable to open audio book</string>
-  <string name="audio_book_player_return">Return book?</string>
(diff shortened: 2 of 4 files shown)


build.gradle
```diff
@@ -19,7 +19,13 @@ ext {
...
    */
 
-  nypl_audiobook_api_version = "0.0.66"
+  nypl_audiobook_api_version = "0.0.67"
+
+  /*
+   * The version of the NYPL AudioEngine provider used by this package.
+   */
+
+  nypl_audiobook_findaway_version = "0.0.67"
```
gioneill: Why do I need to worry about both of these in the host app now?

simplified-app-shared/src/main/res/values/strings.xml
```diff
@@ -252,6 +252,9 @@
   <string name="audio_book_player">Audiobook Player</string>
   <string name="audio_book_player_error_engine_open">Unable to initialize audio engine</string>
   <string name="audio_book_player_error_book_open">Unable to open audio book</string>
+  <string name="audio_book_player_return">Return book?</string>
```
gioneill: Does this copy match the reference on the iOS implementation?

build.gradle
io7m: We've occasionally released new versions of the API package (the API package contains the views too, remember) and that obviously meant incrementing the version number of the API package. However, that has also then meant that I had to release a new version of the Findaway implementation, even though no code changed, just so that I could increment the version number of the Findaway package to match that of the API package. I split the versions in the demo app some time ago, and I thought it was probably time to do it in the actual app too.

simplified-app-shared/src/main/res/values/strings.xml
io7m: Oops. Can you point me in the direction of the right strings?

gioneill: Title:
Your Audiobook Has Finished
Message:
Would you like to return it?

io7m: Fixed in the latest commit.

gioneill:
👍 I'll let you hit the Merge button.

PR #625, opened 2019-01-16
Clarified the two different IDs we use for libraries

leonardr:
This branch addresses https://jira.nypl.org/browse/SIMPLY-1645. It removes the undescriptive `id` and `legacy_id` fields from `Accounts.json` and adds super-descriptive `numeric_id` and `uuid_id` fields to replace them. This will allow the iOS and Android apps to independently migrate towards using the UUID ID instead of the numeric ID. That's where we want to get, because the library registry only has UUID IDs.

Since running the Accounts.json generation script always pulls its data from the library registry, we've picked up a new library as well -- San Mateo County Public Library, in the 'testing' state.


simplified-app-simplye/src/main/assets/Accounts.json
```diff
@@ -128,326 +127,330 @@
...
-        "authPasscodeAllowsLetters": true,
+        "authPasscodeAllowsLetters": false,
         "authPasscodeLength": 0,
         "catalogUrl": "https://bibliomation.lyrasistechnology.org/ansonia/",
-        "id": 30,
         "inProduction": false,
-        "legacy_id": 30,
         "logo": null,
-        "mainColor": "cyan",
+        "mainColor": "blue",
```
leonardr: I've verified that these changes correspond to what Ansonia's authentication document says.

io7m:
@gioneill Said to go ahead with merging this.

PR #633, opened 2019-01-24
Getting and setting the current page's CFI using Readium's JS SDK.

EdwinGuzman:
This is for [JIRA ticket ](https://jira.nypl.org/browse/SIMPLY-1656). The js code here is the same as in the [Simplified-iOS](https://github.com/NYPL-Simplified/Simplified-iOS/pull/1027) PR.

Getting and setting the CFI is done in the `simplified.js` file but there's also minor changes in `ReaderActivity.java` which I'm not sure they are completely correct.
```
final ReaderSimplifiedJavaScriptAPIType simplified_js =
      Objects.requireNonNull(this.simplified_js_api);
```
The above was being called in separate functions but I wanted to use the same instance of the Simplified JS class so I added it as a class variable as `this.simplified_js`.

I had to set before and after the `js.setPageStyleSettings(s);` call. I noticed that there are java functions that call Readium's `bookmarkCurrent()` function multiple times and they receive the current CFI. But that location seems to be getting lost, specifically in `lazyInitSyncManagement` where the function throws an error (`Parameter(s) were unexpectedly null before creating Sync Manager. Abandoning attempt.`) and never goes to the `ReaderSyncManager`. This _could_ most likely be an issue with my set up.

As for the fix, this works better when decreasing the font than for increasing the font--similar to the iOS "fix".

### DIFF
--- a/simplified-app-shared/src/main/java/org/nypl/simplified/app/reader/ReaderActivity.java
+++ b/simplified-app-shared/src/main/java/org/nypl/simplified/app/reader/ReaderActivity.java
@@ -109,7 +109,6 @@ public final class ReaderActivity extends Activity implements
   private @Nullable Container                         epub_container;
   private @Nullable ReaderReadiumJavaScriptAPIType    readium_js_api;
   private @Nullable ReaderSimplifiedJavaScriptAPIType simplified_js_api;
-  private @Nullable ReaderSimplifiedJavaScriptAPIType simplified_js;
   private           ImageView                         view_bookmark;
   private @Nullable ViewGroup                         view_hud;
   private @Nullable ProgressBar                       view_loading;
@@ -433,10 +432,6 @@ public final class ReaderActivity extends Activity implements
 
     this.readium_js_api = ReaderReadiumJavaScriptAPI.newAPI(in_webview);
     this.simplified_js_api = ReaderSimplifiedJavaScriptAPI.newAPI(in_webview);
-
-    final ReaderSimplifiedJavaScriptAPIType simplified_js =
-      Objects.requireNonNull(this.simplified_js_api);
-    this.simplified_js = simplified_js;
 
     in_title_text.setText("");
 
@@ -789,13 +784,13 @@ public final class ReaderActivity extends Activity implements
 
     // Get the CFI from the ReadiumSDK before applying the new
     // page style settings.
-    this.simplified_js.getReadiumCFI();
+    this.simplified_js_api.getReadiumCFI();
 
     js.setPageStyleSettings(s);
 
     // Once they are applied, go to the CFI that is stored in the
     // JS ReadiumSDK instance.
-    this.simplified_js.setReadiumCFI();
+    this.simplified_js_api.setReadiumCFI();
 
     final ReaderColorScheme cs = s.getColorScheme();
     this.applyViewerColorScheme(cs);
@@ -1039,10 +1034,10 @@ public final class ReaderActivity extends Activity implements
         in_web_view.setVisibility(View.VISIBLE);
         in_progress_bar.setVisibility(View.VISIBLE);
         in_progress_text.setVisibility(View.VISIBLE);
-        this.simplified_js.pageHasChanged();
+        this.simplified_js_api.pageHasChanged();
       }, 200L);
     } else {
-      UIThread.runOnUIThread(this.simplified_js::pageHasChanged);
+      UIThread.runOnUIThread(this.simplified_js_api::pageHasChanged);
     }
   }
 
--- a/simplified-app-shared/src/main/java/org/nypl/simplified/app/reader/ReaderActivity.java
+++ b/simplified-app-shared/src/main/java/org/nypl/simplified/app/reader/ReaderActivity.java
@@ -109,6 +109,7 @@ public final class ReaderActivity extends Activity implements
   private @Nullable Container                         epub_container;
   private @Nullable ReaderReadiumJavaScriptAPIType    readium_js_api;
   private @Nullable ReaderSimplifiedJavaScriptAPIType simplified_js_api;
+  private @Nullable ReaderSimplifiedJavaScriptAPIType simplified_js;
   private           ImageView                         view_bookmark;
   private @Nullable ViewGroup                         view_hud;
   private @Nullable ProgressBar                       view_loading;
@@ -432,6 +433,10 @@ public final class ReaderActivity extends Activity implements
 
     this.readium_js_api = ReaderReadiumJavaScriptAPI.newAPI(in_webview);
     this.simplified_js_api = ReaderSimplifiedJavaScriptAPI.newAPI(in_webview);
+
+    final ReaderSimplifiedJavaScriptAPIType simplified_js =
+      Objects.requireNonNull(this.simplified_js_api);
+    this.simplified_js = simplified_js;
 
     in_title_text.setText("");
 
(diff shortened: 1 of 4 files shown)


simplified-app-shared/src/main/java/org/nypl/simplified/app/reader/ReaderActivity.java
```diff
@@ -433,6 +434,10 @@ private void makeInitialReadiumRequest(
     this.readium_js_api = ReaderReadiumJavaScriptAPI.newAPI(in_webview);
     this.simplified_js_api = ReaderSimplifiedJavaScriptAPI.newAPI(in_webview);
 
+    final ReaderSimplifiedJavaScriptAPIType simplified_js =
+      Objects.requireNonNull(this.simplified_js_api);
+    this.simplified_js = simplified_js;
+
```
io7m: Not too sure about this. Why is the separate but referentially equal `simplified_js` field necessary? The `simplified_js_api` field is assigned once in the activity's `onCreate` method and never reassigned (it would actually be declared `final` if Android allowed for that).

simplified-app-shared/src/main/java/org/nypl/simplified/app/reader/ReaderActivity.java
```diff
@@ -781,8 +786,17 @@ private void checkExistingBookmarksAndUpdateUI(
...
+    // Get the CFI from the ReadiumSDK before applying the new
+    // page style settings.
+    this.simplified_js.getReadiumCFI();
+
     js.setPageStyleSettings(s);
 
+    // Once they are applied, go to the CFI that is stored in the
+    // JS ReadiumSDK instance.
+    this.simplified_js.setReadiumCFI();
```
io7m: :+1: Straightforward!

simplified-app-shared/src/main/java/org/nypl/simplified/app/reader/ReaderActivity.java
EdwinGuzman: I wasn't sure about the syntax or if `Objects.requiredNonNull` was needed. So I can probably just use `this.simplified_js_api` for the calls? I'll try it.

io7m: > So I can probably just use this.simplified_js_api for the calls? I'll try it.

Yep, you should be able to.

EdwinGuzman: Thanks for the explanation. I need to get more acquainted with Java and now Kotlin!

gioneill:
I know this is going to be a pain, @EdwinGuzman , but I'd like to know that this change is measurably improving the user's experience.

I'd ask that you ready a large section of text, at a small text size, so that you can conceptually keep track of the "vicinity" of text that should remain close when moving. That way, when you change the size, or select a bookmark, you know if you've move a little bit or a lot. Right now, we've all just read a sentence or two, which won't look familiar, but we don't "size" of the movement.

Does that make sense?

EdwinGuzman:
Not at all (for it being a pain). It does make sense, but to clarify, you'd like to deeper analysis of the movement size from this fixed compared to how it was before?

simplified-app-shared/src/main/java/org/nypl/simplified/app/reader/ReaderActivity.java
```diff
@@ -1039,10 +1034,10 @@ Make the web view visible with a slight delay (as sometimes a
         in_web_view.setVisibility(View.VISIBLE);
         in_progress_bar.setVisibility(View.VISIBLE);
         in_progress_text.setVisibility(View.VISIBLE);
-        this.simplified_js.pageHasChanged();
+        this.simplified_js_api.pageHasChanged();
       }, 200L);
     } else {
-      UIThread.runOnUIThread(this.simplified_js::pageHasChanged);
+      UIThread.runOnUIThread(this.simplified_js_api::pageHasChanged);
```
io7m: :+1:

EdwinGuzman:
The following all done on a Pixel 2 emulator and done to compare the behavior on the `master` branch with the behavior on this branch. These are only two test cases but overall the page that loads when updating the font size is closer to the initial page/CFI value in this PR. When increasing font size, the page loaded was closer to the original CFI value. When decreasing the font size, the page loaded was closer as well and this saw the most improvement, since the behavior in `master` always seemed to take you back to the first page in the chapter. When rotating the view (in the Android emulator), there was no noticeable difference.

## 2br02b

### Font Size Increase
#### Master
Smallest font - 10 pages
Start at page 5 "'Skip it', he said."
-> Font increase on page 5
Goes to page 3 and "'Skip it', he said." is on page 7
-> Font increase on page 7
Goes to page 3 and "'Skip it', he said." is on page 10

#### PR
Smallest font - 10 pages
Start at page 5 "'Skip it', he said."
-> Font increase on page 5
Goes to page 5 and "'Skip it', he said." is on page 7
-> Font increase on page 7
Goes to page 7 and "'Skip it', he said." is on page 10

### Font Size Decrease
#### Master
Third biggest font - 31 pages
Start at page 13 "'Skip it', he said."
-> Decrease font on page 13
Goes to page 1 and "'Skip it', he said." is on page 10
-> Decrease font on page 10
Goes to page 1 and "'Skip it', he said." is on page 7

#### PR
Third biggest font - 31 pages
Start at page 13 "'Skip it', he said."
-> Decrease font on page 13
Goes to page 9 and "'Skip it', he said." is on page 10
-> Decrease font on page 10
Goes to page 6 and "'Skip it', he said." is on page 7

### Rotation
#### Master
Second smallest font size  - 16 pages
Start on page 7 with "'Skip it', he said."
Rotate from portrait to landscape:
Goes to page 11 (of 24) and "'Skip it', he said." is on page 10
Rotate from landscape to portrait:
Goes to page 8 and "'Skip it', he said." is on page 7

#### PR
Second smallest font size  - 10 pages
Start on page 5 with "'Skip it', he said."
Rotate from portrait to landscape:
Goes to page 6 (of 15) and "'Skip it', he said." is on page 7
Rotate from landscape to portrait:
Goes to page 4 and "'Skip it', he said." is on page 5

## I, Robot

### Font Size Increase
#### Master
Smallest font - 42 pages
Start on page 15 "Find a working phone and call 911..."
-> Font increase on page 15
Goes to page 10 and "Find a working phone and call 911..." is on page 24
-> Font increase on page 24
Goes to page 13 and "Find a working phone and call 911..." is on page 37

#### PR
Smallest font - 49 pages
Start on page 15 "Find a working phone and call 911..."
-> Font increase on page 15
Goes to page 10 and "Find a working phone and call 911..." is on page 24
-> Font increase on page 24
Goes to page 13 and "Find a working phone and call 911..." is on page 37

### Font Size Decrease
#### Master
Third biggest font (4th) - 165 pages
Start at page 46 "'Oi, stoppit,' the security guard said, ..."
-> Font decrease on page 46
Goes to page 1 and "'Oi, stoppit,' the security guard said, ..." is on page 33
-> font decrease on page 33
Goes to page 1 and "'Oi, stoppit,' the security guard said, ..." is on page 22

#### PR
Third biggest font (4th) - 165 pages
Start at page 46 "'Oi, stoppit,' the security guard said, ..."
-> Font decrease on page 46
Goes to page 33 and "'Oi, stoppit,' the security guard said, ..." is on page 33
-> font decrease on page 33
Goes to page 21 and "'Oi, stoppit,' the security guard said, ..." is on page 22

### Rotation
#### Master
Second smallest font size - 80 pages
Start on page 12 "'Detective?' he said."
Rotate from portrait to landscape:
Goes to page 15 (of 109) and "'Detective?' he said." is on the same page
Rotate from landscape to portrait:
Goes to page 11 and "'Detective?' he said." is on page 12

#### PR
Second smallest font size - 80 pages
Start on page 12 "'Detective?' he said."
Rotate from portrait to landscape:
Goes to page 15 (of 109) and "'Detective?' he said." is on the same page
Rotate from landscape to portrait:
Goes to page 11 and "'Detective?' he said." is on page 12

PR #635, opened 2019-02-01
Add two new libraries to the beta -- the State Library of Kansas and NYPL's QA server.

leonardr:
This addresses https://jira.nypl.org/browse/SIMPLY-1694. I've given NYPL's QA server its own registry entry, so that we can always access it in the beta independently of NYPL's production server.

Since any new libraries are picked up every time I run the script, this also picks up the State Library of Kansas for the next beta release.


gioneill:
I'm going to go ahead and merge this one. Thanks, Leonard.

PR #636, opened 2019-02-01
Removing a previous non-working way to inject fonts into the...

EdwinGuzman:
...webreader and loading it when the webreader loads. Updating the current font selected through javascript.

This is for ticket [SIMPLY-934](https://jira.nypl.org/browse/SIMPLY-934). This fixes the OpenDyslexic font loading issue but also allows both serif and sans-serif to update correctly.

I mostly removed java code that was calling `ReadiumSDK.reader.plugins.injectFonts.registerFontFace()` which is no longer part of the ReadiumSDK, whether from our fork version of the library or the latest version from Readium. To update this, I used the javascript code (`linkOpenDyslexicFonts`) from the iOS `simplified.js` file that loaded the OpenDyslexic font when a new page gets rendered. This worked fine.

It also appeared that `ReadiumSDK.reader.setBookStyles()` was being called and even though it correctly set the styles for the page (in the iframe's document), the font family was not updated. So I loosely based `updateFontFamily` off that code to update the font family for the body element in the iframe. This works fine.

What I'm not happy about is that now there appears to be two new bugs from this fix which I cannot always replicate but may _not_ be normal user flow.
* If I'm at the end of a chapter and I change the font (from OpenDyslexic to Serif, for example), I get sent to page 1 and there's only 1 page in this chapter. So that's completely not right.
* This may be a problem with the OpenDyslexic font itself: the end of chapters get cut off and it's not always consistent. Sometimes it appears one page is missing but at other times there may be more. I think this happens because the font size slightly increases to accommodate the OpenDyslexic font but the renderer doesn't seem to fix the total page size of the chapter. This I can replicate but how many pages get lost is different. The other two fonts have no issue which leads me to believe it's a specific OpenDyslexic font issue.

This fixes the loading issue but missing pages from the end of a chapter isn't ideal either. If you could check the flow of loading and updating the settings and see if I'm doing anything wrong, that'd be great.

### DIFF
--- a/simplified-app-shared/src/main/java/org/nypl/simplified/app/reader/ReaderActivity.java
+++ b/simplified-app-shared/src/main/java/org/nypl/simplified/app/reader/ReaderActivity.java
@@ -824,9 +824,6 @@ public final class ReaderActivity extends Activity implements
     final ReaderReadiumJavaScriptAPIType js =
       Objects.requireNonNull(this.readium_js_api);
 
-    // is this correct? inject fonts before book opens or after
-    js.injectFonts();
-
     /*
       If there's a bookmark for the current book, send a request to open the
       book to that specific page. Otherwise, start at the beginning.
--- a/simplified-app-shared/src/main/java/org/nypl/simplified/app/reader/ReaderReadiumJavaScriptAPI.java
+++ b/simplified-app-shared/src/main/java/org/nypl/simplified/app/reader/ReaderReadiumJavaScriptAPI.java
@@ -180,52 +180,22 @@ public final class ReaderReadiumJavaScriptAPI
     final ReaderSettingsType r)
   {
     try {
-      final ReaderColorScheme cs = r.getColorScheme();
-      final String color = NullCheck.notNull(
-        String.format("#%06x", cs.getForegroundColor() & 0xffffff));
-      final String background = NullCheck.notNull(
-        String.format("#%06x", cs.getBackgroundColor() & 0xffffff));
-
-      final JSONObject decls = new JSONObject();
-      decls.put("color", color);
-      decls.put("backgroundColor", background);
+      String fontSelected = "";
 
       switch (r.getFontFamily()) {
         case READER_FONT_SANS_SERIF: {
-          decls.put("font-family", "sans-serif");
+          fontSelected = "sans-serif";
           break;
         }
         case READER_FONT_OPEN_DYSLEXIC: {
-          /**
-           * This is defined as a custom CSS font family inside
-           * OpenDyslexic.css, which is referenced from the initially
-           * loaded reader.html file.
-           */
-
-          decls.put("font-family", "OpenDyslexic3");
+          fontSelected = "OpenDyslexic3";
           break;
         }
         case READER_FONT_SERIF: {
-          decls.put("font-family", "serif");
+          fontSelected = "serif";
           break;
         }
       }
-
-      final JSONObject o = new JSONObject();
-      o.put("selector", "*");
-      o.put("declarations", decls);
-
-      final JSONArray styles = new JSONArray();
-      styles.put(o);
-
-      final StringBuilder script = new StringBuilder(256);
-      script.append("ReadiumSDK.reader.setBookStyles(");
-      script.append(styles);
-      script.append("); ");
-      script.append("document.body.style.backgroundColor = \"");
-      script.append(background);
-      script.append("\";");
-      this.evaluate(script.toString());
 
       final ReaderReadiumViewerSettings vs = new ReaderReadiumViewerSettings(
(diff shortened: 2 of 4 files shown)


io7m:
I'm glad to see a lot of this unpleasant code being removed. :+1: 

To be clear though: Some of the code removed was code to set the foreground and background color. I'm always happy to see code go away, but was that intended? I remember you mentioning in a standup that the foreground/background colors didn't work properly, but does this mean that we can't reliably set them at all?

EdwinGuzman:
I forgot to mention that. Removing that piece of code did not affect the foreground/background color changes--sapia still works and the white text on black background is _still_ broken (text doesn't update to white). Not sure if it's an issue directly with this piece of code and I haven't investigated other functions were the color is updated (since I focused on the dyslexic font).

Something else, this seems to me like it's now relying slightly more on the custom simplified.js code rather than using the ReadiumSDK. If this is a problem, I could spend more time figuring out what is not working with their functions.

io7m:
I'm pretty sure the only place we set colors is in that `setPageStyleSettings` method. I can't find any other usages of the `ReaderColorScheme` type. I think we should throw the code away if its absence makes no difference!

EdwinGuzman:
From what I was trying to investigate, after the code calls `js.setPageStyleSettings`, the following function `this.applyViewerColorScheme(cs)` is the one that updates the theme. In that function, there's a `in_root.setBackgroundColor()` method that updates the background color (using `in_root.setBackgroundColor(cs.getBackgroundColor());`) but I can't find a function that's equivalent for the foreground color. Removing this code removes the ability to update the background color so I believe here's the spot to fix the broken white text color update.

simplified-app-shared/src/main/java/org/nypl/simplified/app/reader/ReaderReadiumJavaScriptAPI.java
```diff
@@ -180,53 +180,23 @@ private void evaluateWithResult(
...
-      o.put("declarations", decls);
-
-      final JSONArray styles = new JSONArray();
-      styles.put(o);
-
-      final StringBuilder script = new StringBuilder(256);
-      script.append("ReadiumSDK.reader.setBookStyles(");
-      script.append(styles);
-      script.append("); ");
-      script.append("document.body.style.backgroundColor = \"");
```
EdwinGuzman: I noticed that this *does* update the background color, but even when it's removed, the background color _still_ changes. So it seems that there's multiple pieces of code that does the same thing to the reader?

I also tried updating the foreground color using `document.body.style.color` but this does not work (for the white text on black background). The `setBookStyles` function on the ReadiumSDK does not actually update the background or foreground colors.

simplified-app-shared/src/main/java/org/nypl/simplified/app/reader/ReaderReadiumJavaScriptAPI.java
```diff
@@ -237,30 +235,10 @@ private void evaluateWithResult(
...
-      final StringBuilder script = new StringBuilder(256);
-      script.append("ReadiumSDK.reader.plugins.injectFonts.registerFontFace(");
-      script.append(o);
-      script.append(");");
-
-      this.evaluate(script.toString());
+      // Update the selected user font through the custom Simplified JS script.
+      this.evaluate(
+        NullCheck.notNull(
+          String.format("simplified.updateBookStyles({ 'font-family': '%1s', color: '%2s'});", fontSelected, color)));
```
EdwinGuzman: Now also updating the text color for books that did not previously update. Specifically, this is for the white text and black background theme.

simplified-app-shared/src/main/java/org/nypl/simplified/app/reader/ReaderActivity.java
```diff
@@ -824,9 +824,6 @@ private void checkExistingBookmarksAndUpdateUI(
     final ReaderReadiumJavaScriptAPIType js =
       Objects.requireNonNull(this.readium_js_api);
 
-    // is this correct? inject fonts before book opens or after
```
gioneill: 👍

simplified-app-shared/src/main/java/org/nypl/simplified/app/reader/ReaderReadiumJavaScriptAPI.java
```diff
@@ -182,35 +182,33 @@ private void evaluateWithResult(
     try {
       final ReaderColorScheme cs = r.getColorScheme();
       final String color = NullCheck.notNull(
-        String.format("#%06x", cs.getForegroundColor() & 0xffffff));
+          String.format("#%06x", cs.getForegroundColor() & 0xffffff));
```
gioneill: Looks like we have 2 space indenting, not 4!

PR #639, opened 2019-02-15
Change Accounts.json to explain the login mechanisms in the same terms used by the authentication document

leonardr:
This branch removes `pinRequired` and `authPasscodeAllowsLetters`, replacing them with `pinKeyboard`, which explains which keyboard (if any) to use during PIN entry. If `pinKeyboard` is `No input` then the library does not use a PIN.

This branch also adds `loginKeyboard`, which explains which keyboard to use during entry of the patron identifier.

The possible values are defined in our [extensions to the authentication document spec](https://github.com/NYPL-Simplified/Simplified/wiki/Authentication-For-OPDS-Extensions#keyboard).

These values come direct from the library's authentication documents.

No new libraries are in here -- there was one new library but it doesn't have its patron authentication set up yet, so i removed it.

In conjunction with the corresponding mobile work, this will make a fair amount of progress on https://jira.nypl.org/browse/SIMPLY-1463.


simplified-app-simplye/src/main/assets/Accounts.json
```diff
@@ -1746,16 +1797,16 @@
...
         "id_numeric": 52,
         "id_uuid": "urn:uuid:6d6a34d9-ba86-40b1-bf39-2a12d2f457d2",
         "inProduction": false,
+        "loginKeyboard": "Default",
         "logo": null,
         "mainColor": "blue",
         "name": "Massachusetts Library System",
         "needsAuth": true,
-        "pinRequired": false,
+        "pinKeyboard": "Default",
```
leonardr: This library switched from HTTP to HTTPS, and it now requires a PIN.

simplified-app-simplye/src/main/assets/Accounts.json
```diff
@@ -3087,17 +3150,16 @@
         "supportsSimplyESync": true
     },
     {
-        "authPasscodeAllowsLetters": false,
-        "authPasscodeLength": 0,
-        "catalogUrl": "http://circ.mls.minitex.net/WEYM/",
+        "catalogUrl": "https://circ.mls.minitex.net/WEYM/",
```
leonardr: This library switched from HTTP to HTTPS, and it now requires a PIN.

simplified-app-simplye/src/main/assets/Accounts.json
```diff
@@ -1357,17 +1397,17 @@
         "supportsSimplyESync": true
     },
     {
-        "authPasscodeAllowsLetters": true,
         "cardCreatorUrl": "https://archive.org/account/login.createaccount.php",
         "catalogUrl": "https://bookserver.archive.org/simple",
         "id_numeric": 5,
         "id_uuid": "urn:uuid:f60b644e-4955-4996-a4e5-a192feb4e7f8",
         "inProduction": false,
+        "loginKeyboard": "Email address",
```
leonardr: This is the motivating case behind this change -- the Internet Archive is telling us to use the email keyboard when asking people to log in.

PR #642, opened 2019-03-06
Added the New Fairfield library. Promoted Internet Archive to production.

leonardr:
These are the only changes I'm aware of from the past two weeks.


leonardr:
Actually there's one more potential change -- promoting Internet Archive from beta to production. I've got a request in asking if they want to do that now or wait for the next release.

PR #647, opened 2019-05-09
More production promotions

leonardr:
Going into a little more detail on this PR since it has a new reviewer.

This PR recreates the Accounts.json file used on both iOS and Android to construct the list of libraries shown to users. I do these periodically -- this is currently the only way libraries go from not being in SimplyE to being in SimplyE.

The list of libraries comes from the library registry's database. Pretty much all the information _about_ the libraries -- their logos, their patron support email addresses, their color scehemes, which keyboard to use when asking patrons to type in their password, etc. -- comes from the library's current authentication document.

So most of the changes in this diff are really small, but there are a couple really important changes, some big and some small.

The really important _small_ changes are the ones that change a library's "inProduction" key from false to true. This takes a library that used to only show up in beta releases of SimplyE, and makes it so that it _also_ shows up in the releases we push to the App Store.

In this case the two libraries promoted to production are Milford (in Connecticut) and Ruth Enlow (in Maryland).

This branch also removes three libraries hosted by Minitex which have either pulled out of the testing process, or are no longer being hosted by Minitex:  Nevins and Weymouth (Minnesota) and the Massachusetts state system. I confirmed these in conversation with Paul Swanson of Minitex.

Finally, this branch adds a new library to the testing system: the Open Textbooks feed at http://open.minitex.org/textbooks/. Previously this library was being added manually to some builds but not others. Including it in the library registry guarantees it will be in Accounts.json, eliminating the manual step.


simplified-app-simplye/src/main/assets/Accounts.json
```diff
@@ -2634,7 +2615,7 @@
         "mainColor": "blue",
         "name": "San Francisco Public Library",
         "needsAuth": true,
-        "pinKeyboard": "Number pad",
+        "pinKeyboard": "Default",
```
leonardr: Usually these libraries go from 'default' for the PIN keyboard to 'number pad'. San Francisco is going the other way; the original change was a mistake.

simplified-app-simplye/src/main/assets/Accounts.json
```diff
@@ -133,14 +133,15 @@
         "supportsSimplyESync": true
     },
     {
+        "authPasscodeAllowsLetters": true,
```
benanderman: I see this all throughout `Accounts.json`, so I'm sure it's not a problem, but I don't actually see anywhere in the code that uses it.

simplified-app-simplye/src/main/assets/Accounts.json
```diff
@@ -2279,6 +2240,26 @@
         "supportsReservations": true,
         "supportsSimplyESync": true
     },
+    {
+        "authPasscodeLength": 0,
+        "id_numeric": 1263,
+        "id_uuid": "urn:uuid:5278562c-d642-4fda-ad7e-1613077cfb8d",
```
benanderman: I don't see `id_uuid` used in the code either.

simplified-app-simplye/src/main/assets/Accounts.json
leonardr: I think this was recently replaced by the 'which keyboard should be used" code.

simplified-app-simplye/src/main/assets/Accounts.json
leonardr: That's a little worrying to me; I'm assuming you use the `id_numeric` instead?

The `id_uuid` is a replacement for `id_numeric`. These UUIDs are how the library registry identifies the libraries. I thought we switched over already but maybe it was only on Android.

If we haven't switched yet, there'll be a migration step for people who have data on their phones that's filed under the numeric IDs, and you'll need to keep the id_numeric -> id_uuid mapping around for quite a while, even when the rest of Accounts.json is gone, until we're sure that everyone is migrated.

benanderman: Yes, `id_numeric` is used instead, and you're right, we'll need that for a long time, since it's used to store user logins.

PR #650, opened 2019-05-31
Added Corona Public Library.

leonardr:
* Add a new library (Corona Public Library) directly into production
* Remove the PIN entry from four libraries: New Fairfield, Hartford, New Haven, and Ferguson
* Add a link to the authentication document to every library. This will let us write Android code that works as though it had access to the API endpoint, without first requiring that we actually start getting and parsing OPDS from that endpoint.


simplified-app-simplye/src/main/assets/Accounts.json
```diff
@@ -1,6 +1,7 @@
 [
     {
         "authPasscodeLength": 4,
+        "cardCreatorUrl": "https://patrons.librarysimplified.org/",
```
leonardr: We still need special client-side code to handle this URL by launching the native-code card creator rather than a web view.

https://jira.nypl.org/browse/SIMPLY-1997 covers the server-side work necessary to allow a custom URI scheme here.

io7m:
The build failed due to a not-forceful-enough `git push`. All the changes look good to me.

PR #651, opened 2019-07-16
SIMPLY-1970: Integrating PDF reader using page number as the bookmark (last read location)

MalcolmMcFly:


build.gradle
```diff
@@ -70,4 +70,14 @@ subprojects {
       }
     }
   }
+
+  // Android dependency has different version for the compile and runtime
+  project.configurations.all {
```
io7m: This is fine for now, but I think there was a reason we had to use newer builds tools than this. If this is introduced by Matt's dependency, then it needs to be fixed there so that this can be removed.

simplified-app-shared/src/main/java/org/nypl/simplified/app/catalog/CatalogBookReadController.kt
```diff
@@ -29,96 +31,114 @@ import org.slf4j.LoggerFactory
  */
 
 class CatalogBookReadController(
-  val activity: AppCompatActivity,
-  val analytics: AnalyticsType,
-  val profile: ProfileReadableType,
-  val account: AccountType,
-  val id: BookID,
-  val entry: FeedEntryOPDS) : OnClickListener {
+        val activity: AppCompatActivity,
```
io7m: I think we're mashing each other's whitespace. Let's sort out an editor config that we can all import so that we're all using the same settings. I don't have any particular preference for formatting, but the codebase is currently using two spaces instead of four, so that might be the path of least resistance.

simplified-app-shared/src/main/java/org/nypl/simplified/app/pdf/PdfReaderActivity.kt
```diff
@@ -0,0 +1,173 @@
...
+
+        this.profile = Simplified.getServices()
+                .profilesController
+                .profileCurrent()
+
+        this.account = profile.account(accountId)
+        this.books = account.bookDatabase
+        this.entry = books.entry(id)
+        this.handle = entry.findFormatHandle(BookDatabaseEntryFormatHandlePDF::class.java)!!
+        this.documentPageIndex = handle.format.lastReadLocation!!
```
io7m: Could do with some error checking here: The `findFormatHandle` method and `books.entry` method might return `null` or raise exceptions. It's unlikely, but this would crash the activity.

simplified-app-shared/src/main/java/org/nypl/simplified/app/pdf/PdfReaderParameters.kt
```diff
@@ -0,0 +1,20 @@
+package org.nypl.simplified.app.pdf
+
+import org.nypl.simplified.accounts.api.AccountID
+import org.nypl.simplified.books.api.BookID
+import java.io.File
+import java.io.Serializable
+
+/**
+ * Represents the parameters to pass to the [PdfReaderActivity].
+ * @param documentTtile Title of the PDF
```
io7m: Typo. :smile:

simplified-app-shared/src/main/java/org/nypl/simplified/app/pdf/PdfReaderActivity.kt
```diff
@@ -0,0 +1,180 @@
...
+
+            val b = Bundle()
+            b.putSerializable(this.PARAMS_ID, parameters)
+            val i = Intent(from, PdfReaderActivity::class.java)
+            i.putExtras(b)
+            from.startActivity(i)
+        }
+    }
+
+    private val log: Logger = LoggerFactory.getLogger(PdfReaderActivity::class.java)
```
io7m: Logger!

simplified-app-shared/src/main/java/org/nypl/simplified/app/pdf/PdfReaderActivity.kt
```diff
@@ -0,0 +1,180 @@
...
+        this.account = profile.account(accountId)
+        this.books = account.bookDatabase
+
+        try {
+            this.entry = books.entry(id)
+            this.handle = entry.findFormatHandle(BookDatabaseEntryFormatHandlePDF::class.java)!!
+            this.documentPageIndex = handle.format.lastReadLocation!!
+        } catch (e: BookDatabaseException) {
+            log.error("Could not get lastReadLocation, defaulting to the 1st page")
+            e.printStackTrace()
```
io7m: Two things here: The three expressions above the catch clause may throw any subtype of `Exception`. Consider what happens if `this.handle == null`...

Secondly, it's better to use the logger you declared above (the line I marked with "Logger!") as then the error would make it to Bugsnag.

```
this.logger.error("could not get last read location: ", e)
```

MalcolmMcFly: Good catch!

PR #652, opened 2019-07-24
Initial attempt to backport Readium improvements

io7m:
This attempts to port @EdwinGuzman 's old Readium improvements to
the current development branch. This seems to be working in that
OpenDyslexic is now usable again, and all color schemes appear to work
correctly. Some books still have a problem with the white-text-on-black
color scheme, but this appears to be a book-specific issue and may
not be fixable (I believe some books include CSS that can't be
overridden without specific knowledge of the book's stylesheets).

References: https://github.com/NYPL-Simplified/android/commit/57f661f976119d8dfc2e6071084f0945c7ba3c79
Affects: http://jira.nypl.org/browse/SIMPLY-1933

### DIFF
--- a/simplified-app-shared/src/main/java/org/nypl/simplified/app/reader/ReaderActivity.java
+++ b/simplified-app-shared/src/main/java/org/nypl/simplified/app/reader/ReaderActivity.java
@@ -760,13 +760,6 @@ public final class ReaderActivity extends ProfileTimeOutActivity implements
 
     p.setRootUrls(hs.getURIBase().toString(), null);
 
-    LOG.debug("onReadiumFunctionInitialize: scheduling font injection");
-    // is this correct? inject fonts before book opens or after
-    UIThread.runOnUIThreadDelayed(() -> {
-      LOG.debug("onReadiumFunctionInitialize: injecting fonts now");
-      this.readium_js_api.injectFonts();
-    }, 300L);
-
     /*
      * If there's a bookmark for the current book, send a request to open the
      * book to that specific page. Otherwise, start at the beginning.
--- a/simplified-app-shared/src/main/java/org/nypl/simplified/app/reader/ReaderColorSchemes.java
+++ b/simplified-app-shared/src/main/java/org/nypl/simplified/app/reader/ReaderColorSchemes.java
@@ -26,9 +26,9 @@ public final class ReaderColorSchemes {
       case SCHEME_BLACK_ON_BEIGE:
         return Color.argb(0xff, 242, 228, 203);
       case SCHEME_BLACK_ON_WHITE:
-        return Color.WHITE;
+        return 0xff000000 | (Color.WHITE & 0xffffff);
       case SCHEME_WHITE_ON_BLACK:
-        return Color.BLACK;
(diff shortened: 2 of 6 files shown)


simplified-app-shared/src/main/java/org/nypl/simplified/app/reader/ReaderActivity.java
```diff
@@ -760,13 +760,6 @@ public void onReadiumFunctionInitialize() {
 
     p.setRootUrls(hs.getURIBase().toString(), null);
 
-    LOG.debug("onReadiumFunctionInitialize: scheduling font injection");
-    // is this correct? inject fonts before book opens or after
-    UIThread.runOnUIThreadDelayed(() -> {
-      LOG.debug("onReadiumFunctionInitialize: injecting fonts now");
-      this.readium_js_api.injectFonts();
```
EdwinGuzman: This is not needed anymore? Did this ever work?

simplified-app-shared/src/main/java/org/nypl/simplified/app/reader/ReaderColorSchemes.java
```diff
@@ -26,9 +26,9 @@ public static int background(final ReaderColorScheme scheme)
       case SCHEME_BLACK_ON_BEIGE:
         return Color.argb(0xff, 242, 228, 203);
       case SCHEME_BLACK_ON_WHITE:
-        return Color.WHITE;
+        return 0xff000000 | (Color.WHITE & 0xffffff);
```
EdwinGuzman: What does the `& 0xffffff` do?

simplified-app-shared/src/main/java/org/nypl/simplified/app/reader/ReaderActivity.java
io7m: It was actually your old patch that removed the `injectFonts()` method. The code seems to work without it (OpenDyslexic is usable here).

simplified-app-shared/src/main/java/org/nypl/simplified/app/reader/ReaderColorSchemes.java
io7m: The values returned from this method have to be packed AARRGGBB integers to be consumed by the Android API. I think some of the constants are defined with 0 in the alpha channel, so that expression masks off the RGB bits and then combines them with a value that represents a fully-opaque alpha channel. This was necessary to avoid the header and footer in the reader appearing to be black.

io7m:
> Looks good. Are you okay with injecting/loading fonts more on the reader/js side than the server/java side?

Whatever works! This is the best Readium behaviour I've seen to date, so presumably this is the way we want to go...

simplified-app-shared/src/main/java/org/nypl/simplified/app/reader/ReaderActivity.java
EdwinGuzman: Ah right, I guess I missed this call for the function.

danielweck:
Regarding the OpenDyslexic font, this may be related:
https://github.com/readium/readium-css/issues/56

PR #653, opened 2019-08-09
SIMPY-2195: Check-Style/Linter Upgrades on Mobile

MalcolmMcFly:


build.gradle
```diff
@@ -71,3 +71,11 @@ subprojects {
     }
   }
 }
+
+task installGitHook(type: Copy) {
+  from new File(rootProject.rootDir, 'pre-commit')
+  into { new File(rootProject.rootDir, '.git/hooks') }
+  fileMode 0777
+}
```
io7m: This makes me a little nervous. I'm not hugely enthusiastic about Gradle playing with version control, although I can see why it's necessary. I would be happier if there was a check in place to see if the `pre-commit` hook needed to be added rather than just overwriting whatever is already there.

Additionally: Mode `0777` is a security vulnerability. The `7` at the end means that any other user on the system can overwrite this file with arbitrary code and have you execute it under your uid next time you commit. I'd expect mode `0755` to be sufficient (or perhaps even just `0700`).

README.md
```diff
@@ -83,3 +83,32 @@ for documentation on how to customize branding of the application.
...
 
 After setup is complete, the project can be opened in Android Studio and built as normal.
+
+# Ktlint
+
+SimplyE now has Ktlint enabled. 
+
+**What does this mean?**
+
+This means you no longer get away with committing less than stellar styled code anymore. Anytime you try to commit, your code will be checked for coolness (good style). If your code is cool then nothing different happens, its committed like always. If it is not cool (bad style), then you get an angry *commit failed* message that forces you to correct your style issues before the commit is accepted. Ktlint will provide you with the exact line and directions on exactly how to fix your style issue. Example:
```
io7m: its -> it's

:stuck_out_tongue:

simplified-app-simplye/build.gradle
```diff
@@ -1,5 +1,6 @@
 apply plugin: 'com.android.application'
 apply plugin: 'kotlin-android'
+apply from: '../ktlint.gradle'
```
io7m: Is this actually the correct module to which linting should be applied?

I would expect the bulk of linting to be applied in all of the other modules; `simplified-app-simplye` contains very little actual source code.

README.md
```diff
@@ -83,3 +83,32 @@ for documentation on how to customize branding of the application.
...
+
+**ktlint**
+```
+simplified-app-simplye/src/main/java/org/nypl/simplified/simplye/SimplyEAccountFallback.kt:20:1: Unexpected indentation (6) (it should be 8) (cannot be auto-corrected)
+```
+
+ **I'll feel bad if my git yells at me for bad style**
+ 
+ Luckily you can always check your own style before committing attempting to commit your code. Simple run the following command from project root:
```
io7m: Simple -> Simply

:stuck_out_tongue:

io7m: > Luckily you can always check your own style before committing attempting to commit your code

Can't quite parse that sentence. Think maybe "committing" isn't supposed to be there.

pre-commit
```diff
@@ -0,0 +1,15 @@
...
+
+echo "Running git pre-commit hook"
+
+./gradlew ktlint
+
+RESULT=$?
+
+git stash pop -q
+
+# return 1 exit code if running checks fails
```
io7m: Maybe just `exit $RESULT`

PR #656, opened 2019-08-30
SIMPLY-2260: Integrate Instabug for QA Builds

MalcolmMcFly:
For some reason can't get this to build, now debugging


simplified-app-shared/src/main/java/org/nypl/simplified/app/Simplified.kt
```diff
@@ -17,6 +19,8 @@ import org.nypl.simplified.observable.ObservableReadableType
 import org.nypl.simplified.prefs.Prefs
 import org.slf4j.LoggerFactory
 import java.io.File
+import java.io.IOException
+import java.util.*
```
io7m: :nauseated_face:

MalcolmMcFly: wildcard import?

simplified-app-shared/build.gradle
```diff
@@ -113,6 +113,7 @@ dependencies {
   api libraries.readiumSharedJS
   api libraries.slf4j
   api libraries.viewPagerIndicator
+  api libraries.instabug
```
io7m: I'm OK with this for now, given that there's no API, but we will need to hide the setup behind an interface when we do the repository separation in order to make it into an optional dependency. The reason being: Declaring this in the `app-shared` module means that all app frontends that depend on the app-shared module _must_ now use Instabug unconditionally. LFA in particular won't be able to do this for many deployments.

simplified-app-shared/src/main/java/org/nypl/simplified/app/Simplified.kt
io7m: Android Studio is great at sneaking in those star imports...

simplified-app-simplye/src/main/assets/instabug.conf
```diff
@@ -0,0 +1 @@
+instabug.token = 0fb114d5c17b376f626847515f55c162
```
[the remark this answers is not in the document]
MalcolmMcFly: oh god, I thought I added this to the `.gitignore` file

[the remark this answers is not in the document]
io7m: Maybe you ignored it after it was already added to the index... Git will helpfully commit anyway in that situation.

simplified-app-shared/src/main/java/org/nypl/simplified/app/services/SimplifiedServices.kt
```diff
@@ -32,12 +34,7 @@ import org.nypl.simplified.analytics.api.Analytics
 import org.nypl.simplified.analytics.api.AnalyticsConfiguration
 import org.nypl.simplified.analytics.api.AnalyticsEvent
 import org.nypl.simplified.analytics.api.AnalyticsType
-import org.nypl.simplified.app.AdobeDRMServices
-import org.nypl.simplified.app.Bugsnag
-import org.nypl.simplified.app.BundledContentResolver
-import org.nypl.simplified.app.NetworkConnectivityType
-import org.nypl.simplified.app.R
-import org.nypl.simplified.app.ScreenSizeInformationType
+import org.nypl.simplified.app.*
```
io7m: :nauseated_face:

MalcolmMcFly: How does this keep getting added! I need to set my IDE to stop doing this.

io7m: I use this setting:

settings

However, even that doesn't fully prevent it. The IDE will still insert start imports, it's just that the code formatter will clean them up afterwards if you ask it to. It seems to be specific to Kotlin in the IDE, doesn't happen in Java.

io7m: Will that linting tool in your other PR catch these? I'd be surprised if there wasn't a feature to disallow them.

MalcolmMcFly: Ktlint will catch these, and I will use this setting too

PR #657, opened 2019-09-11
Refactor link handling and provider categorization

io7m:
This refactors the code to use a single sealed Link type
everywhere that safely indicates whether or not a given link is
templated. Previously a templated link just caused an error due to
templated links not being valid URIs. This was necessary because the
NYPL registry is now returning templated links, and it wasn't possible
to properly diagnose SIMPLY-2285.
    
This also fixes SIMPLY-2285, which turned out to be a simple issue of
referring to the wrong JSON object when trying to parse isProduction
and isAutomatic flags from cached provider descriptions.
    
Fix: https://jira.nypl.org/browse/SIMPLY-2285


io7m:
I've just realized that "Update README" shouldn't have been in the PR, it should have already been added to `develop`. I was working from a slightly out-of-date working copy. It's harmless, so I don't think it's worth redoing the PR.

PR #659, opened 2019-09-13
Adds in the `PdfReaderActivity` to the manifest of the vanilla build.

mattniehoff:
Adds a reference to `PdfReaderActivity` to the vanilla flavor's `AndroidManifest.xml`. This fixes the issue where attempting to open a PDF crashes in the app.


io7m:
Embarassing that we left this one out... :sweat_smile:

PR #663, opened 2019-12-18
Updated Kotlin, build tools, and enabled Readium logging

MalcolmMcFly:
Simple changes.


kyles-ep:
I can only really speak to the readium line, which is correct

PR #664, opened 2020-01-07
Bump ktlint version to 0.36.0; move ktlint tasks to "quality" group.

twaddington:
Just a quick little change I made as I was poking around the project.

I moved the ktlint tasks to a new "quality" group. This is something I've done on other projects and is kinda nice because they're easier to spot when you run `./gradlew tasks`.

Happy to revert the group change and just do a version bump if y'all prefer it as-is.

### DIFF
--- a/ktlint.gradle
+++ b/ktlint.gradle
@@ -5,15 +5,15 @@ configurations {
     ktlint
 }
 dependencies {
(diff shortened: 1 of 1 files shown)


twaddington:
Had an issue with Travis. Reopening as https://github.com/NYPL-Simplified/android/pull/665

PR #667, opened 2020-01-07
Show app version in main settings fragment.

twaddington:
Just a little change to show the app version on the main settings fragment.

- [Screenshot_1578430635](https://user-images.githubusercontent.com/107117/71929099-c2489180-314d-11ea-8a90-bd017b7ba52d.png)

### DIFF
--- a/simplified-ui-settings/src/main/java/org/nypl/simplified/ui/settings/SettingsFragmentMain.kt
+++ b/simplified-ui-settings/src/main/java/org/nypl/simplified/ui/settings/SettingsFragmentMain.kt
@@ -1,5 +1,6 @@
 package org.nypl.simplified.ui.settings
 
+import android.content.pm.PackageManager
 import android.os.Bundle
 import androidx.preference.Preference
 import androidx.preference.PreferenceFragmentCompat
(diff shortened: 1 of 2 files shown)


simplified-ui-settings/src/main/java/org/nypl/simplified/ui/settings/SettingsFragmentMain.kt
```diff
@@ -87,6 +88,14 @@ class SettingsFragmentMain : PreferenceFragmentCompat() {
   }
 
   private fun configureVersion(preference: Preference) {
+    try {
+      val context = requireContext()
+      val pkgManager = context.packageManager
+      val pkgInfo = pkgManager.getPackageInfo(context.packageName, 0)
+      preference.summary = "${pkgInfo.versionName} (${pkgInfo.versionCode})"
+    } catch (e: PackageManager.NameNotFoundException) {
+      //  Pass
```
twaddington: If an exception is thrown the version won't be shown (no different than the existing behavior).

io7m:
One minor nitpick here: The other settings items use `...` to indicate that clicking the item will take you to another page. This would lose that indicator, making it inconsistent with the other items. I'm not sure adding `...` to the end of the version number would look particularly pleasant either.

I think I'm probably bikeshedding.

PR #669, opened 2020-01-08
Resolves a good chunk of the compiler warnings

twaddington:
Went through and resolved a good chunk of the compiler warnings. I opened the app and poked around and things seem to work okay but will require some testing.


simplified-viewer-audiobook/src/main/java/org/nypl/simplified/viewer/audiobook/AudioBookViewer.kt
```diff
@@ -38,17 +38,7 @@ class AudioBookViewer : ViewerProviderType {
...
-      from = activity,
-      parameters = AudioBookPlayerParameters(
-        formatAudio.manifest!!.manifestFile,
-        formatAudio.manifest!!.manifestURI,
-        book.id,
-        book.entry,
-        TODO()
-      )
-    )
+    TODO()
```
io7m: Erk!

This is definitely wrong (as in the code was wrong before you replaced it with a single `TODO()` call). I need to go back and fill in the missing parameter, otherwise loading an audio book is going to result in a crash. :blush:

io7m:
This all seems fine except for that one audiobook-related `TODO()`. If you want, you can take a shot at working out what the original `TODO()` parameter would have been. :wink: 

Otherwise, I should be able to find the time to correct this today.

twaddington:
@io7m I can take a crack at correcting that method, if you haven't already started.

io7m:
> @io7m I can take a crack at correcting that method, if you haven't already started.

Perhaps treat it as a learning exercise!

io7m:
This looks good to me. I'm glad they finally resolved the incremental annotation processor issue. That warning has been there for over a year...

PR #670, opened 2020-01-08
Add ktlint tasks to all subprojects

twaddington:
The ktlint tasks were not being added to subprojects, so wasn't throwing any warnings when run.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -142,6 +142,12 @@ allprojects {
   version = project.ext["VERSION_NAME"]
 }
 
+// Add ktlint tasks to all subprojects
+
+subprojects { project ->
+  apply from: file("$rootDir/ktlint.gradle")
+}
+
 subprojects { project ->
 
   /*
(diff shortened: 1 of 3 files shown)


io7m:
Looks good. I assume you've run ktlint on all the modules? If so... How are we looking in terms of warnings/errors? Malcolm originally tried adding it to all modules and there were too many errors to deal with up front, so we started incrementally adding ktlint where we could. That was a while ago, though...

twaddington:
> How are we looking in terms of warnings/errors? Malcolm originally tried adding it to all modules and there were too many errors to deal with up front, so we started incrementally adding ktlint where we could. &mdash; @io7m 

I ran ktlint on all modules and there are still a lot of warnings. I was gonna take a crack at addressing some of those today.

It doesn't appear that we fail the build on ktlint errors, so I think adding the task should be fine. We can run ktlint against a single module, or all modules, depending on our needs.

io7m:
Let's go!

PR #671, opened 2020-01-08
Print build messages related to artifact signing only when relevant.

twaddington:
Previously, log messages related to artifact signing (PGP/GPG) would be
printed immediately on project configuration.

This change causes these messages to only be printed before a signing task
will actually be executed. For example, when executing the task
'publishBasicJarPublicationToMavenLocal'.

This makes builds less chatty.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -151,18 +151,6 @@ subprojects { project ->
 
 subprojects { project ->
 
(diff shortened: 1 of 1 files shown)


io7m:
Looks good!

PR #672, opened 2020-01-10
Fixed ktlint warnings; junit.framework.Assert is deprecated.

twaddington:
Ran ktlintFormat to resolve a bunch of issues and manually verified by reviewing the diff. Had to fix a couple things it wasn't able to resolve automatically.

We should have a clean ktlint report after this is merged.


simplified-accounts-api/src/main/java/org/nypl/simplified/accounts/api/AccountAuthenticationAdobePostActivationCredentials.kt
```diff
@@ -27,5 +27,5 @@ data class AccountAuthenticationAdobePostActivationCredentials(
    * @return The device token
    */
 
-  val userID: AdobeUserID)
-
+  val userID: AdobeUserID
+)
```
io7m: I'm glad we're switching to this paren style everywhere, it'd been bugging me.

simplified-books-controller/src/main/java/org/nypl/simplified/books/controller/BookBorrowTask.kt
```diff
@@ -35,13 +35,26 @@ import org.nypl.simplified.books.book_database.api.BookFormats
 import org.nypl.simplified.books.book_registry.BookRegistryType
 import org.nypl.simplified.books.book_registry.BookStatus
 import org.nypl.simplified.books.book_registry.BookStatusDownloadErrorDetails
-import org.nypl.simplified.books.book_registry.BookStatusDownloadErrorDetails.*
```
io7m: Grr, these star imports need to die. Even with Android Studio configured never to add them, they still sometimes get added automatically.

io7m:
Looks good to me!

What I'd like to do is reinstate the "require running ktlint" aspect that @MalcolmMcFly introduced, but not do it the way we did it originally (a commit hook). The interaction between having a long-running process, `git stash`, and a commit hook, lead to problems.

Could we maybe make `ktlint` run very early in the build (on all modules), prior to compiling any sources. Perhaps with a command-line property to skip running it (for convenience in IDE builds).

simplified-books-controller/src/main/java/org/nypl/simplified/books/controller/BookBorrowTask.kt
twaddington: Yeah, these can be tricky. We'd had a flag set disable the "star-imports" check but I thought I'd take a shot at removing it.

twaddington:
> Could we maybe make ktlint run very early in the build (on all modules), prior to compiling any sources? Perhaps with a command-line property to skip running it (for convenience in IDE builds). &mdash; @io7m 

Typically what I'll do is hook ktlint into the "test" task. I can submit a follow-up for that change.

PR #674, opened 2020-01-11
Refactored Travis scripts.

twaddington:


io7m:
I'm fine with this.

I was considering adding a Gradle task to manually increment the version code, and have the Travis CI scripts call that (passing in something morally equivalent to `TRAVIS_BUILD_NUMBER` as an increment). I think I might defer that work until I actually find myself needing it, though. :smile:

twaddington:
@io7m thanks for taking a look! Yeah, I suspect we'll want to do more work on this in the near future. I think it'd be helpful to know what our new workflow will be for distributing internal builds though. I know Malcolm is interested in using Firebase's app distribution.

With more details we should be able to simplify the publish script considerably.

killerpuppytails:
FYI I'm hoping to have the Firebase privacy assessment done by the end of the week.

PR #677, opened 2020-01-21
Refactored catalog feed lanes to use a horizontally scrolling RecyclerView.

twaddington:


simplified-books-covers/src/main/java/org/nypl/simplified/books/covers/BookCoverProvider.kt
```diff
@@ -97,31 +97,31 @@ class BookCoverProvider private constructor(
...
+        .into(imageView, fallbackToGeneration)
     } else {
       this.log.debug("{}: {}: loading generated uri {}", tag, entry.bookID, uriGenerated)
 
-      val request = this.picasso.load(uriGenerated.toString())
-      request.tag(tag)
-      request.resize(width, height)
-      request.transform(badgePainter)
-      request.into(imageView, callbackFinal)
-      request.noFade()
```
twaddington: Turns out `request.noFade()` wasn't working because it was being applied after the call to `request.into(...)`.

Chaining these calls uncovered this error.

simplified-tests-sandbox/src/main/AndroidManifest.xml
```diff
@@ -51,11 +46,6 @@
       android:name="org.nypl.simplified.tests.sandbox.ErrorPageActivity"
       android:exported="true"
       android:theme="@style/SimplifiedTheme_ActionBar_DeepPurple" />
-
-    <activity
```
twaddington: This activity doesn't exist so I removed the manifest reference.

simplified-tests-sandbox/src/main/AndroidManifest.xml
```diff
@@ -32,11 +32,6 @@
       android:exported="true"
       android:theme="@style/SimplifiedTheme_ActionBar_DeepPurple" />
 
-    <activity
```
twaddington: This activity seemed dependent on the HorizontalScrollView implementation, and wasn't used anywhere, so I removed it.

twaddington:
> Any chance you can show me what this looks like? &mdash; @io7m 

Yep! The error state [looks like this](https://user-images.githubusercontent.com/107117/72852177-a8856f00-3c62-11ea-93c1-a8af0d319f53.png). It's pretty basic.

I see the 10PRINT generated covers. You're correct, that should take precedence. This is just a fallback for the fallback.

io7m:
I like it!

io7m:
I'm mostly in favour of the changes, having tried the build. Couple of small things:

1. I'm not entirely convinced of the aesthetics of the "downloading" icons. This is purely a visual thing; I think they're a bit dark for the app's current theme. That's obviously easy to adjust.

2. I'm seeing an odd delay before generated covers appear. I get a blank white cover until all the others are finished loading. See the very last page in the video. This is on a real device (Samsung Galaxy A6).

https://ataxia.io7m.com/2020/01/22/device-2020-01-22-130014.mp4

twaddington:
Thanks for the feedback @io7m.

1. Let me look into tweaking that visually today. It's best to have a placeholder, otherwise the covers kind of pop in out of nowhere and can shift around.
2. I see what you mean about the white cover before the generated covers appear. Let me look into that. I hadn't seen it in my testing.

io7m:
Oh yeah, don't get me wrong - the placeholders are necessary. Think the current shade tends to draw the eye more than the actual cover images in some cases though, thanks to the light background.

twaddington:
Got it! I'll try to lighten them up a bit.

twaddington:
@io7m I fixed the delay you were seeing with the generated covers. I had forgotten to set the placeholder image on that request, so when it started it reverted to a blank image.

I lightened the placeholder images up a bit and I think it looks better. What do you think?

- [Screenshot_1579759198](https://user-images.githubusercontent.com/107117/72960453-a1de2100-3d62-11ea-85d8-03d95895a399.png)

simplified-main/src/main/java/org/nypl/simplified/main/MainServices.kt
```diff
@@ -575,7 +575,7 @@ internal object MainServices {
     badgeLookup: BookCoverBadgeLookupType
   ): BookCoverProviderType {
     val execCovers =
-      NamedThreadPools.namedThreadPool(1, "cover", 19)
+      NamedThreadPools.namedThreadPool(2, "cover", 19)
```
twaddington: This makes everything feel faster since the images are loading less sequentially.

I think 4 is the default but thought 2 was a sensible place to start.

io7m:
> I lightened the placeholder images up a bit and I think it looks better. What do you think?

Looks good to me! Amazing what a difference a few percentage points of brightness makes.

PR #680, opened 2020-02-04
SIMPLY-2529: Fixed the aspect ratio of catalog images.

twaddington:
Previously, catalog images were stretched or squished to fit in a fixed
size image view. This change updates the images to have a fixed height
with a variable width. This means covers will maintain their aspect
ratio. As a result, covers appear much sharper and more readable.

### Screens

- [covers-before](https://user-images.githubusercontent.com/107117/73719817-5d914000-46d5-11ea-895f-9d83127b91c1.png)
- [covers-after](https://user-images.githubusercontent.com/107117/73719818-5d914000-46d5-11ea-92b6-81f05f5aab4c.png)

### DIFF
--- a/simplified-books-covers/src/main/java/org/nypl/simplified/books/covers/BookCoverGenerator.java
+++ b/simplified-books-covers/src/main/java/org/nypl/simplified/books/covers/BookCoverGenerator.java
@@ -1,10 +1,6 @@
 package org.nypl.simplified.books.covers;
 
 import android.graphics.Bitmap;
-import android.graphics.Bitmap.Config;
-import android.graphics.Canvas;
-import android.graphics.Color;
-import android.graphics.Paint;
 
 import com.io7m.jnull.NullCheck;
 
@@ -67,8 +63,8 @@ public final class BookCoverGenerator implements BookCoverGeneratorType
 
   @Override public Bitmap generateImage(
     final URI u,
-    final int width,
-    final int height)
+    int width,
+    int height)
     throws IOException
   {
     try {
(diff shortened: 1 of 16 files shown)


simplified-ui-catalog/src/main/res/layout/feed_with_groups.xml
```diff
@@ -14,9 +13,10 @@
     android:id="@+id/feedWithGroupsList"
     android:layout_width="0dp"
     android:layout_height="0dp"
+    android:clipChildren="false"
+    android:clipToPadding="false"
```
twaddington: This prevents the CardView drop shadow from being clipped.

simplified-books-covers/src/main/java/org/nypl/simplified/books/covers/BookCoverGenerator.java
```diff
@@ -81,19 +77,20 @@ public BookCoverGenerator(
       final String author_maybe = params.get("author");
       final String author = author_maybe == null ? "" : NullCheck.notNull(author_maybe);
 
+      if (width == 0) {
+        width = (int) Math.round(height * .75);
```
twaddington: The width is now variable, so may not be known in advance. The code below expects the width to be greater than zero. Since we're generating covers we can give it any width we want. Calculate a reasonable width from the given height instead.

simplified-books-covers/src/main/java/org/nypl/simplified/books/covers/BookCoverGenerator.java
```diff
@@ -81,19 +77,20 @@ public BookCoverGenerator(
...
+        height = (int) Math.round(width / .75);
+      }
+
       final TenPrintInputBuilderType ib = TenPrintInput.newBuilder();
       ib.setAuthor(author);
       ib.setTitle(title);
       ib.setCoverHeight(height);
       final TenPrintInput i = ib.build();
       final Bitmap cover = this.generator.generate(i);
-      final Bitmap container = Bitmap.createBitmap(width, height, Config.RGB_565);
```
twaddington: This block of code was adding white margins to the sides of the generated cover images. This looked weird inside the CardView so I removed it. Happy to consider other suggestions.

### Screens

- [Before](https://user-images.githubusercontent.com/107117/73720160-2c653f80-46d6-11ea-8940-5225cc3f5144.png)
- [After](https://user-images.githubusercontent.com/107117/73720162-2c653f80-46d6-11ea-9ff7-0940c5957240.png)

io7m: I forget what the purpose of this was. Looks like it was just there to ensure the image had the expected size, and was in `RGB_565`.

io7m:
:+1:

simplified-books-covers/src/main/java/org/nypl/simplified/books/covers/BookCoverGenerator.java
io7m: Actually, I've got a feeling that this was a leftover from having to fight with a `HorizontalScrollView` inside a `ListView`. Thankfully we no longer have that problem!

twaddington: Makes sense. I figured it was just adding some padding to the sides of the image.

PR #682, opened 2020-02-07
Removed Bugsnag

MalcolmMcFly:
**What's this do?**
The PR removes Bugsnag .

Since Bugsnag was removed from SimplyE, it should also be removed from core.

No tests.

Ensure that this built with the version of SimplyE that has Bugsnag removed.

In Readme I have removed all references of Bugsnag.


io7m:
It's gone! :tada:

MalcolmMcFly:
Screen Shot 2020-02-19 at 7 53 16 PM

PR #683, opened 2020-02-14
Enable "Related books..." feeds

io7m:
**What's this do?**
This adds the missing code necessary to show and use Related Books
feeds in book detail views.

Fixes: https://jira.nypl.org/browse/SIMPLY-2565

Navigate to the book detail view of any book in the catalog that you know has related books in the collection. Observe that the "Related books..." section in the detail view is no longer greyed-out. Click the text to go to the related books feed.

None.

Not applicable.

@io7m


simplified-ui-catalog/src/main/java/org/nypl/simplified/ui/catalog/CatalogFragmentBookDetail.kt
```diff
@@ -372,6 +373,47 @@ class CatalogFragmentBookDetail : Fragment() {
...
+    val targetFeed =
+      feedModel.resolveFeed(
+        title = context.resources.getString(R.string.catalogRelatedBooks),
+        uri = feedRelated,
+        isSearchResults = false
+      )
+    this.findNavigationController().openFeed(targetFeed)
+  }
+
+  private fun createOrGetFeedModel(
```
io7m: Minor technical debt here. This method is declared twice over the whole codebase.

PR #684, opened 2020-02-20
SIMPLY-2567: Show reading progress in Readium 2 viewer.

twaddington:
**What's this do?**
Show reading progress in Readium 2 viewer.

https://jira.nypl.org/browse/SIMPLY-2568

Depends on https://github.com/NYPL-Simplified/Simplified-R2-Android/pull/2.

**Screens**
[Screenshot_1582188868](https://user-images.githubusercontent.com/107117/74918291-ba4f5480-537d-11ea-9216-7a45187677b8.png)


io7m:
I've no problem with the actual content of the code, but shouldn't this stuff be in the other project? I would have expected to see this as part of the R2 views module.

twaddington:
> I've no problem with the actual content of the code, but shouldn't this stuff be in the other project? &mdash; @io7m 

It could, yes. We could integrate it into the `SR2ReaderFragment` directly. I hadn't been thinking about it that way but it would be fairly simple to move.

PR #688, opened 2020-03-06
SIMPLY-2567: Don't activate leanback mode if the screen reader is enabled.

twaddington:
This changes the behavior of the new reader to show a persistent toolbar
and system navigation if the screen reader (TalkBack) is enabled.
Previously, leanback mode would activate automatically putting the
reader into a fullscreen view. This can make it hard for users to
navigate when using accessibility services.


io7m:
Looks good!

PR #692, opened 2020-03-16
Feature/audiobook 4.0.0

io7m:
**What's this do?**
This makes a number of changes necessary to use the AudioBook 4.0.0 API. It is currently able to fulfill and play DPLA audio books, but a few more changes will be required in order to play OD audiobooks.

To fulfill these two tickets:
https://jira.nypl.org/browse/SIMPLY-2301
https://jira.nypl.org/browse/SIMPLY-1732

We should test at least one existing RBDigital audio book, and at least one Findaway audio book to check for regressions. We should then test the books available in the DPLA test collection. Testing, in this case, means borrowing books and playing them.

The changes to handle the new audio book content types involved replacing our slightly archaic MIME type handling with a more modern approach using an external library (which I actually maintain myself). There could be regressions here with regards to borrowing, although the test suite didn't indicate any problems.

Not applicable.

@io7m


io7m:
I've borrowed and played both DPLA audio books in the test collection with this.

io7m:
Currently we depend on a snapshot version of the AudioBook API. I don't forsee API-incompatible changes for the rest of the OD work though, so I can release and push 4.0.0 to Central today and we can avoid depending on a snapshot version for release. :smile:

io7m:
Erk, I've just realized that this is missing one important commit! Watch this space...

simplified-links-json/src/main/java/org.nypl.simplified.links.json/LinkSerialization.kt
```diff
@@ -17,7 +17,7 @@ object LinkSerialization {
    */
 
   fun serializeLink(link: Link): ObjectNode {
-    val node = this.mapper.createObjectNode()
+    val node = mapper.createObjectNode()
```
twaddington: Why'd you remove `this` here? Not complaining just curious :)

simplified-main/src/main/res/values/stringsBoot.xml
```diff
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 
 <resources>
+  <string name="bootingAudioBookExtensions">Starting Audio Book extensions…</string>
```
twaddington: Instead of having the rich text ellipsis hard coded it's usually preferable to use the unicode entity.

```
<string name="bootingAudioBookExtensions">Starting Audio Book extension&#8230;</string>
```

Less fragile and it makes it easier to do things like grep for that specific entity.

simplified-links-json/src/main/java/org.nypl.simplified.links.json/LinkSerialization.kt
io7m: Oof. Not intentional. Suspect that may've been an IDE issue.

simplified-main/src/main/res/values/stringsBoot.xml
io7m: This one was an IDE suggestion - the original was three periods "...". I agree it's better to use the unicode entity and let the XML parser expand it.

twaddington: Sounds good. Should be an easy search and replace when you're working on the translation ticket.

simplified-books-controller/src/main/java/org/nypl/simplified/books/controller/BookBorrowWithDefaultAcquisitionTask.kt
```diff
@@ -31,31 +36,60 @@ class BookBorrowWithDefaultAcquisitionTask(
...
   private val borrowStrings =
     this.services.requireService(BookBorrowStringResourcesType::class.java)
+  private val bookRegistry =
+    this.services.requireService(BookRegistryType::class.java)
+
+  /**
+   * The initial book value.
+   */
+
+  private val bookInitial: Book =
```
twaddington: You can drop the explicit type here since it can be inferred directly from the assignment.

simplified-books-database-api/src/main/java/org/nypl/simplified/books/book_database/api/BookFormats.kt
```diff
@@ -142,6 +153,21 @@ object BookFormats {
...
+    abstract fun supportedContentTypes(): Set<MIMEType>
+
+    /**
+     * @return `true` if the format handle supports content of the given content type
+     */
+
+    fun supports(
+      contentType: MIMEType
+    ): Boolean {
+      for (supportedType in this.supportedContentTypes()) {
```
twaddington: I think you could simplify this and make it more idiomatic with `any`:

```kotlin
fun supports(contentType: MIMEType): Boolean {
    return this.supportedContentTypes().any { supportedType ->
        contentType.fullType == supportedType.fullType
    }
}
```

simplified-viewer-audiobook/src/main/java/org/nypl/simplified/viewer/audiobook/AudioBookManifests.kt
```diff
@@ -0,0 +1,156 @@
...
+    onManifestFulfillmentEvent: (ManifestFulfillmentEvent) -> Unit
+  ): PlayerResult<ManifestFulfilled, ManifestFulfillmentErrorType> {
+    this.logger.debug("downloadManifest")
+
+    val strategies =
+      ManifestFulfillmentStrategies.findStrategy(ManifestFulfillmentBasicType::class.java)
+        ?: throw UnsupportedOperationException()
+
+    val fulfillCredentials =
+      if (credentials != null) {
```
twaddington:
```kotlin
val fulfillCredentials = credentials?.let {
    ManifestFulfillmentBasicCredentials(
        userName = it.barcode().value(),
        password = it.pin().value()
    )
}
```

simplified-viewer-audiobook/src/main/java/org/nypl/simplified/viewer/audiobook/AudioBookPlayerActivity.kt
```diff
@@ -395,6 +401,38 @@ class AudioBookPlayerActivity : AppCompatActivity(),
...
+    services: ServiceDirectoryType,
+    extensions: List<PlayerExtensionType>
+  ) {
+    val feedbooksConfigService =
+      services.optionalService(AudioBookFeedbooksServiceType::class.java)
+
+    if (feedbooksConfigService != null) {
+      this.log.debug("feedbooks configuration service is available; configuring extension")
+      val extension =
+        extensions.filter { ex -> ex is FeedbooksPlayerExtension }
```
twaddington: Pretty sure you can simplify this using the built-in [filterIsInstance](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/filter-is-instance.html) method.

```kotlin
val extension = extensions
    .filterIsInstance<FeedbooksPlayerExtension>()
    .firstOrNull()
```

Actually, you may be able to simplify it even further like this:

```kotlin
val extension = extensions.firstOrNull { it is FeedbooksPlayerExtension }
```

io7m: So it wasn't just my imagination. I was looking for the method that had that signature and just couldn't see it... :roll_eyes:

PR #693, opened 2020-03-20
SIMPLY-2627: Remove global 'Accept-Encoding: identity' header.

twaddington:
**What's this do?**
This change allows the default HttpURLConnection to accept gzip
encoded responses.

This will help us reduce bandwidth used and increase the speed of larger requests.

- https://github.com/NYPL-Simplified/library_registry/pull/163
- https://jira.nypl.org/browse/SIMPLY-2627

Make sure books can still be downloaded.

None

Yes, me.

### DIFF
--- a/simplified-downloader-core/src/main/java/org/nypl/simplified/downloader/core/DownloaderHTTP.java
+++ b/simplified-downloader-core/src/main/java/org/nypl/simplified/downloader/core/DownloaderHTTP.java
@@ -223,7 +223,7 @@ public final class DownloaderHTTP implements DownloaderType
             this.log.debug("download cancelled");
             this.listener.onDownloadCancelled(this);
           } else {
-            if (this.total != expected) {
+            if (this.total < expected) {
               this.log.error(
                 "received {} bytes but expected {}",
                 Long.valueOf(this.total),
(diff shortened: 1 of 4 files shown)


io7m:
If the books can still be downloaded, this all looks fine to me. I think we should just go with this and put out any fires that might start afterwards. If the test suite completes, then we're probably fine!

simplified-downloader-core/src/main/java/org/nypl/simplified/downloader/core/DownloaderHTTP.java
```diff
@@ -223,7 +223,7 @@ private static String getContentType(
             this.log.debug("download cancelled");
             this.listener.onDownloadCancelled(this);
           } else {
-            if (this.total != expected) {
+            if (this.total < expected) {
```
io7m: I prefer the outright paranoia of the original check, but I don't care much either way given the moribund nature of this bit of code. :wink:

twaddington: Yeah, this just prevents the code from throwing an error if the content length is zero, which is what should happen if we receive a gzip encoded response.

twaddington:
Yep! Tests pass and I tried downloading a few books and everything seems good.

PR #697, opened 2020-03-30
SIMPLY-2648: Show the problem report message when a book borrow fails.

twaddington:
**What's this do?**
This change shows a detailed "You have reached your loan limit" message on the book detail screen instead of the generic "operation failed" message.

Additionally, this patch also updates the error page layout to fix an issue that sometimes made contents unreadable.

https://jira.nypl.org/browse/SIMPLY-2648

1. Settings -> App version -> Debug settings -> Show non-production libraries.
2. Add a "New York Public Library - QA Server" account.
3. Sign-in (ask Tristan for credentials, if needed).
4. Attempt to check out any book.
5. Ensure the correct error message is displayed.
6. Tap "details" and ensure the screen works as expected.

n/a

n/a

Yes, Tristan did.

**Screenshots**
- [Updated error message](https://user-images.githubusercontent.com/107117/77945277-3ee29c00-7275-11ea-8de6-5f2cee858ae9.png)
- [Error details - Before](https://user-images.githubusercontent.com/107117/77945283-41dd8c80-7275-11ea-8fe9-25239dec9d9a.png)
- [Error details - After](https://user-images.githubusercontent.com/107117/77945282-4144f600-7275-11ea-8677-5010b10e43ac.png)

### DIFF
--- a/simplified-ui-errorpage/src/main/java/org/nypl/simplified/ui/errorpage/ErrorPageFragment.kt
+++ b/simplified-ui-errorpage/src/main/java/org/nypl/simplified/ui/errorpage/ErrorPageFragment.kt
@@ -1,11 +1,14 @@
 package org.nypl.simplified.ui.errorpage
 
+import android.graphics.Typeface
 import android.os.Bundle
+import android.text.Spannable
+import android.text.SpannableStringBuilder
+import android.text.style.StyleSpan
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
 import android.widget.Button
-import android.widget.TableLayout
 import android.widget.TextView
 import androidx.fragment.app.Fragment
 import androidx.recyclerview.widget.LinearLayoutManager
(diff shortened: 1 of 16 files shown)


io7m:
I really like the new visuals. I'm just reading through the patch now.

One minor concern: The reason we went back to that generic error message is that I'm not completely certain that the message coming from the books controller is always going to be a sensible string. Particularly in the case of Adobe DRM errors, we sometimes get nonsense.

io7m:
Cancel that, I see you're only taking the string from the problem report or using the generic string if the problem report isn't available. That'll work without issue.

simplified-ui-errorpage/src/main/java/org/nypl/simplified/ui/errorpage/ErrorPageFragment.kt
```diff
@@ -95,25 +92,24 @@ class ErrorPageFragment : Fragment() {
...
-          tableRow.findViewById<TextView>(R.id.errorAttributeKey)
-        val tableCell1 =
-          tableRow.findViewById<TextView>(R.id.errorAttributeValue)
-
-        tableCell0.text = key
-        tableCell1.text = value
-        this.errorAttributesTable.addView(tableRow)
+      this.errorDetails.text = ""
+
+      val ssb = SpannableStringBuilder()
```
io7m: Hadn't heard of SpannableStringBuilder before. :+1:

PR #699, opened 2020-04-01
Removed simplified-mime from README.

twaddington:


io7m:
:+1: 

For reference, the table can be generated with:

```
$ java src/misc/ReadMe.java
```

Assuming that `java` is from a JDK 13+ install.

twaddington:
@io7m neat!

PR #700, opened 2020-04-01
SIMPLY-2627: Load testing libraries only when required.

twaddington:
**What's this do?**
This change updates the NYPL account provider registry to only load
testing libraries when necessary. This should speed up the first app
launch.

https://jira.nypl.org/browse/SIMPLY-2627

1. Ensure the account list is populated normally.
2. Navigate to "Settings -> Version -> Build (Tap to enable debug settings)"
3. Toggle "Show non-production libraries" to true.
4. Ensure the account list shows non-production accounts as expected.

This pull-request includes tests.

n/a

n/a

Yep, I tested this.

### DIFF
--- a/simplified-accounts-registry-api/src/main/java/org/nypl/simplified/accounts/registry/api/AccountProviderRegistryType.kt
+++ b/simplified-accounts-registry-api/src/main/java/org/nypl/simplified/accounts/registry/api/AccountProviderRegistryType.kt
@@ -42,9 +42,12 @@ interface AccountProviderRegistryType {
 
   /**
    * Refresh the available account providers from all sources.
+   *
+   * @param includeTestingLibraries A hint for providers indicating whether
+   * testing libraries should be loaded. May be ignored by some providers.
    */
 
-  fun refresh()
+  fun refresh(includeTestingLibraries: Boolean)
 
   /**
    * Return an immutable read-only of the account provider descriptions.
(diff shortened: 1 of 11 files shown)


simplified-accounts-registry-api/src/main/java/org/nypl/simplified/accounts/registry/api/AccountProviderRegistryType.kt
```diff
@@ -42,9 +42,12 @@ interface AccountProviderRegistryType {
 
   /**
    * Refresh the available account providers from all sources.
+   *
+   * @param includeTestingLibraries A hint for providers indicating whether
+   * testing libraries should be loaded. May be ignored by some providers.
    */
 
-  fun refresh()
+  fun refresh(includeTestingLibraries: Boolean)
```
io7m, line 50: This is fine for now, but we'll need a more general mechanism than this shortly to handle things like "libraries near me" requests. Yes, that subject again :smile:

io7m:
I'm fine with this.

simplified-accounts-registry-api/src/main/java/org/nypl/simplified/accounts/registry/api/AccountProviderRegistryType.kt
twaddington: Yeah, I completely agree, it's not super flexible.

PR #701, opened 2020-04-01
Fix tests when run in 'America/Los_Angeles' timezone

twaddington:
Was consistently getting four test failures when run locally on my workstation.


io7m:
:tada: 

This one has been a sporadic failure since 2015! I've never been able to reproduce it.

PR #703, opened 2020-04-03
Overhaul the login dialog

io7m:
**What's this do?**
This adjusts the login dialog so that it is now a full-screen fragment
instead of a Dialog in the Android sense. It hooks up some missing
logic to allow for displaying error details on login failures.

https://jira.nypl.org/browse/SIMPLY-2681

* On an account that requires authentication, when not logged in, try to borrow a book from
a book detail page or from a scrolling feed. Enter some credentials into the login form that you know to be incorrect; observe that an error message is displayed and that an error details button is present and does something.

* On an account that requires authentication, when not logged in, try to borrow a book from
a book detail page or from a scrolling feed. Enter some credentials into the login form that you know to be correct; observe that the dialog goes away when the login process is complete and that the book borrowing starts.

* Observe that the login dialog doesn't look like it spent time in a car crusher.

None.

Not applicable.

@io7m 
Testing: Tested a successful login on an NYPL account
Testing: Tested a failed login on a Hotchkiss account

### DIFF
--- a/simplified-ui-catalog/src/main/java/org/nypl/simplified/ui/catalog/CatalogFragmentLoginDialog.kt
+++ b/simplified-ui-catalog/src/main/java/org/nypl/simplified/ui/catalog/CatalogFragmentLoginDialog.kt
@@ -1,8 +1,6 @@
 package org.nypl.simplified.ui.catalog
 
-import android.content.DialogInterface
 import android.os.Bundle
-import android.view.Gravity
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
@@ -12,7 +10,7 @@ import android.widget.EditText
 import android.widget.ProgressBar
 import android.widget.TextView
 import androidx.annotation.UiThread
-import androidx.fragment.app.DialogFragment
+import androidx.fragment.app.Fragment
 import androidx.lifecycle.ViewModelProviders
 import com.io7m.jfunctional.Some
 import io.reactivex.disposables.Disposable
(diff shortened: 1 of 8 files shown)


simplified-ui-catalog/src/main/java/org/nypl/simplified/ui/catalog/CatalogFragmentLoginDialog.kt
```diff
@@ -328,29 +306,63 @@ class CatalogFragmentLoginDialog : DialogFragment() {
...
+        }
       }
 
       is AccountLoginState.AccountLoggedIn,
-      is AccountLoginState.AccountLoggingOut,
-      is AccountLoginState.AccountLogoutFailed ->
-        this.dismiss()
+      is AccountLoginState.AccountLoggingOut -> {
+        this.findNavigationController().popBackStack()
+        Unit
```
twaddington, line 346: If you remove the return at the top of this function you probably don't need Unit here.

io7m: Unfortunately, that means that the `when` becomes a statement rather than expression and so isn't subjected to exhaustiveness checks. Later on, someone will add a new case to the `AccountLoginState` sealed class, and the compiler won't tell us that this `when` needs to be updated...

twaddington: Wow, I was actually not aware that the behavior of `when` changed when used as an expression. That's an interesting gotcha.

io7m: I think there's a combination of problems here. The JVM doesn't yet have support for sealed classes (that's one of the things currently underway in Project Amber). The Kotlin designers decided that `when` statements need not be exhaustive, presumably for making code convenient to write :-1: 

Unfortunately, it doesn't make sense for a `when` _expression_ not to be exhaustive... In an expression context, every branch of the `when` _needs_ to either yield a value or not terminate. This implies that you need to handle every case. This makes expression `when` behave differently to statement `when`. Modern versions of Java have the same distinction... The old style statement `switch` doesn't need to be exhaustive, while a `switch` expression does.

PR #705, opened 2020-04-07
SIMPLY-2694: Retry 403 Forbidden response with auth, when available.

twaddington:
**What's this do?**
This change fixes a failure with biblioboard where a 403 Forbidden
response is returned instead of a 401 Unauthorized. Instead, of
immediately giving up when a 403 is received, try again with
authentication like we would do with a 401.

Another improvement would be to proactively include the authentication
with the first attempt instead of waiting for a challenge response.

https://jira.nypl.org/browse/SIMPLY-2694

1. Enable non-production libraries.
2. Add an account for Washington State Library.
3. Attempt to download a book and note that it succeeds where it would previously fail.

n/a

n/a

Yes, I did.


io7m:
> Another improvement would be to proactively include the authentication
with the first attempt instead of waiting for a challenge response.

There was a reason we specifically didn't do this, unfortunately it was too many years ago for me to remember...

twaddington:
@io7m I had a feeling there might have been a reason for that so I left it alone for now. Something to consider in a future refactor.

PR #707, opened 2020-04-08
SIMPLY-2312: Single Activity Card Creator

MalcolmMcFly:
**What's this do?**
Product wants SimplyE Android users to be able to create library cards from within the app.
The implementation uses modern Android APIs like ViewModels, ViewBinding, Coroutines, Navigation Components, LiveData, and Retrofit/Moshi for networking. To sum it up simply, the Card Creator simply validates users input along a wizard as 3 endpoints are called to created a new library card. Those endpoints in order are:
1.  [v2/validate/address](https://github.com/NYPL-Simplified/card-creator/wiki/API---V2#post-v2validateaddress)
2. [v2/validate/username](https://github.com/NYPL-Simplified/card-creator/wiki/API---V2#post-v2validateusername)
3. [v2/create_patron](https://github.com/NYPL-Simplified/card-creator/wiki/API---V2#post-v2create_patron)

_The QA endpoints for creating a library card do not work, so all these hit production_

[SIMPLY-2312: Single Activity Card Creator](https://jira.nypl.org/browse/SIMPLY-2312)

First, make sure you have `cardcreator.conf` in the assets directory of `simplified-app-vanilla`.
Testing this can be tricky if you are not in New York. For developers, simply comment the following lines to bypass the New York location check in `LocationFragment.kt`:
```
    // Go to next screen
    binding.nextBtn.setOnClickListener {
      //if (isNewYork) {
        logger.debug("User navigated to the next screen")
        nextAction = LocationFragmentDirections.actionNext()
        navController.navigate(nextAction)
      //} else {
      //  val data = Intent()
      //  activity!!.setResult(Activity.RESULT_CANCELED, data)
      //  activity!!.finish()
      //}
    }
```
For QA who are not in New York, you may need to use a location spoofer like the [Fake GPS](https://play.google.com/store/apps/details?id=com.lexa.fakegps&hl=en_US)

n/a

In progress of updating the README

@MalcolmMcFly


simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/network/CardCreatorService.kt
```diff
@@ -0,0 +1,81 @@
+package org.nypl.simplified.cardcreator.network
+
+import okhttp3.Credentials
+import okhttp3.Interceptor
+import okhttp3.OkHttpClient
+import okhttp3.logging.HttpLoggingInterceptor
+import org.nypl.simplified.cardcreator.models.*
```
io7m, line 7: `*` :-1:

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/viewmodels/AddressViewModel.kt
```diff
@@ -0,0 +1,32 @@
...
+
+  fun validateAddress(address: Address, username: String, password: String) {
+    viewModelScope.launch {
+      try {
+        val cardCreatorService = CardCreatorService(username, password)
+        val response = cardCreatorService.validateAddress(address)
+        validateAddressResponse.postValue(response)
+      } catch (e: Exception) {
+        logger.debug("validateAddress call failed!")
+        e.printStackTrace()
```
io7m, line 27: Better to use:

```
logger.debug("validateAddress call failed: ", e)
```

If the last argument of an `SLF4J` call is an exception, the exception stack trace will be printed (and, importantly, will make it into our log files).

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/ui/LocationFragment.kt
```diff
@@ -0,0 +1,263 @@
...
+          logger.debug( "Region is: ${address.adminArea} ${address.countryCode} ")
+          binding.regionEt.setText("${address.adminArea} ${address.countryCode}", TextView.BufferType.EDITABLE)
+
+          if (address.countryCode == "US" && (address.adminArea == "New York" || address.adminArea == "NY")) {
+            logger.debug("User is in New York")
+            isNewYork = true
+          }
+      } catch (e: Exception) {
+        logger.debug("Error checking to see if user is in New York")
+        e.printStackTrace()
```
io7m, line 237: See the note below about `printStackTrace`

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/ui/LocationFragment.kt
```diff
@@ -0,0 +1,263 @@
...
+        binding.regionEt.setText("${address.adminArea} ${address.countryCode}", TextView.BufferType.EDITABLE)
+
+        if (address.countryCode == "US" && (address.adminArea == "New York" || address.adminArea == "NY")) {
+          logger.debug("User is in New York")
+          isNewYork = true
+        }
+      }
+    } catch (e: Exception) {
+      logger.debug("Error checking to see if user is in New York")
+      e.printStackTrace()
```
io7m, line 165: See notes below about `printStackTrace`

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/ui/LocationFragment.kt
```diff
@@ -0,0 +1,263 @@
...
+        val isNetworkLocationEnabled = locationManager.isProviderEnabled(LocationManager.NETWORK_PROVIDER)
+
+        if (isNetworkLocationEnabled) {
+          locationManager.requestLocationUpdates(LocationManager.NETWORK_PROVIDER, minTimeUpdates, minDistanceUpdates, this)
+          location = locationManager.getLastKnownLocation(LocationManager.NETWORK_PROVIDER)
+        }
+
+      } catch (e: Exception) {
+        logger.debug("Error getting current location")
+        e.printStackTrace()
```
io7m, line 130: See notes below about `printStackTrace`

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/viewmodels/AddressViewModel.kt
MalcolmMcFly: Oh I wasn't aware you could use it like that!

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/ui/LocationFragment.kt
io7m: Is there something better we can do if this fails? The previous card creator had this exact problem with essentially swallowing errors...

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/ui/HomeAddressFragment.kt
```diff
@@ -0,0 +1,175 @@
...
+   */
+  private fun validateForm() {
+    binding.nextBtn.isEnabled = (binding.spState.selectedItem.toString() != getString(R.string.required)
+      && binding.etZip.text.length == addressCharsMin
+      && binding.etStreet1.text.length >= addressCharsMin
+      && binding.etCity.text.length >= addressCharsMin)
+  }
+
+  override fun onNothingSelected(parent: AdapterView<*>?) {
+    TODO("Not yet implemented")
```
io7m, line 168: `TODO` :smile:

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/ui/ConfirmationFragment.kt
```diff
@@ -0,0 +1,175 @@
...
+        }
+        f = File(file.absolutePath + File.separator + "library-card" + ".png")
+        val outStream = FileOutputStream(f)
+        bitmap.compress(Bitmap.CompressFormat.PNG, 10, outStream)
+        outStream.close()
+        addCardToGallery(f)
+        Toast.makeText(activity!!, getString(R.string.card_saved), Toast.LENGTH_SHORT).show()
+      }
+    } catch (e: Exception) {
+      e.printStackTrace()
```
io7m, line 120: See notes below about `printStackTrace`

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/viewmodels/UsernameViewModel.kt
```diff
@@ -0,0 +1,32 @@
...
+
+  fun validateUsername(username: String, authUsername: String, authPassword: String) {
+    viewModelScope.launch {
+      try {
+        val cardCreatorService = CardCreatorService(authUsername, authPassword)
+        val response = cardCreatorService.validateUsername(Username(username))
+        validateUsernameResponse.postValue(response)
+      } catch (e: Exception) {
+        logger.debug("validateUsername call failed!")
+        e.printStackTrace()
```
io7m, line 27: `printStackTrace`

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/viewmodels/PatronViewModel.kt
```diff
@@ -0,0 +1,32 @@
...
+
+  fun createPatron(patron: Patron, username: String, password: String) {
+    viewModelScope.launch {
+      try {
+        val cardCreatorService = CardCreatorService(username, password)
+        val response = cardCreatorService.createPatron(patron)
+        createPatronResponse.postValue(response)
+      } catch (e: Exception) {
+        logger.debug("createPatron call failed!")
+        e.printStackTrace()
```
io7m, line 27: `printStackTrace`

io7m:
I'm mostly fine with this, just some logging and imports that need cleaning up.

I think if we were going to be keeping this code, I'd rather you'd used `jackson-databind` instead of `moshi`; it has the same annotation-based interface, but strong security guarantees with deserialization class whitelisting, and we're already using it throughout the codebase. I'd also rather we did a new `HTTP` implementation wrapping OkHttp and used that everywhere rather than introducing another HTTP library.

Given that we're going to switch to a web-based creator in the not-too-distant future, I'm not bothered about any of the above. Let's keep it as-is for now.

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/ui/LocationFragment.kt
MalcolmMcFly: @io7m We could show a dialog here and direct the user to their device location settings here

MalcolmMcFly:
> I think if we were going to be keeping this code, I'd rather you'd used `jackson-databind` instead of `moshi`; it has the same annotation-based interface, but strong security guarantees with deserialization class whitelisting, and we're already using it throughout the codebase. I'd also rather we did a new `HTTP` implementation wrapping OkHttp and used that everywhere rather than introducing another HTTP library.

I knew you would say this 😂

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/ui/AccountInformationFragment.kt
```diff
@@ -0,0 +1,124 @@
...
+  private val usernameAvailable = "available-username"
+
+  private val viewModel: UsernameViewModel by viewModels()
+
+  override fun onCreateView(
+    inflater: LayoutInflater,
+    container: ViewGroup?,
+    savedInstanceState: Bundle?
+  ): View? {
+    _binding = FragmentAccountInformationBinding.inflate(inflater, container, false)
```
twaddington, line 49: 🎉

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/ui/AccountInformationFragment.kt
```diff
@@ -0,0 +1,124 @@
...
+    return binding.root
+  }
+
+  override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+    super.onViewCreated(view, savedInstanceState)
+
+    navController = Navigation.findNavController(activity!!, R.id.card_creator_nav_host_fragment)
+
+    sharedPreferences = activity!!.getSharedPreferences(Cache.DEFAULT_PREFERENCE_NAME, Context.MODE_PRIVATE)
```
twaddington, line 58: Prefer `requireActivity()` and `requireContext()` to `activity!!`.

twaddington: You can also do something like this to get a locally scoped non-null instance:

```kotlin
val activity = requireActivity()
```

Under the hood it's basically doing:

```kotlin
val activity = this.activity ?: throw IllegalStateException()
```

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/ui/AccountInformationFragment.kt
```diff
@@ -0,0 +1,124 @@
...
+
+    binding.pinEt.addTextChangedListener(object : TextWatcher {
```
twaddington, line 60: Is "Et" short for "EditText"?

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/ui/ConfirmationFragment.kt
```diff
@@ -0,0 +1,175 @@
...
+    card.isDrawingCacheEnabled = true
+    val bitmap = card.drawingCache
+    val f: File
+    try {
+      if (Environment.getExternalStorageState() == Environment.MEDIA_MOUNTED) {
+        val file = File(Environment.getExternalStorageDirectory(), Environment.DIRECTORY_PICTURES)
+        if (!file.exists()) {
+          file.mkdirs()
+        }
+        f = File(file.absolutePath + File.separator + "library-card" + ".png")
```
twaddington, line 112: You can do this instead:

```kotlin
f = File(file.absolutePath, "library-card.png")
```

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/ui/ConfirmationFragment.kt
```diff
@@ -0,0 +1,175 @@
...
+      e.printStackTrace()
+    }
+  }
+
+  /**
+   * Adds library card image to device photo gallery
+   */
+  private fun addCardToGallery(card: File) {
+    val mediaScanIntent = Intent(Intent.ACTION_MEDIA_SCANNER_SCAN_FILE)
+    val contentUri: Uri = Uri.fromFile(card)
```
twaddington, line 129: You should be able to omit the type here.

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/ui/ConfirmationFragment.kt
```diff
@@ -0,0 +1,175 @@
...
+      } else {
+        createDigitalCard()
+      }
+    }
+  }
+
+  /**
+   * Create library card from ImageView
+   */
+  private fun createDigitalCard() {
```
twaddington, line 101: Might be worth doing this work on a background thread using a coroutine.

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/ui/ConfirmationFragment.kt
```diff
@@ -0,0 +1,175 @@
...
+        return
+      }
+    }
+  }
+
+  /**
+   * Returns result to caller with card details
+   */
+  private fun returnResult() {
+    val data = Intent()
```
twaddington, line 165: Would be a little more idiomatic to use "apply" here:

```kotlin
val data = Intent().apply {
  putExtra("type", type)
  // ...
  putExtra("message", message)
}
```

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/ui/LocationFragment.kt
```diff
@@ -0,0 +1,263 @@
...
+   */
+  // TODO: This function is doing to much, break it up into smaller pieces
+  private fun isInNewYorkState(): Boolean {
+    logger.debug("Checking to see if user is in New York")
+    showLoading(true)
+    val locationManager = activity!!.getSystemService(Context.LOCATION_SERVICE) as LocationManager
+    binding.nextBtn.isEnabled = false
+    val maxResults = 1
+    val location = getLocation()
+    val geocoder = Geocoder(activity!!, Locale.getDefault())
```
twaddington, line 147: I've had trouble in the past with the Geocoder not working on some devices. We'll want to keep a close eye on this.

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/ui/LocationFragment.kt
```diff
@@ -0,0 +1,263 @@
...
+      )
+    } else {
+      logger.debug("Location permission granted")
+      try {
+        logger.debug("Getting current location")
+        val locationManager = activity!!.getSystemService(Context.LOCATION_SERVICE) as LocationManager
+        val isNetworkLocationEnabled = locationManager.isProviderEnabled(LocationManager.NETWORK_PROVIDER)
+
+        if (isNetworkLocationEnabled) {
+          locationManager.requestLocationUpdates(LocationManager.NETWORK_PROVIDER, minTimeUpdates, minDistanceUpdates, this)
```
twaddington, line 124: You might consider using the method `requestSingleUpdate` here if all you need is a coarse location fix:

https://developer.android.com/reference/android/location/LocationManager#requestSingleUpdate(android.location.Criteria,%20android.location.LocationListener,%20android.os.Looper)

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/utils/Cache.kt
```diff
@@ -0,0 +1,166 @@
...
+import org.nypl.simplified.cardcreator.models.AccountInformation
+import org.nypl.simplified.cardcreator.models.AddressDetails
+import org.nypl.simplified.cardcreator.models.PersonalInformation
+
+/**
+ * Util class for handling saving/retrieving address across card creator process
+ * This class acts as private key value pair caching system
+ */
+class Cache internal constructor(private val sharedPreferences: SharedPreferences) {
+
```
twaddington, line 14: I'd suggest adding a secondary constructor here that passes in a default SharedPreferences instance:

```kotlin
class Cache internal constructor(
  private val preferences: SharedPreferences
) {
  constructor(context: Context) : this(
    context.getSharedPreferences("CARD_CREATOR", Context.MODE_PRIVATE)
  )
}
```

This should eliminate a bunch of boilerplate code in your fragments.

twaddington:
Looks like you've also got a few failing ktlint checks.

simplified-app-vanilla/src/main/res/xml/network_security_config.xml
```diff
@@ -1,5 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
 <network-security-config>
+  <domain-config cleartextTrafficPermitted="true">
```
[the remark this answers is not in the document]
MalcolmMcFly: Interesting, I'll try this again. I could not hit the patron endpoints without adding this.

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/ui/ConfirmationFragment.kt
MalcolmMcFly: Oh yea! I will add that.

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/ui/LocationFragment.kt
MalcolmMcFly: Yea, pretty much the best we can do without using the [Fused Location Provider API](https://developers.google.com/location-context/fused-location-provider)

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/ui/LocationFragment.kt
MalcolmMcFly: Yes, this is better

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/ui/LocationFragment.kt
twaddington: Is this required? What if they don't want to give us their location? Is there a manual work around?

MalcolmMcFly: We might need to loop @killerpuppytails in on this one. I would suppose we can't let them sign up for the card if we can't without a doubt determine whether or not they are in New York at the time of creating the card. But since they can have a work/school address in New York...maybe they are on vacation and what to sign up for SimplyE....with our current flow, they can't do that.

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/ui/LocationFragment.kt
MalcolmMcFly: [deprecated](https://developer.android.com/reference/android/location/LocationManager#requestSingleUpdate(java.lang.String,%2520android.app.PendingIntent))

simplified-cardcreator/src/main/res/values/strings.xml
```diff
@@ -1,47 +1,143 @@
...
+    <string name="required">Required</string>
+    <string name="optional">Optional</string>
+    <string name="street_1">Street 1</string>
+    <string name="street_2">Street 2</string>
+    <string name="city">City</string>
+    <string name="state">State</string>
+    <string name="zip">Zip</string>
+    <string-array name="states_array">
+        <item>Required</item>
+        <item>Alabama (AL))</item>
```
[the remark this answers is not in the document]
MalcolmMcFly: Wow, good catch

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/ui/ConfirmationFragment.kt
MalcolmMcFly: prettier

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/ui/LocationFragment.kt
killerpuppytails: Yes, their location is required, and they MUST be in the confines of NYS when they apply. It's the only way we can truly confirm they satisfy the requirements of card creation (they must live in, work in, or be visiting NYS) without them coming in to a branch or uploading ID (which we want to avoid).

twaddington: Got it! Thanks y'all.

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/ui/LocationFragment.kt
twaddington: Yeah, it's deprecated in Android R, but we can still use it since no one has that yet. They just want us moving to `getCurrentLocation` eventually.

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/ui/LocationFragment.kt
```diff
@@ -0,0 +1,263 @@
...
+        binding.headerStatusDescTv.text = getString(R.string.new_york_success)
+        binding.headerStatusDescTv.setTextColor(ContextCompat.getColor(activity!!, R.color.trans_black))
+      }
+    }
+    logger.debug("Stopping location updates")
+    locationManager.removeUpdates(this)
+  }
+
+  // These are not needed but were invited to the party by Google
+  override fun onStatusChanged(provider: String?, status: Int, extras: Bundle?) { TODO("Not yet implemented") }
```
twaddington, line 259: This will cause the app to crash with a `NotImplementedError` if a location provider's status changes during our request. Unlikely, but since we can't control when the system invokes these callbacks we should remove these calls to `TODO(...)`.

simplified-cardcreator/src/main/res/values/strings.xml
```diff
@@ -1,47 +1,154 @@
...
+    <string name="optional">Optional</string>
+    <string name="street_1">Street 1</string>
+    <string name="street_2">Street 2</string>
+    <string name="city">City</string>
+    <string name="state">State</string>
+    <string name="zip">Zip</string>
+    <string-array name="states_array">
+        <item>Required</item>
+        <item>Alabama (AL)</item>
+        <item>Alaska (AK))</item>
```
twaddington, line 58: Another extraneous closing parenthesis.

simplified-cardcreator/src/main/res/values/strings.xml
```diff
@@ -1,47 +1,154 @@
...
+        <item>Arizona (AZ)</item>
+        <item>Arkansas (AR)</item>
+        <item>California (CA)</item>
+        <item>Colorado (CO)</item>
+        <item>Connecticut (CT)</item>
+        <item>Delaware (DE)</item>
+        <item>Florida (FL)</item>
+        <item>Georgia (GA)</item>
+        <item>Hawaii (HI)</item>
+        <item>Idaho	(ID)</item>
```
twaddington, line 68: Tab

simplified-cardcreator/src/main/res/values/strings.xml
```diff
@@ -1,47 +1,154 @@
...
+        <item>South Carolina (SC)</item>
+        <item>South Dakota (SD)</item>
+        <item>Tennessee	(TN)</item>
+        <item>Texas	(TX)</item>
+        <item>Utah (UT)</item>
+        <item>Vermont (VT)</item>
+        <item>Virginia (VA)</item>
+        <item>Washington (WA)</item>
+        <item>West Virginia	(WV)</item>
+        <item>Wisconsin	(WI)</item>
```
twaddington, line 105: Tab

simplified-cardcreator/src/main/res/values/strings.xml
```diff
@@ -1,47 +1,154 @@
...
+        <item>Rhode Island (RI)</item>
+        <item>South Carolina (SC)</item>
+        <item>South Dakota (SD)</item>
+        <item>Tennessee	(TN)</item>
+        <item>Texas	(TX)</item>
+        <item>Utah (UT)</item>
+        <item>Vermont (VT)</item>
+        <item>Virginia (VA)</item>
+        <item>Washington (WA)</item>
+        <item>West Virginia	(WV)</item>
```
twaddington, line 104: Tab

simplified-cardcreator/src/main/res/values/strings.xml
```diff
@@ -1,47 +1,154 @@
...
+        <item>North Dakota (ND)</item>
+        <item>Ohio (OH)</item>
+        <item>Oklahoma (OK)</item>
+        <item>Oregon (OR)</item>
+        <item>Pennsylvania (PA)</item>
+        <item>Rhode Island (RI)</item>
+        <item>South Carolina (SC)</item>
+        <item>South Dakota (SD)</item>
+        <item>Tennessee	(TN)</item>
+        <item>Texas	(TX)</item>
```
twaddington, line 99: Tab

simplified-cardcreator/src/main/res/values/strings.xml
```diff
@@ -1,47 +1,154 @@
...
+        <item>North Carolina (NC)</item>
+        <item>North Dakota (ND)</item>
+        <item>Ohio (OH)</item>
+        <item>Oklahoma (OK)</item>
+        <item>Oregon (OR)</item>
+        <item>Pennsylvania (PA)</item>
+        <item>Rhode Island (RI)</item>
+        <item>South Carolina (SC)</item>
+        <item>South Dakota (SD)</item>
+        <item>Tennessee	(TN)</item>
```
twaddington, line 98: Tab

simplified-cardcreator/src/main/res/values/strings.xml
```diff
@@ -1,47 +1,154 @@
...
+        <item>Georgia (GA)</item>
+        <item>Hawaii (HI)</item>
+        <item>Idaho	(ID)</item>
+        <item>Illinois (IL))</item>
+        <item>Indiana (IN))</item>
+        <item>Iowa (IA)</item>
+        <item>Kansas (KS)</item>
+        <item>Kentucky (KY)</item>
+        <item>Louisiana (LA)</item>
+        <item>Maine	(ME)</item>
```
twaddington, line 75: Tab

simplified-cardcreator/src/main/res/values/strings.xml
```diff
@@ -1,47 +1,154 @@
...
+        <item>Arkansas (AR)</item>
+        <item>California (CA)</item>
+        <item>Colorado (CO)</item>
+        <item>Connecticut (CT)</item>
+        <item>Delaware (DE)</item>
+        <item>Florida (FL)</item>
+        <item>Georgia (GA)</item>
+        <item>Hawaii (HI)</item>
+        <item>Idaho	(ID)</item>
+        <item>Illinois (IL))</item>
```
twaddington, line 69: Extra paren

simplified-cardcreator/src/main/res/values/strings.xml
```diff
@@ -1,47 +1,154 @@
...
+        <item>Indiana (IN))</item>
```
twaddington, line 70: Extra paren

MalcolmMcFly:
@twaddington  So the tab bugs are weird, they don't show in my IDE

twaddington:
Looks good! Just need to fix some ktlint failures and a couple issues with the string resources.

twaddington:
> So the tab bugs are weird, they don't show in my IDE &mdash; @MalcolmMcFly 

Yeah, I assume they snuck in because this array was copied from somewhere. You might consider turning on "Show whitespace" in your editor, at least temporarily, to see if that reveals them.

Screen Shot 2020-04-10 at 9 08 32 AM

PR #712, opened 2020-04-14
Add a "sync now" analytics event

io7m:
**What's this do?**
This adds an analytics event that implementations can respond to by
sending any buffered data to the server. This isn't used by Circulation
analytics (which doesn't buffer anything), but is used by LFA analytics
for debugging. This also adds a button to the developer settings menu
that manually publishes the event.

LFA have an analytics implementation that heavily buffers data (due to having to work in extremely poor network conditions). For ease of debugging, we need to be able to tell the analytics system "Send everything you have right now!".

No testing required. The sync event is never published by the application outside of the debug menu.

None.

Not applicable.

@io7m

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -11,7 +11,7 @@ POM_SCM_CONNECTION=scm:git:git://github.com/NYPL-Simplified/Simplified-Android-C
 POM_SCM_DEV_CONNECTION=scm:git:ssh://git@github.com/NYPL-Simplified/Simplified-Android-Core
 POM_SCM_URL=http://github.com/NYPL-Simplified/Simplified-Android-Core
 POM_URL=http://github.com/NYPL-Simplified/Simplified-Android-Core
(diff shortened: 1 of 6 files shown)


gradle.properties
```diff
@@ -11,7 +11,7 @@ POM_SCM_CONNECTION=scm:git:git://github.com/NYPL-Simplified/Simplified-Android-C
 POM_SCM_DEV_CONNECTION=scm:git:ssh://git@github.com/NYPL-Simplified/Simplified-Android-Core
 POM_SCM_URL=http://github.com/NYPL-Simplified/Simplified-Android-Core
 POM_URL=http://github.com/NYPL-Simplified/Simplified-Android-Core
-VERSION_NAME=5.0.2
```
twaddington, line 14: 👍

io7m: Does this need a merge back from `develop`? Unsure what happened there.

PR #714, opened 2020-04-21
Use the new AudioBook API

io7m:
**What's this do?**
This incorporates all the upgrades needed to use the new AudioBooks API.

Briefly:

- We now use the AudioBooks API to download manifests during the borrowing task
- We now use the same code to download fresh manifests when the player starts up
- We respond to expired links in manifests by trying to fetch a new manifest

A wide range of Jira tasks involving new audio book providers.

At a minimum, borrowing of audio books should be tested to ensure that it's still possible to borrow them. Other than this, testing simply involves seeing if audio books from the new providers will download and/or play correctly.

None.

Not applicable, internal changes only.

@io7m painfully and at length.


io7m:
Going to hold off merging this until we get the build out with Crashlytics and friends.

io7m:
I'll sort that conflict out after I push `5.1.0` to Central.

PR #715, opened 2020-04-21
SIMPLY-2708: First name with 3 letters doesn't allow new signup to continue

MalcolmMcFly:
**What's this do?**
This removes the 4 character minimum from first and last name during the card creation process.

[SIMPLY-2708: First name with 3 letters doesn't allow new signup to continue](https://jira.nypl.org/browse/SIMPLY-2708)  
During QA, Joe found that he cannot create a card using the name `Bob`. @killerpuppytails has verified with Antonio that there is no character minimum on patron names.

Attempt to create a card with just 1 character for the first and last name fields in the card creator.

n/a

n/a

@MalcolmMcFly


MalcolmMcFly:
> Is this text input protected against whitespace? If not, might be worth using `isNotBlank`. Imagine I type a single space character in there...

I think `trim()` will remove any white space.

PR #717, opened 2020-04-22
Log profile selections

io7m:
**What's this do?**
This adds a missing analytics event: Profile selections were not
being published.

LFA need this analytics event.

No testing required. This has no effect on non-LFA builds.

None.

Not applicable.

@io7m


io7m:
I don't think this one is worth reviewing. It doesn't change behaviour on non-LFA builds, and I've tested it there.

PR #718, opened 2020-04-23
Add an Instabug service

io7m:
**What's this do?**
This provides an Instabug SPI and calls it on boot to allow external
builds to optionally use Instabug.

To integrate Instabug into new SimplyE builds.

No testing required here. If an Instabug service isn't present, nothing happens.

None.

Not applicable (we don't document these services, maybe we should!)

@io7m @MalcolmMcFly


io7m:
@MalcolmMcFly reviewing is just a formality, as you've already seen and implemented this service. :laughing:

PR #720, opened 2020-04-26
Perform some build.gradle cleanup

twaddington:


build.gradle
```diff
@@ -192,17 +189,6 @@ subprojects { project ->
       apply plugin: "java-library"
       apply plugin: "kotlin"
 
-      /*
-       * Apply the Animal Sniffer plugin to check that code is Android compatible.
-       */
-
-      apply plugin: "ru.vyarus.animalsniffer"
```
twaddington, line 199: @io7m are you actively using this? It seemed out of date and unused so I opted to remove it. However, I'm happy to revert this if it's something you're actively using.

simplified-opds-core/build.gradle
```diff
@@ -3,8 +3,6 @@ dependencies {
 
   implementation project(":simplified-json-core")
 
-  implementation libraries.commonsLang
-  implementation libraries.googleGSON
```
twaddington, line 7: These deps seemed unused. I ran `assemble` and everything worked okay.

build.gradle
```diff
@@ -354,6 +340,11 @@ Automatic-Module-Name: ${POM_AUTOMATIC_MODULE_NAME}
               name = "Malcolm Woods"
               url = "https://github.com/MalcolmMcFly"
             }
+            developer {
```
io7m, line 343: This is overdue. :smile:

build.gradle
io7m: Technically we're still using it for reasons of paranoia. Some of the modules are jar-typed rather than aar-typed, and this just adds an extra layer of safety to ensure we aren't calling APIs that aren't supported by Android.

It's not essential.

simplified-opds-core/build.gradle
io7m: Glad to see them gone.

build.gradle
twaddington: No problem, I reverted this! Just wanted to see if there was any cruft I could clear out.

PR #721, opened 2020-04-26
Removed simplified-app-vanilla-with-profiles and added new flavor to simplified-app-vanilla.

twaddington:
**What's this do?**
This change removes the `simplified-app-vanilla-with-profiles` module and adds a new build flavor to `simplified-app-vanilla` named `vanillaWithProfiles` .

The "with-profiles" app was largely duplicative of the base "vanilla" sample app. Merging the two together eliminated some redundant code and will make them easier to maintain in the future. For example, it will no longer be necessary to update both projects when a new activity is added (e.g. card-creator).

The two flavors should be built and run to verify they work as before.

```
./gradlew clean
./gradlew :simplified-app-vanilla:assemble
./gradlew :simplified-app-vanilla:installVanillaDebug
./gradlew :simplified-app-vanilla:installVanillaWithProfilesDebug
```

n/a

Yes.

@twaddington


io7m:
This seems OK. My only issue with this is that the output APK files have odd names:

```
./simplified-app-vanilla/build/outputs/apk/vanillaWithProfiles/debug/5.0.0-7007-vanillaWithProfiles-debug.apk
./simplified-app-vanilla/build/outputs/apk/vanilla/debug/5.0.0-7007-vanilla-debug.apk
```

Doesn't really fit with the names as they've appeared in the `android-binaries` repos to date.

twaddington:
> Doesn't really fit with the names as they've appeared in the `android-binaries` repos to date. &mdash; @io7m 

Yeah, I noticed that 😕I can fix the archive names but it requires a little bit of extra code. Are we relying on those names to be consistent or is it more of an aesthetic concern?

io7m:
It might upset pushing to Central. I'm not sure yet... I'm hoping the publication plugin(s) do the necessary mangling to get artifact names into a sensible format.

We don't depend on pushing those APKs to Central in any way, but the current publication process seems to be all-or-nothing, so it does need to keep working.

twaddington:
> We don't depend on pushing those APKs to Central in any way, but the current publication process seems to be all-or-nothing, so it does need to keep working. &mdash; @io7m 

Got it! That makes sense. Would it make sense to skip publishing APKs and just do Jars and AARs? Or do we want those APKs archived?

io7m:
> Got it! That makes sense. Would it make sense to skip publishing APKs and just do Jars and AARs? Or do we want those APKs archived in central?

I'm not sure. It's nice to have, but we don't depend on it.

twaddington:
> It's nice to have, but we don't depend on it. &mdash; @io7m 

Hrm, well I added a commit that disables publishing for `android-application` projects. It's easy to revert if we decide we'd rather keep those archives.

I also extracted the publishing configuration block from `build.gradle` and moved it to `publishing.gradle` to clean that file up a bit.

io7m:
I think we should go for it. :tada:

PR #723, opened 2020-04-27
Refresh account providers frequently

io7m:
**What's this do?**
This updates the book account sync task to re-resolve account providers
before syncing loans. This ensures that we're always using up-to-date
information from the authentication document.

Additionally, a button has been added in the developer settings menu
to manually trigger a sync of all accounts.

Fix: https://jira.nypl.org/browse/SIMPLY-2629

I'm not completely sure how this should be tested as I'm not sure what the problem was that prompted this ticket. @leonardr Do you have any ideas?

None.

Documentation on the accounts system may be incoming.

@io7m

### DIFF
--- a/simplified-ui-settings/src/main/java/org/nypl/simplified/ui/settings/SettingsFragmentVersion.kt
+++ b/simplified-ui-settings/src/main/java/org/nypl/simplified/ui/settings/SettingsFragmentVersion.kt
@@ -24,6 +24,7 @@ import org.nypl.simplified.accounts.registry.api.AccountProviderRegistryType
 import org.nypl.simplified.adobe.extensions.AdobeDRMExtensions
 import org.nypl.simplified.analytics.api.AnalyticsEvent
 import org.nypl.simplified.analytics.api.AnalyticsType
+import org.nypl.simplified.books.controller.api.BooksControllerType
 import org.nypl.simplified.boot.api.BootFailureTesting
 import org.nypl.simplified.buildconfig.api.BuildConfigurationServiceType
 import org.nypl.simplified.navigation.api.NavigationControllers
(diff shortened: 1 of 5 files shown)


simplified-ui-settings/src/main/java/org/nypl/simplified/ui/settings/SettingsFragmentVersion.kt
```diff
@@ -214,6 +221,25 @@ class SettingsFragmentVersion : Fragment() {
...
+
+      try {
+        this.profilesController.profileCurrent()
+          .accounts()
+          .values
+          .forEach { account ->
+            this.booksController.booksSync(account)
+          }
+      } catch (e: Exception) {
+        this.logger.error("ouch: ", e)
```
twaddington, line 239: 😂

leonardr:
It's tough to test this without access to a web proxy on your phone, but if you did have that capability, here's how to test it:

* Set your phone's location to New York City and open SimplyE as a new user. When asked to select a library, NYPL and Brooklyn Public Library should be at the top of the list.
* Close SimplyE.
* Set your phone's location to somewhere far away from New York City.
* Wait an hour.
* Open SimplyE and try to add a library. The library list should be fetched fresh and you should see an alphabetical list of libraries _without_ NYPL or Brooklyn at the top.

I have questions about the implementation which I'll put in a separate comment.

leonardr:
I deleted an earlier comment because I confused the feed of libraries from the registry and the authentication document for a specific library.

We can test this on NYPL QA by changing the display name of the library. If you choose NYPL QA as your library, and then I go into the admin interface and change the display name of the library, I would expect that change to be reflected in SimplyE an hour later. Does that work?

io7m:
I think that should work. It _may_ be instantaneous if you click the "sync" button in the hidden developer settings, but I'm not sure.

PR #724, opened 2020-04-28
Removed the multidex support library.

twaddington:
Multidex is on by default in API 21 and above.

> If your `minSdkVersion` is set to 21 or higher, multidex is enabled by
default and you do not need the multidex support library.

https://developer.android.com/studio/build/multidex#mdex-gradle


io7m:
One down.

PR #725, opened 2020-04-29
Switch to a model where the registry resolves accounts

io7m:
**What's this do?**
Previously, we had a model where an `AccountProviderDescription` had
a `resolve()` method, and this `resolve()` method was capable of performing
an effectful computation that would produce a full `AccountProvider`. The
`AccountProvider` also had a `toDescription()` method that could turn it
back into an equivalent `AccountProviderDescription` value. In the absence
of I/O and/or network errors, `provider.toDescription().resolve() == provider`.

This was fine, except that we also had a convenient immutable
implementation of the `AccountProvider` class that preserved the
`provider.toDescription().resolve() == provider` property, but where
`resolve()` didn't perform any side effects (such as consulting the
Library Registry for new information). These description/provider
values could sometimes make their way into the app's registry, and
would then make it impossible to update the provider later.

This commit changes the model: We have one `AccountProvider` and one
`AccountProviderDescription` type everywhere, and those types are merely
dumb carriers of data and don't reference any kind of resolution logic.
The accounts source SPI and account registry API have been updated with
a new `resolve()` method that does the work that the previous account
provider types did. Effectively, we move to the familiar model where
an API (the registry) asks each registered source if it can resolve
a given `AccountProviderDescription`, and it delegates the job to the
first source that responds in the affirmative.

This commit touches many files, but the changes are minor (some
type renaming). This now appears to result in the correct behaviour
with regards to updating account provider descriptions from the
authentication document; we _know_ for certain that everywhere there's
a call to `resolve()`, new information will be received and saved.

Fix: https://jira.nypl.org/browse/SIMPLY-2629

The same way `SIMPLY-2629` was originally tested.

None.

The `README.md` has been regenerated.

@io7m


io7m:
Merge conflict resolved.

io7m:
I did test this earlier against the QA server and that's why the QA server now has that amusing title. It doesn't work as quickly as I'd like - presumably due to some cache effects that I don't fully understand - but it does work. As @leonardr put it, updating eventually is better than never updating.

MalcolmMcFly:
@io7m can you share the link for the testing for SIMPLY-2629?

io7m:
> @io7m can you share the link for the testing for SIMPLY-2629?

That was in the previous PR: https://github.com/NYPL-Simplified/Simplified-Android-Core/pull/723

@leonardr changed the title on the NYPL QA server (in the authentication document), and I checked that the app displayed the changes. It wasn't immediate, but it eventually did.

PR #726, opened 2020-04-29
Ensure that loans are synced

io7m:
**What's this do?**
This makes the necessary change to run the BooksSync task each time a
profile is selected. In LFA builds, this is a no-op. In NYPL builds,
this happens once on program startup.

This appears to be the root cause of the stale loans issue that has been seen recently.

Check that loans are up-to-date each time the application starts.

None.

Not applicable. Re-adding behaviour that was accidentally removed.

@io7m @MalcolmMcFly


io7m:
Yeah, they've been "unstable" forever. To be clear, that just refers to API stability. When we get to API level 28, we can dump them entirely for the standard `CompletableFuture` :tada:

For what it's worth, they don't appear to have updated the API for years. I think they're just avoiding promising not to change it.

PR #731, opened 2020-05-11
SIMPLY-2403: Retrieve annotations links from loans feeds

io7m:
**What's this do?**
This implements the changes required for 2403: When the user syncs
their loans, the feed is inspected for a link with an annotation
service relation. If a link is found, the account provider is updated
with the new annotations service URI.

In the process of implementing this, two nasty bugs were discovered:

  1. In the accounts database, the call to setAccountProvider()
     was updating the description of the account on disk, but was
     not updating the visible account provider on the account
     instance. So, effectively: `a.setAccountProvider(p) -> a.accountProvider != p`
     This was fixed, and many extra assertions were added to the
     code to aggressively ensure that this doesn't happen again.

  2. Authentication document links were being lost on account
     descriptions, meaning that account providers were losing
     authentication descriptions when resolved. This was
     corrected, and some assertions were added to ensure that
     this doesn't happen again.

https://jira.nypl.org/browse/SIMPLY-2403

Open an NYPL account in the settings section. Ensure that the switch to
enable/disable bookmark syncing can be toggled. Open a book and check
to see if your bookmarks are syncing across devices.

None.

Not applicable.

@io7m tried the code on various accounts, and saw the bookmarks were downloaded from the server.

### DIFF
--- a/simplified-accounts-api/src/main/java/org/nypl/simplified/accounts/api/AccountProviderType.kt
+++ b/simplified-accounts-api/src/main/java/org/nypl/simplified/accounts/api/AccountProviderType.kt
@@ -2,6 +2,7 @@ package org.nypl.simplified.accounts.api
 
 import org.joda.time.DateTime
 import org.nypl.simplified.links.Link
+import org.nypl.simplified.opds.core.OPDSFeedConstants.AUTHENTICATION_DOCUMENT_RELATION_URI_TEXT
 import java.net.URI
 import javax.annotation.concurrent.ThreadSafe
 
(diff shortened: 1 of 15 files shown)


simplified-accounts-api/src/main/java/org/nypl/simplified/accounts/api/AccountProviderType.kt
```diff
@@ -231,15 +235,19 @@ interface AccountProviderType : Comparable<AccountProviderType> {
...
+        id = id,
+        title = displayName,
+        updated = updated,
+        links = links.toList(),
+        images = imageLinks.toList(),
+        isAutomatic = addAutomatically,
+        isProduction = isProduction
+      )
+
+    check((this.authenticationDocumentURI != null) == (accountProviderDescription.authenticationDocumentURI != null))
```
twaddington, line 249: Is there a test we could add or modify to catch this in the future?

io7m: There should be! By that I mean that I don't think there is right now...

PR #732, opened 2020-05-12
Feature/card creator bugs

MalcolmMcFly:
**What's this do?**
This adds an ellipses to the location loading test. This also removes the sign up option from the settings screen for non-NYPL users. This also checks the location again when the user returns from the settings screen.

[SIMPLY-2734: In SA Card creator, please animate loading messages](https://jira.nypl.org/browse/SIMPLY-2734)
[SIMPLY-2732: SA Card Creator doesn't error out when an applicant is outside NY](https://jira.nypl.org/browse/SIMPLY-2732)

Access the card creator and proceed to the location check screen to observe the animated ellipses. Access settings from a library other that NYPL and you will find the sign up button is no longer there.

n/a

n/a

@MalcolmMcFly

### DIFF
--- a/simplified-ui-settings/src/main/java/org/nypl/simplified/ui/settings/SettingsFragmentAccount.kt
+++ b/simplified-ui-settings/src/main/java/org/nypl/simplified/ui/settings/SettingsFragmentAccount.kt
@@ -697,7 +697,7 @@ class SettingsFragmentAccount : Fragment() {
    * Hides or show sign up options if is user in accessing the NYPL
    */
   private fun hideCardCreatorForNonNYPL() {
-    if(this.account.provider.cardCreatorURI != null) {
+    if (this.account.provider.cardCreatorURI != null) {
       settingsCardCreator.visibility = View.GONE
     }
   }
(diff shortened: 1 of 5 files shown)


simplified-ui-settings/src/main/java/org/nypl/simplified/ui/settings/SettingsFragmentAccount.kt
```diff
@@ -85,13 +86,15 @@ class SettingsFragmentAccount : Fragment() {
...
   private lateinit var loginProgressText: TextView
+  private lateinit var settingsCardCreator: ConstraintLayout
   private lateinit var parameters: SettingsFragmentAccountParameters
   private lateinit var profilesController: ProfilesControllerType
   private lateinit var uiThread: UIThreadServiceType
   private var accountSubscription: Disposable? = null
   private var profileSubscription: Disposable? = null
   private val cardCreatorResultCode = 101
   private var cardCreatorService: CardCreatorServiceType? = null
+  private var cardCreatorLibrary = "The New York Public Library"
```
io7m, line 97: Should remove this, see below!

simplified-ui-settings/src/main/java/org/nypl/simplified/ui/settings/SettingsFragmentAccount.kt
```diff
@@ -687,6 +694,15 @@ class SettingsFragmentAccount : Fragment() {
     }
   }
 
+  /**
+   * Hides or show sign up options if is user in accessing the NYPL
+   */
+  private fun hideCardCreatorForNonNYPL() {
+    if(this.account.provider.displayName != cardCreatorLibrary) {
```
io7m, line 701: This should be:
```
if (this.account.provider.cardCreatorURI != null) { ...
```
That'll stop the code from breaking if anyone else introduces a card creator in their library.

io7m:
Looks good!

PR #733, opened 2020-05-18
Migrate from external to internal storage

io7m:
**What's this do?**
This registers an unconditional boot hook that moves any and all
application data from external storage to internal storage.

This ensures that we never have any state that outlives an application uninstall.

Affects: https://jira.nypl.org/browse/SIMPLY-2767

If the app starts up and all of your existing accounts are intact, it worked.

None.

Not applicable.

@io7m tested that the accounts still existed after the migration, and also tested that the app didn't explode after it started up with no data to migrate.

### DIFF
--- a/simplified-main/src/main/java/org/nypl/simplified/main/MainServices.kt
+++ b/simplified-main/src/main/java/org/nypl/simplified/main/MainServices.kt
@@ -149,10 +149,8 @@ internal object MainServices {
    */
 
   const val CURRENT_DATA_VERSION = "v4.0"
-  const val SIMPLYE = "org.nypl.simplified.simplye"
 
   private data class Directories(
-    val directoryPrivateBaseVersioned: File,
     val directoryStorageBaseVersioned: File,
     val directoryStorageDownloads: File,
     val directoryStorageDocuments: File,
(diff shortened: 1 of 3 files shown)


simplified-main/src/main/java/org/nypl/simplified/main/MainServices.kt
```diff
@@ -162,24 +160,15 @@ internal object MainServices {
   private fun initializeDirectories(context: Context): Directories {
     this.logger.debug("initializing directories")
 
-    val directoryPrivateBase =
-      context.filesDir
-    val directoryPrivateBaseVersioned =
-      File(directoryPrivateBase, this.CURRENT_DATA_VERSION)
-    val directoryStorageBase =
-      this.determineDiskDataDirectory(context)
```
twaddington, line 170: Is the method `determineDiskDataDirectory ` still used?

twaddington:
@io7m the documentation for `getExternalFilesDir` states that:

> This is like `getFilesDir()` in that these files will be deleted when the application is uninstalled

https://developer.android.com/reference/kotlin/android/content/Context#getexternalfilesdir

If that is indeed the case, I'm unsure that this patch would fix what it's intended to.

simplified-main/src/main/java/org/nypl/simplified/main/MainServices.kt
twaddington: If it is still used can we at least remove the block referencing external storage?

io7m: My mistake, I thought I deleted it. Pushed a commit that removes the unused method.

io7m:
I believed that external storage was deleted when the app was uninstalled too, but it seems like at least on Risa's device, that evidently doesn't always happen.

In any case, it means that all of our state is in one predictable directory instead of being spread across (potentially) two.

io7m:
To be clear: I'm not convinced that this patch fixes much beyond centralizing state. However, if we're running into the device bug that it appears we're running into, then this patch will stamp it out.

twaddington:
Looks good!

PR #735, opened 2020-06-11
Feature/open ebooks 2020

io7m:
**What's this do?**
This makes a large pile of updates necessary to support Open eBooks:

* Implement a browser-based OAuth login workflow
* Update the settings screen to support OAuth
* Update the settings screen to support multiple authentication methods
* Update the catalog UI code to use shared logic for triggering logins and book borrowing
* Split the settings fragments relating to accounts into their own module, and use these fragments instead of the duplicative full-screen catalog login fragment (it's finally gone! :tada:)

* https://jira.nypl.org/browse/SIMPLY-2794
* https://jira.nypl.org/browse/SIMPLY-2795
* https://jira.nypl.org/browse/SIMPLY-2797
* https://jira.nypl.org/browse/SIMPLY-1884
* https://jira.nypl.org/browse/SIMPLY-2224

There may be more, but this covers most of the work.

We need to produce an Open eBooks build using the separate repository after this is deployed to Central. We'll have to come up with a QA plan as there's a lot of ground to cover given that noone has updated Open eBooks since 2014...

We don't yet have a way to switch accounts in the UI.

Minimal documentation was added to the new module. The module list is about to be updated.

@io7m Has tried logging in to Open eBooks, and various cases such as cancelling logins, checking that existing login logic around accounts such as NYPL isn't broken, etc.


io7m:
`ktlint` errors in the build, despite me having run `ktlint` before I committed. :weary:

PR #737, opened 2020-06-18
Juvenile card creator

MalcolmMcFly:
**What's this do?**
This adds the ability to create cards for a patron's dependents by adding new ISSO Token API calls and new NYPL Platforms to hit juvenile card creator endpoints. New views have been added to the existing navigation graph and the Accounts screen has been updated to reflect new juvenile card creator options.

[SIMPLY-2312: SA: Card creator](https://jira.nypl.org/browse/SIMPLY-2312)
[SIMPLY-2711: Allow an already-logged-in patron to create new juv-only cards](https://jira.nypl.org/browse/SIMPLY-2734)
[SIMPLY-2725: Restrict patron to 3 child cards](https://jira.nypl.org/browse/SIMPLY-2725)
[SIMPLY-2734: In SA Card creator, please animate loading messages](https://jira.nypl.org/browse/SIMPLY-2734)

To test this functionality, enable the bypass New York for the debugging menu. The simply login to any any adult account and attempt to create a juvenile account. Ensure you have properly include the `cardcreator.conf` file in the assets directory or the card creator functionality will not be enabled.

One thing to watch out for is that when calling the endpoint to check if a patron is allowed to create juvenile cards, the call sometimes fails with a `502` error. This is something we need to discuss with the server team and could potential derail QA.

n/a

@MalcolmMcFly

### DIFF
--- a/simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/CardCreatorActivity.kt
+++ b/simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/CardCreatorActivity.kt
@@ -3,6 +3,7 @@ package org.nypl.simplified.cardcreator
 import android.os.Bundle
 import androidx.appcompat.app.AppCompatActivity
 import org.nypl.simplified.cardcreator.databinding.ActivityCardCreatorBinding
+import org.nypl.simplified.cardcreator.utils.Cache
 
 /**
  * Main view responsible for patrons to create library cards
@@ -16,5 +17,14 @@ class CardCreatorActivity : AppCompatActivity() {
     super.onCreate(savedInstanceState)
     binding = ActivityCardCreatorBinding.inflate(layoutInflater)
     setContentView(binding.root)
+    if (intent.extras.getBoolean("isLoggedIn")) {
+      Cache(this).isJuvenileCard = true
+      binding.toolbarTitleTv.text = getString(R.string.create_child_card)
+    }
+  }
+
+  override fun onDestroy() {
+    super.onDestroy()
+    Cache(this).clear()
   }
 }
--- a/simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/CardCreatorActivity.kt
+++ b/simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/CardCreatorActivity.kt
@@ -19,7 +19,7 @@ class CardCreatorActivity : AppCompatActivity() {
     setContentView(binding.root)
     if (intent.extras.getBoolean("isLoggedIn")) {
       Cache(this).isJuvenileCard = true
-      binding.toolbarTitleTv.text = getString(R.string.create_child_card)
+      binding.toolbarTitleTv.setText(R.string.create_child_card)
     }
   }
 
--- a/simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/model/JuvenilePatron.kt
+++ b/simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/model/JuvenilePatron.kt
@@ -1,0 +1,19 @@
+package org.nypl.simplified.cardcreator.model
+
+import com.squareup.moshi.Json
+
+sealed class JuvenilePatron
+
+data class UsernameParent(
+  @field:Json(name = "name") val name: String,
+  @field:Json(name = "parentUsername") val parentUsername: String,
+  @field:Json(name = "username") val username: String,
+  @field:Json(name = "pin") val pin: String
+) : JuvenilePatron()
+
+data class BarcodeParent(
+  @field:Json(name = "barcode") val barcode: String,
+  @field:Json(name = "name") val name: String,
+  @field:Json(name = "username") val username: String,
+  @field:Json(name = "pin") val pin: String
+) : JuvenilePatron()
--- a/simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/model/JuvenilePatronResponse.kt
+++ b/simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/model/JuvenilePatronResponse.kt
@@ -1,0 +1,6 @@
+package org.nypl.simplified.cardcreator.model
+
+data class JuvenilePatronResponse(
+  val `data`: Data,
+  val status: Int
+)
--- a/simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/network/CardCreatorService.kt
+++ b/simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/network/CardCreatorService.kt
@@ -54,11 +54,9 @@ internal interface CardCreatorService {
 
   companion object {
     operator fun invoke(authUsername: String, authPassword: String): CardCreatorService {
-      val logging = run {
-        val httpLoggingInterceptor = HttpLoggingInterceptor()
-        httpLoggingInterceptor.apply {
-          httpLoggingInterceptor.level = HttpLoggingInterceptor.Level.HEADERS
-        }
+
+      val logging = HttpLoggingInterceptor().apply {
+        level = HttpLoggingInterceptor.Level.HEADERS
       }
 
       val auth = Interceptor {
--- a/simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/network/CardCreatorService.kt
+++ b/simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/network/CardCreatorService.kt
@@ -4,12 +4,12 @@ import okhttp3.Credentials
 import okhttp3.Interceptor
 import okhttp3.OkHttpClient
 import okhttp3.logging.HttpLoggingInterceptor
-import org.nypl.simplified.cardcreator.models.Address
-import org.nypl.simplified.cardcreator.models.ValidateAddressResponse
-import org.nypl.simplified.cardcreator.models.Username
-import org.nypl.simplified.cardcreator.models.ValidateUsernameResponse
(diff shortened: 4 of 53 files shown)


simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/AccountFragment.kt
```diff
@@ -876,11 +884,13 @@ class AccountFragment : Fragment() {
       }
       is AsLogoutButtonEnabled -> {
         this.loginButton.setText(R.string.accountLogout)
+        this.signUpLabel.text = "Want a card for your child?"
```
io7m, line 887: Guessing these were supposed to be the string resources below.

simplified-ui-accounts/src/main/res/values/strings.xml
```diff
@@ -13,6 +13,8 @@
   <string name="accountDelete">Delete</string>
   <string name="accountEULAStatement">I agree to the terms of the End User License Agreement.</string>
   <string name="accountLogin">Log in</string>
+<!--  <string name="accountWantChildCard">Want a card for your child?</string>-->
```
io7m, line 16: See above. :smile:

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/CardCreatorActivity.kt
```diff
@@ -16,5 +17,14 @@ class CardCreatorActivity : AppCompatActivity() {
...
     setContentView(binding.root)
+    if (intent.extras.getBoolean("isLoggedIn")) {
+      Cache(this).isJuvenileCard = true
+      binding.toolbarTitleTv.text = getString(R.string.create_child_card)
+    }
+  }
+
+  override fun onDestroy() {
+    super.onDestroy()
+    Cache(this).clear()
```
io7m, line 28: Does this introduce a lifecycle issue? What happens if the user rotates their device and destroys/recreates the activity? I'm not sure if that's actually a problem for your cache or not.

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/CardCreatorActivity.kt
```diff
@@ -16,5 +17,14 @@ class CardCreatorActivity : AppCompatActivity() {
     super.onCreate(savedInstanceState)
     binding = ActivityCardCreatorBinding.inflate(layoutInflater)
     setContentView(binding.root)
+    if (intent.extras.getBoolean("isLoggedIn")) {
+      Cache(this).isJuvenileCard = true
+      binding.toolbarTitleTv.text = getString(R.string.create_child_card)
```
twaddington, line 22: You should be able to omit the getString call here. The TextView class includes a setText method that accepts a resource id value:

https://developer.android.com/reference/kotlin/android/widget/TextView#settext_3

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/model/JuvenilePatron.kt
```diff
@@ -0,0 +1,19 @@
...
+  @field:Json(name = "parentUsername") val parentUsername: String,
+  @field:Json(name = "username") val username: String,
+  @field:Json(name = "pin") val pin: String
+) : JuvenilePatron()
+
+data class BarcodeParent(
+  @field:Json(name = "barcode") val barcode: String,
+  @field:Json(name = "name") val name: String,
+  @field:Json(name = "username") val username: String,
+  @field:Json(name = "pin") val pin: String
```
twaddington, line 18: I think you can omit these field annotations since the serialized name is the same as the data class member. Would make it more readable.

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/model/JuvenilePatronResponse.kt
```diff
@@ -0,0 +1,6 @@
+package org.nypl.simplified.cardcreator.model
+
+data class JuvenilePatronResponse(
+  val `data`: Data,
```
twaddington, line 4: 😂

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/network/CardCreatorService.kt
```diff
@@ -57,7 +57,7 @@ internal interface CardCreatorService {
       val logging = run {
         val httpLoggingInterceptor = HttpLoggingInterceptor()
         httpLoggingInterceptor.apply {
```
twaddington, line 59: This call to apply appears to be extraneous as currently written. I'd recommend changing to the following:

```
val httpLoggingInterceptor = HttpLoggingInterceptor().apply {
    level = HttpLoggingInterceptor.Level.BASIC
}
```

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/network/NYPLISSOService.kt
```diff
@@ -0,0 +1,55 @@
...
+    @Part client_id: MultipartBody.Part,
+    @Part client_secret: MultipartBody.Part
+  ): ISSOTokenData
+
+  companion object {
+    operator fun invoke(): NYPLISSOService {
+
+      val logging = run {
+        val httpLoggingInterceptor = HttpLoggingInterceptor()
+        httpLoggingInterceptor.apply {
```
twaddington, line 35: See above.

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/network/NYPLPlatformService.kt
```diff
@@ -0,0 +1,84 @@
...
+        val request = it.request().newBuilder()
+          .addHeader("Authorization", "Bearer $token")
+          .addHeader("Content-Type", "application/json")
+          .addHeader("Accept", "application/json").build()
+        it.proceed(request)
+      }
+
+      val logging = run {
+        val httpLoggingInterceptor = HttpLoggingInterceptor()
+        httpLoggingInterceptor.apply {
```
twaddington, line 63: See above.

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/utils/Cache.kt
```diff
@@ -175,4 +177,34 @@ class Cache internal constructor(private val sharedPreferences: SharedPreference
...
+  var isJuvenileCard: Boolean? = null
+    get() {
+      if (field == null) {
+        field = sharedPreferences.getBoolean(KEY_JUVENILE_CARD, false)
+      }
+      return field
+    }
+    set(value) {
+      field = value
+      sharedPreferences.edit { putBoolean(KEY_JUVENILE_CARD, value!!) }
```
twaddington, line 193: Is it possible to get a null value here? Might be safer to do something like

```kotlin
sharedPreferences.edit { putBoolean(KEY_JUVENILE_CARD, value ?: false) }
```

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/utils/FragmentUtils.kt
```diff
@@ -0,0 +1,23 @@
...
+fun Fragment.getCache(): Cache {
+  return Cache(this.requireActivity())
+}
+
+/**
+ * Determines whether or not the user identifier is a barcode or username
+ */
+fun Fragment.isBarcode(identifier: String): Boolean {
+  return try {
+    Double.parseDouble(identifier); true
```
twaddington, line 19: You might be able to use `toDoubleOrNull` here:

```kotlin
return identifier.toDoubleOrNull() != null
```

https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/to-double-or-null.html

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/viewmodel/PlatformViewModel.kt
```diff
@@ -0,0 +1,75 @@
...
+      }
+    }
+  }
+
+  /**
+   * Determines whether or not the user identifier is a barcode or username
+   */
+  fun isBarcode(identifier: String): Boolean {
+    return try {
+      parseDouble(identifier); true
```
twaddington, line 70: See above.

simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/AccountFragment.kt
```diff
@@ -409,13 +410,22 @@ class AccountFragment : Fragment() {
...
+        if (this.account.provider.cardCreatorURI?.scheme.equals(nyplCardCreatorScheme)) {
+          cardCreator.openCardCreatorActivity(
+            this,
+            this.activity,
+            this.cardCreatorResultCode,
+            this.account.loginState is AccountLoggedIn,
+            this.authenticationBasicUser.text.toString().trim()
+          )
+          // Launch external card creator
+        } else {
```
io7m, line 423: The logic for this is a bit wrong. The actual card creator and web calls look correct, but the branches should look more like:

```
// Does the library require the NYPL card creator?
if (cardCreatorURI.scheme.equals(nyplCardCreatorScheme)) {
  // Does this build actually have an NYPL card creator?
  if (cardCreator != null) {
    // Open NYPL card creator
  }
} else {
  // Open web card creator instead
}
```
There needs to be similar logic to check if the button should be enabled/disabled.

PR #742, opened 2020-07-05
[SIMPLY-2865] Fix Toolbar size in host layout.

twaddington:
**What's this do?**
This fixes an issue where the Toolbar was the incorrect size.

https://jira.nypl.org/browse/SIMPLY-2865

Visibly note that the Toolbar is now the correct size on all screens.

N/A

N/A

@twaddington


twaddington:
> So there's a standard attribute for the height, and the value isn't necessarily `64dp`?

It's typically 48dp but varies depending on the class of device you're using.

PR #744, opened 2020-07-07
[SIMPLY-2773] Fixes several issues with the library selection screen and the default account.

twaddington:
**What's this do?**
Fixes several issues with the library selection screen and the default account.

There are still two outstanding issues related to accounts:
1) Because of the way fragments are cached in the backstack you'll sometimes see the wrong catalog until you tap "Catalog" in the bottom navigation. Tapping "refresh" won't fix the issue.
2) There is still no way to switch between multiple active accounts.

https://jira.nypl.org/browse/SIMPLY-2773

1) Install the update.
2) Navigate to the system's app settings and tap "Clear Storage" to ensure no cached data exists for the app.
3) Launch the app and note the library selection screen is presented.
4) From here you'll want to test several things, starting over from Step 2 each time.
  a) Tap "Find Your Library", then navigate "Back", you should still see the library selection screen.
  b) Tap "Find Your Library", then navigate "Home", kill the app by opening the recent apps tray and swiping it away; opening the app again should again present the library selection screen.
  c) Tap "Find Your Library", select any library, you should see the expected catalog for your selection.
  b) Tap "Add a Library Later", you should see the age gate for SimplyE.

n/a

n/a

@twaddington

### DIFF
--- a/simplified-accounts-api/src/main/java/org/nypl/simplified/accounts/api/AccountAuthenticationAdobePreActivationCredentials.kt
+++ b/simplified-accounts-api/src/main/java/org/nypl/simplified/accounts/api/AccountAuthenticationAdobePreActivationCredentials.kt
@@ -36,7 +36,7 @@ data class AccountAuthenticationAdobePreActivationCredentials(
   val deviceManagerURI: URI?,
 
   /**
-   * The post activation credentials that resulted from device activation, if any device activation has been performed
+   * The post activation credentials that resulted from device activation, iff any device activation has been performed
    */
 
   val postActivationCredentials: AccountAuthenticationAdobePostActivationCredentials?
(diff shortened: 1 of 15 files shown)


simplified-accounts-api/src/main/java/org/nypl/simplified/accounts/api/AccountAuthenticationAdobePreActivationCredentials.kt
```diff
@@ -36,7 +36,7 @@ data class AccountAuthenticationAdobePreActivationCredentials(
   val deviceManagerURI: URI?,
 
   /**
-   * The post activation credentials that resulted from device activation, iff any device activation has been performed
```
io7m, line 39: Actually, this is short for "if and only if".

"If" means "If there are credentials → Device activation has been performed." but doesn't necessarily imply "If device activation has been performed → There are credentials".

"Iff" means "If there are credentials → Device activation has been performed" AND "If device activation has been performed → There are credentials".

This isn't the first time this has come up, and it won't be the last. :yum:

twaddington: Ah, I see! I hadn't encountered that shorthand before. Easy enough to revert that change.

PR #745, opened 2020-07-14
Fixed various bugs found in QA

MalcolmMcFly:
**What's this do?**
- Removed "Report issue..." button in layouts and various fragments
- Fixed broken juvenile card creator payload from sealed class. Networking layer was having issues parsing json payload sent in POST requests
- Fixed bug where parent was signed out after creating juvenile cards. Previously when the activityForResult is returned to the calling launching fragment, it would would automatically sign in the newly created user. Now we return CANCELLED to avoid this flow.

n/a

- View settings and book details and verify "Report issues..." is not longer visible.
- Create a juvenile library card with NYPL

n/a

n/a

@MalcolmMcFly

### DIFF
--- a/simplified-ui-accounts/src/main/res/layout/account.xml
+++ b/simplified-ui-accounts/src/main/res/layout/account.xml
@@ -257,37 +257,37 @@
         app:layout_constraintTop_toTopOf="parent" />
     </androidx.constraintlayout.widget.ConstraintLayout>
 
-<!--    <androidx.constraintlayout.widget.ConstraintLayout-->
-<!--      android:id="@+id/accountReportIssue"-->
-<!--      android:layout_width="match_parent"-->
-<!--      android:layout_height="wrap_content"-->
-<!--      android:background="?attr/selectableItemBackground">-->
+    <androidx.constraintlayout.widget.ConstraintLayout
+      android:visibility="gone"
+      android:id="@+id/accountReportIssue"
+      android:layout_width="match_parent"
+      android:layout_height="wrap_content"
+      android:background="?attr/selectableItemBackground">
 
-<!--      <View-->
-<!--        android:layout_width="0dp"-->
-<!--        android:layout_height="1dp"-->
-<!--        android:background="?android:attr/listDivider"-->
-<!--        app:layout_constraintEnd_toEndOf="parent"-->
-<!--        app:layout_constraintStart_toStartOf="parent"-->
-<!--        app:layout_constraintTop_toTopOf="parent" />-->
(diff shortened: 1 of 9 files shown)


simplified-ui-accounts/src/main/res/layout/account.xml
```diff
@@ -257,36 +257,36 @@
...
-        android:layout_marginBottom="16dp"
-        android:enabled="false"
-        android:text="@string/accountReportIssue"
-        app:layout_constraintBottom_toBottomOf="parent"
-        app:layout_constraintEnd_toEndOf="parent"
-        app:layout_constraintStart_toStartOf="parent"
-        app:layout_constraintTop_toTopOf="parent" />
-
-    </androidx.constraintlayout.widget.ConstraintLayout>
+<!--    <androidx.constraintlayout.widget.ConstraintLayout-->
```
io7m, line 260: Was this the only way it could be made invisible? Surely a call to `setVisibility(GONE)` would have had the same effect...

simplified-ui-catalog/src/main/java/org/nypl/simplified/ui/catalog/CatalogFragmentBookDetail.kt
```diff
@@ -192,8 +192,8 @@ class CatalogFragmentBookDetail : Fragment() {
       layout.findViewById(R.id.bookDetailButtons)
     this.related =
       layout.findViewById(R.id.bookDetailRelated)
-    this.report =
-      layout.findViewById(R.id.bookDetailReport)
+//    this.report =
+//      layout.findViewById(R.id.bookDetailReport)
```
io7m, line 197:
```
this.report.visibility = View.GONE
```

MalcolmMcFly:
@io7m do we get any performance improvements if we avoid drawing the widget in the first place?

io7m:
`View.VISIBLE` is the normal "show this UI element" setting. `View.INVISIBLE` means "show this UI element in the layout, but don't actually draw it", so you get an empty space where the UI element was. `View.GONE` means "effectively remove this view from the view hierarchy", so I think in performance terms it's the same as not having the element in the first place (plus you get to avoid commenting out a load of declarations :yum:).

MalcolmMcFly:
@io7m sounds good to me, will update. Can probably just set `visibility = gone` in the xml and avoid any class changes as well.

io7m:
Yep! Even better to do it in the XML, actually.

io7m:
Looks good!

PR #748, opened 2020-07-20
[SIMPLY-2880] Show the change account menu icon in the toolbar, when appropriate.

twaddington:
**What's this do?**
Moved the change account menu icon to the left side of the toolbar. This
icon will be displayed whenever the catalog tab is at the root. This
mirrors the behavior of the Twitter app. Tapping the icon now shows the
account list in a bottom sheet dialog.

https://jira.nypl.org/browse/SIMPLY-2880

Try adding, removing, and switching accounts.

n/a

n/a

@twaddington 

**Screenshots**
- [Screenshot_1595259089](https://user-images.githubusercontent.com/107117/87956152-89c23280-ca63-11ea-96b6-22c4e0eba24f.png)
- [Screenshot_1595259085](https://user-images.githubusercontent.com/107117/87956166-8dee5000-ca63-11ea-9f81-ae4d4e22ed22.png)


simplified-ui-catalog/src/main/java/org/nypl/simplified/ui/catalog/CatalogFragmentFeed.kt
```diff
@@ -687,23 +686,54 @@ class CatalogFragmentFeed : Fragment() {
...
+          }
+          else -> toolbarHost.toolbarUnsetArrow()
+        }
+      } else {
+        toolbar.navigationIcon = toolbarHost.toolbarIconBackArrow(context)
+        toolbar.navigationContentDescription = null
+        toolbar.setNavigationOnClickListener { navigationController.popBackStack() }
+      }
+    } catch (e: Exception) {
+      // Note: The call to findNavigationController may throw an IllegalArgumentException.
```
io7m, line 722: `BottomNavigator` sucks. :smile:

twaddington: lol, but that method is in our code 😅

io7m: It's in our code, but it's written the way it is due to limitations imposed by `BottomNavigator` (we can't guarantee, at any given moment, that we actually have access to the navigator). There's like a one line change to guarantee that we always have access, but it causes sporadic fragment transaction exceptions if we make it. :roll_eyes:

PR #750, opened 2020-07-20
Fixed back navigation when reviewing card info

MalcolmMcFly:
**What's this do?**
Fixed a bug when navigating back from the review screen.

Found in QA

Create a library card and when you are at the review screen, attempt to navigate back.


twaddington:
Looks like you have some ktlint failures.

PR #753, opened 2020-07-23
[SIMPLY-2858] Maintain the order of account providers as returned by the server.

twaddington:
**What's this do?**
The server may return account providers sorted with most relevant
results at the top. For example, provideres that are geographically
close to the patron's current location (as determined by IP Address).

This patch does two things:

1) Remove the local sorting of the providers in the account registry
fragment.
2) Use a LinkedHashMap wrapped with a synchronizedMap instead of the
ConcurrentHashMap that was used before.

https://jira.nypl.org/browse/SIMPLY-2858

Manually update the URI(s) in `AccountProviderSourceNYPLRegistry` *or* use Charles Proxy to intercept the request to `https://libraryregistry.librarysimplified.org/libraries` and add the query string `?_location=40.7,-74`.

Verify that the NYPL appears at the top of the list as expected.

n/a

n/a

@twaddington

### DIFF
--- a/simplified-accounts-registry/src/main/java/org/nypl/simplified/accounts/registry/AccountProviderRegistry.kt
+++ b/simplified-accounts-registry/src/main/java/org/nypl/simplified/accounts/registry/AccountProviderRegistry.kt
@@ -43,7 +43,7 @@ class AccountProviderRegistry private constructor(
   @Volatile
   private var statusRef: AccountProviderRegistryStatus = Idle
 
-  private val descriptions = ConcurrentHashMap<URI, AccountProviderDescription>()
+  private val descriptions = Collections.synchronizedMap(LinkedHashMap<URI, AccountProviderDescription>())
   private val descriptionsReadOnly = Collections.unmodifiableMap(this.descriptions)
   private val resolved = ConcurrentHashMap<URI, AccountProviderType>()
   private val resolvedReadOnly = Collections.unmodifiableMap(this.resolved)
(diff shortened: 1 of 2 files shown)


io7m:
I'm a little unsure as to how exactly this implements what was described. Removing the sorting makes sense, but otherwise changing the map implementation just puts the results in insertion order. I suppose if we make the assumption that the server always returns the results in order of relevance, then insertion order is usually equivalent to order by relevance. Having an insertion-order-preserving hash map implementation is a pretty subtle way to express that! :sweat: 

This map is also going to be populated from the on-disk cache, and I believe that cache can be in any order.

io7m:
What I'm saying is: It's not _wrong_, but I think it's also not obvious that it's right.

twaddington:
Yeah, I think the right way to do this would be to update the provider description to include an index that we could use to sort on. We could also create an index when we parse the server result and serialize that to disk so we can recreate that sort order.

You're right that this patch just updates our code so that the insertion order is maintained throughout our code. It fixes the issue for now but it also doesn't really codify the sorting in a specific way.

io7m:
I'm fine with merging this for now, but we'll have to revisit later. I think that's on the roadmap anyway.

twaddington:
> I'm fine with merging this for now, but we'll have to revisit later. I think that's on the roadmap anyway.

Okay, I'm going to merge this now but will keep your concerns in mind. I agree that it's not really a good long-term solution. Thanks for your feedback!

PR #754, opened 2020-07-23
Allow patrons to add a new account directly from the account picker dialog.

twaddington:
**What's this do?**
Adds an 'Add new account' option to the account switcher and makes the currently active account more obvious.

Improved Patron experience.

1. Tap the account icon in the upper left of the catalog.
2. Verify that the 'Add new account' option is shown at the bottom of the list.
3. Tap 'Add new account' and verify the account registry fragment is displayed, as expected.

n/a

n/a

@twaddington

**Screens**
- [Screenshot_1595529410](https://user-images.githubusercontent.com/107117/88324631-d4001980-ccd8-11ea-90e5-ca207952a5c7.png)

### DIFF
--- a/simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/AccountRegistryFragment.kt
+++ b/simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/AccountRegistryFragment.kt
@@ -248,7 +248,7 @@ class AccountRegistryFragment : Fragment() {
     if (host is ToolbarHostType) {
       host.toolbarClearMenu()
       host.toolbarSetTitleSubtitle(
-        title = this.requireContext().getString(R.string.accounts),
+        title = this.requireContext().getString(R.string.accountAdd),
         subtitle = ""
       )
       host.toolbarSetBackArrowConditionally(
--- a/simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/AccountPickerDialogFragment.kt
+++ b/simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/AccountPickerDialogFragment.kt
@@ -1,5 +1,8 @@
 package org.nypl.simplified.ui.accounts
 
+import android.graphics.Color
+import android.graphics.PorterDuff
+import android.graphics.PorterDuffColorFilter
 import android.graphics.Typeface
 import android.os.Bundle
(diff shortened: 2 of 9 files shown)


io7m:
This all looks good to me. One thing though: To be LFA-friendly, we need to respect the value of `SettingsConfigurationServiceType.allowAccountsAccess`. That is, if that property is `false`, users shouldn't be allowed to add or delete accounts. Normally this flag just hides access to the accounts section in the settings screen, but if they can now add accounts via the picker, then it needs to be handled there too.

twaddington:
> To be LFA-friendly, we need to respect the value of `SettingsConfigurationServiceType.allowAccountsAccess`. That is, if that property is `false`, users shouldn't be allowed to add or delete accounts. Normally this flag just hides access to the accounts section in the settings screen, but if they can now add accounts via the picker, then it needs to be handled there too. &mdash; @io7m 

Excellent feedback Mark, thanks. I'll make that change.

twaddington:
@io7m updated!

io7m:
There's one outstanding issue with this, but we don't need to correct it right away: The `ui-settings` module shouldn't depend on the `ui-catalog` module; it's bad form for a large implementation module to depend on another large implementation module. The problem here is that the `allowAccountsAccess` flag is only accessible via a service who's API is exported from the `ui-settings` module. What we need to do is move all of those configuration services into the `buildconfig` API.

Go ahead and merge this. I'll submit a PR for the buildconfig change afterwards.

simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/AccountRegistryFragment.kt
```diff
@@ -248,7 +248,7 @@ class AccountRegistryFragment : Fragment() {
     if (host is ToolbarHostType) {
       host.toolbarClearMenu()
       host.toolbarSetTitleSubtitle(
-        title = this.requireContext().getString(R.string.accounts),
+        title = this.requireContext().getString(R.string.accountAdd),
```
MalcolmMcFly, line 251: `R.string.account_add` ?

twaddington:
> Everything looks good to me. I didn't realize that we were camel-casing string resource names. If the project uses that convention then you can ignore that comment.

Yeah, all the string names in the `simplified-ui-accounts` module are camel cased.

PR #755, opened 2020-07-24
Consolidate configuration services

io7m:
**What's this do?**
This moves the existing catalog/settings configuration services
into one buildconfig service. This allows for removing the dependency
between the ui-settings and ui-catalog modules. Additionally, I ran the
"find unnecessary module dependencies" IDEA inspection and found (and
removed) a few unnecessary dependencies (mostly simplified-json-core).

Minor technical debt cleanup.

N/A

None.

N/A

@io7m ran the vanilla app


twaddington:
> Additionally, I ran the "find unnecessary module dependencies" IDEA inspection and found (and removed) a few unnecessary dependencies (mostly simplified-json-core). &mdash; @io7m 

Nice 🎉

io7m:
Forgot to remove the actual dependency though. :roll_eyes: 

One minute...

twaddington:
@io7m I don't see the `simplified-json-core` removals?

io7m:
You're right... I asked IDEA to remove the module dependencies, but it seems like it removed the dependencies from its internal model but didn't actually modify the `build.gradle` files.

io7m:
Turns out the IDEA inspection gives false positives due to transitive dependencies.

PR #756, opened 2020-07-25
[SIMPLY-2909] Fix filtering by collection when viewing local feeds.

twaddington:
**What's this do?**
This patch fixes several issues with the "Books" and "Holds" screens.

- The correct title is now shown for the "Holds" screen.
- A bug was fixed that caused the "collection" filter to not work correctly.
- A bug was fixed that caused the "sort by" facet to not work correctly when also filtering by collection.
- The facet groups now show options sorted by name.
- The account provider and book format are now shown when viewing a feed that has been collected from multiple accounts.

https://jira.nypl.org/browse/SIMPLY-2909

1) Navigate to the "Books" or "Holds" tab and verify that filtering by collection works as expected.
  a. Also verify that sorting works correctly.

n/a

n/a


simplified-ui-catalog/src/main/java/org/nypl/simplified/ui/catalog/CatalogBookAvailabilityStrings.kt
```diff
@@ -25,50 +15,6 @@ import java.util.concurrent.TimeUnit
...
 
-  /**
-   * Produce a human-readable string for the given OPDS availability value.
-   *
-   * @param resources The application resources
-   * @param availability The availability value
-   *
-   * @return A descriptive string
-   */
-  fun opdsAvailabilityString(
```
io7m, line 36: BEGONE!!!

_ahem_

Glad to see this gone.

PR #758, opened 2020-07-30
SIMPLY-2925: mailto: links in EULA don't work

MalcolmMcFly:
**What's this do?**
This adds a new custom WebViewClient that can handle `mailto:` inside of a webview

https://jira.nypl.org/browse/SIMPLY-2925

Launch the app fresh, and click on the email link in the EULA.

n/a

n/a

@MalcolmMcFly

### DIFF
--- a/simplified-ui-splash/src/main/java/org/nypl/simplified/ui/splash/MailtoWebViewClient.java
+++ b/simplified-ui-splash/src/main/java/org/nypl/simplified/ui/splash/MailtoWebViewClient.java
@@ -25,19 +25,15 @@ public class MailtoWebViewClient extends WebViewClient {
         if (url.startsWith(MailTo.MAILTO_SCHEME)) {
             final Activity activity = mActivityRef.get();
             if (activity != null) {
-                try {
-                    MailTo mt = MailTo.parse(url);
-                    Intent i = newEmailIntent(activity, mt.getTo(), mt.getSubject());
-                    if (i.resolveActivity(activity.getPackageManager()) != null) {
-                        activity.startActivity(i);
-                    }
-                } catch (Exception e) {
-                    e.printStackTrace();
+                MailTo mt = MailTo.parse(url);
+                Intent i = newEmailIntent(activity, mt.getTo(), mt.getSubject());
+                if (i.resolveActivity(activity.getPackageManager()) != null) {
+                    activity.startActivity(i);
                 }
                 return true;
             }
         } else {
-            view.loadUrl(url);
+            return false;
         }
         return true;
     }
--- a/simplified-ui-splash/src/main/java/org/nypl/simplified/ui/splash/MailtoWebViewClient.java
+++ b/simplified-ui-splash/src/main/java/org/nypl/simplified/ui/splash/MailtoWebViewClient.java
@@ -4,6 +4,7 @@ import android.app.Activity;
 import android.content.Context;
 import android.content.Intent;
 import android.net.MailTo;
+import android.net.Uri;
 import android.webkit.WebView;
 import android.webkit.WebViewClient;
 
@@ -21,13 +22,18 @@ public class MailtoWebViewClient extends WebViewClient {
 
     @Override
     public boolean shouldOverrideUrlLoading(WebView view, String url) {
-        if (url.startsWith("mailto:")) {
+        if (url.startsWith(MailTo.MAILTO_SCHEME)) {
             final Activity activity = mActivityRef.get();
             if (activity != null) {
-                MailTo mt = MailTo.parse(url);
-                Intent i = newEmailIntent(activity, mt.getTo(), mt.getSubject(), mt.getBody(), mt.getCc());
-                activity.startActivity(i);
-                view.reload();
+                try {
+                    MailTo mt = MailTo.parse(url);
+                    Intent i = newEmailIntent(activity, mt.getTo(), mt.getSubject());
+                    if (i.resolveActivity(activity.getPackageManager()) != null) {
+                        activity.startActivity(i);
+                    }
+                } catch (Exception e) {
+                    e.printStackTrace();
+                }
                 return true;
             }
         } else {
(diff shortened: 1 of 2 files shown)


simplified-ui-splash/src/main/java/org/nypl/simplified/ui/splash/MailtoWebViewClient.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    @Override
+    public boolean shouldOverrideUrlLoading(WebView view, String url) {
+        if (url.startsWith("mailto:")) {
+            final Activity activity = mActivityRef.get();
+            if (activity != null) {
+                MailTo mt = MailTo.parse(url);
+                Intent i = newEmailIntent(activity, mt.getTo(), mt.getSubject(), mt.getBody(), mt.getCc());
+                activity.startActivity(i);
```
twaddington, line 29: It's possible `startActivity` could throw an exception if there is no Mail app on the device. Kind of an unusual situation but always nice to handle this exception just in case.

https://developer.android.com/reference/android/content/ActivityNotFoundException

simplified-ui-splash/src/main/java/org/nypl/simplified/ui/splash/MailtoWebViewClient.java
```diff
@@ -0,0 +1,48 @@
...
+                view.reload();
```
twaddington, line 30: What is the purpose of reloading the WebView content here?

simplified-ui-splash/src/main/java/org/nypl/simplified/ui/splash/MailtoWebViewClient.java
```diff
@@ -0,0 +1,48 @@
...
+        return true;
+    }
+
+    private Intent newEmailIntent(Context context, String address, String subject, String body, String cc) {
+        Intent intent = new Intent(Intent.ACTION_SEND);
+        intent.putExtra(Intent.EXTRA_EMAIL, new String[] { address });
+        intent.putExtra(Intent.EXTRA_TEXT, body);
+        intent.putExtra(Intent.EXTRA_SUBJECT, subject);
+        intent.putExtra(Intent.EXTRA_CC, cc);
+        intent.setType("message/rfc822");
```
twaddington, line 45: This is a bit of a hack (I found the StackOverflow that suggested this).

It's always been a bit of black magic to figure out the right combination of Intent extras to compose an email.

What you really want here is `Intent.ACTION_SENDTO` instead of the more generic `Intent.ACTION_SEND`. Try something like this:

```java
    Intent intent = new Intent(Intent.ACTION_SENDTO);
    intent.setData(Uri.parse("mailto:")); // only email apps should handle this
    intent.putExtra(Intent.EXTRA_EMAIL, addresses);
    intent.putExtra(Intent.EXTRA_SUBJECT, subject);
    intent.putExtra(Intent.EXTRA_CC, cc);
    if (intent.resolveActivity(getPackageManager()) != null) {
        startActivity(intent);
    }
```

As a bonus this accounts for the lack of a Mail app as well.

https://developer.android.com/guide/components/intents-common#ComposeEmail

simplified-ui-splash/src/main/java/org/nypl/simplified/ui/splash/MailtoWebViewClient.java
```diff
@@ -0,0 +1,48 @@
...
+public class MailtoWebViewClient extends WebViewClient {
+    private final WeakReference<Activity> mActivityRef;
+
+    public MailtoWebViewClient(Activity activity) {
+        mActivityRef = new WeakReference<>(activity);
+    }
+
+    @Override
+    public boolean shouldOverrideUrlLoading(WebView view, String url) {
+        if (url.startsWith("mailto:")) {
```
twaddington, line 24: You could use `if (url.startsWith(MailTo.MAILTO_SCHEME)) {` here for extra safety.

simplified-ui-splash/src/main/java/org/nypl/simplified/ui/splash/MailtoWebViewClient.java
```diff
@@ -0,0 +1,48 @@
...
+            final Activity activity = mActivityRef.get();
+            if (activity != null) {
+                MailTo mt = MailTo.parse(url);
```
twaddington, line 27: Today I learned about the `MailTo` class 😝

simplified-ui-splash/src/main/java/org/nypl/simplified/ui/splash/MailtoWebViewClient.java
```diff
@@ -0,0 +1,52 @@
...
+                    if (i.resolveActivity(activity.getPackageManager()) != null) {
+                        activity.startActivity(i);
+                    }
+                } catch (Exception e) {
+                    e.printStackTrace();
+                }
+                return true;
+            }
+        } else {
+            view.loadUrl(url);
```
twaddington, line 40: I tried to leave this comment earlier, but apparently it didn't take for some reason.

The docs specifically warn about calling `loadUrl` and then returning `true` from this method. Instead, you should just return `false` and let the WebView continue loading as normal. Like this:

```java
if (url.startsWith(MailTo.MAILTO_SCHEME)) {
    // ...
    return true;
}
return false;
```

> Note: Do not call WebView#loadUrl(String) with the same URL and then return true. This unnecessarily cancels the current load and starts a new load with the same URL. The correct way to continue loading a given URL is to simply return false, without calling WebView#loadUrl(String).

https://developer.android.com/reference/android/webkit/WebViewClient#shouldOverrideUrlLoading(android.webkit.WebView,%20java.lang.String)

simplified-ui-splash/src/main/java/org/nypl/simplified/ui/splash/MailtoWebViewClient.java
```diff
@@ -0,0 +1,52 @@
...
+    public MailtoWebViewClient(Activity activity) {
+        mActivityRef = new WeakReference<>(activity);
+    }
+
+    @Override
+    public boolean shouldOverrideUrlLoading(WebView view, String url) {
+        if (url.startsWith(MailTo.MAILTO_SCHEME)) {
+            final Activity activity = mActivityRef.get();
+            if (activity != null) {
+                try {
```
twaddington, line 28: You can drop the `try / catch` here since you're calling `resolveActivity ` instead.

simplified-ui-splash/src/main/java/org/nypl/simplified/ui/splash/MailtoWebViewClient.java
```diff
@@ -0,0 +1,48 @@
...
+                Intent i = newEmailIntent(activity, mt.getTo(), mt.getSubject());
+                if (i.resolveActivity(activity.getPackageManager()) != null) {
+                    activity.startActivity(i);
+                }
+                return true;
+            }
+        } else {
+            return false;
+        }
+        return true;
```
twaddington, line 38: This return will never execute so you can remove it.

MalcolmMcFly: @twaddington which one? the `return false;`?

twaddington: Ah, I didn't see the case where activity could be null. We talked in Slack so I'll resolve this.

PR #759, opened 2020-07-30
SIMPLY-2928: Accept 'application/json' Content-Type for overdrive audiobooks.

twaddington:
**What's this do?**
Fixes an issue where overdrive audiobooks were failing to download due to an unexpected content type.

https://jira.nypl.org/browse/SIMPLY-2928

Attempt to download and listen to an Overdrive audiobook.

n/a

n/a

@twaddington tested in the vanilla app using test credentials from Leonard.


twaddington:
> This change looks simple enough assuming you have tested and verify that it is working.

Yep! I'll drop a video into Slack.

PR #761, opened 2020-08-01
[SIMPLY-2931] Fixed an issue where the wrong book availability was shown after placing or revoking a hold.

twaddington:
**What's this do?**
Show the correct book status after placing or revoking a hold.

https://jira.nypl.org/browse/SIMPLY-2931

Steps to reproduce the error and verify the fix are in the Jira ticket.

n/a

n/a

### DIFF
--- a/simplified-ui-catalog/src/main/java/org/nypl/simplified/ui/catalog/CatalogFragmentBookDetail.kt
+++ b/simplified-ui-catalog/src/main/java/org/nypl/simplified/ui/catalog/CatalogFragmentBookDetail.kt
@@ -32,7 +32,7 @@ import org.nypl.simplified.books.book_registry.BookWithStatus
 import org.nypl.simplified.books.controller.api.BooksControllerType
 import org.nypl.simplified.books.covers.BookCoverProviderType
 import org.nypl.simplified.buildconfig.api.BuildConfigurationServiceType
-import org.nypl.simplified.feeds.api.FeedEntry
+import org.nypl.simplified.feeds.api.FeedEntry.FeedEntryOPDS
 import org.nypl.simplified.navigation.api.NavigationControllers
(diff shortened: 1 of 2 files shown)


simplified-ui-catalog/src/main/java/org/nypl/simplified/ui/catalog/CatalogFragmentBookDetail.kt
```diff
@@ -249,9 +249,9 @@ class CatalogFragmentBookDetail : Fragment() {
 
     val status =
       this.bookRegistry.bookOrNull(this.parameters.bookID)
-        ?: this.synthesizeBookWithStatus()
+        ?: this.synthesizeBookWithStatus(this.parameters.feedEntry)
 
-    this.onBookStatusUI(status)
+    this.onBookChangedUI(status)
```
twaddington, line 254: This method was renamed solely for consistency with the same method in `CatalogPagedViewHolder`.

simplified-ui-catalog/src/main/java/org/nypl/simplified/ui/catalog/CatalogFragmentBookDetail.kt
```diff
@@ -264,19 +264,23 @@ class CatalogFragmentBookDetail : Fragment() {
 
     this.bookRegistrySubscription =
       this.bookRegistry.bookEvents()
-        .subscribe(this::onBookStatusEvent)
+        .subscribe(this::onBookChanged)
```
twaddington, line 267: This method was renamed solely for consistency with the same method in `CatalogPagedViewHolder`.

io7m:
I love this and am frightened of it (and that whole area of the code in general).

Let's do it!

PR #762, opened 2020-08-01
[SIMPLY-2934] Load covers from feed entry if no local cover exists.

twaddington:
**What's this do?**
Fixes an issue where covers were not being shown for held or loaned books.

https://jira.nypl.org/browse/SIMPLY-2934

Book covers should be shown for items in your Books or Holds feeds.

n/a

n/a

Yes, @twaddington ran the vanilla app.

### DIFF
--- a/simplified-books-covers/src/main/java/org/nypl/simplified/books/covers/BookCoverProvider.kt
+++ b/simplified-books-covers/src/main/java/org/nypl/simplified/books/covers/BookCoverProvider.kt
@@ -29,7 +29,7 @@ class BookCoverProvider private constructor(
   private val badgeLookup: BookCoverBadgeLookupType
 ) : BookCoverProviderType {
 
-  private val log: Logger = LoggerFactory.getLogger(BookCoverProvider::class.java)
+  private val logger: Logger = LoggerFactory.getLogger(BookCoverProvider::class.java)
   private val coverTag: String = "cover"
   private val thumbnailTag: String = "thumbnail"
 
@@ -81,7 +81,7 @@ class BookCoverProvider private constructor(
 
     val badgePainter = BookCoverBadgePainter(entry, this.badgeLookup)
     if (uriSpecified != null) {
-      this.log.debug("{}: {}: loading specified uri {}", tag, entry.bookID, uriSpecified)
+      this.logger.debug("{}: {}: loading specified uri {}", tag, entry.bookID, uriSpecified)
 
       val fallbackToGeneration = object : Callback {
         override fun onSuccess() {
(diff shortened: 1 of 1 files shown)


simplified-books-covers/src/main/java/org/nypl/simplified/books/covers/BookCoverProvider.kt
```diff
@@ -129,21 +128,15 @@ class BookCoverProvider private constructor(
...
+    val bookWithStatus =
+      this.bookRegistry.bookOrNull(entry.bookID)
+    return bookWithStatus?.book?.cover?.toURI() ?: mapOptionToNull(entry.feedEntry.cover)
   }
 
   private fun thumbnailURIOf(entry: FeedEntry.FeedEntryOPDS): URI? {
-    val bookOpt = this.bookRegistry.book(entry.bookID)
-    if (bookOpt is Some<BookWithStatus>) {
-      val book = bookOpt.get()
-      return book.book.cover?.toURI()
```
twaddington, line 144: This is a local file URI that is supposed to contain a reference to a cover. However, after looking through the code this value appears to never be written to. I suspect this must be to support apps like LFA?

Either way this value always returns null.

The fix was pretty simple and doesn't change the logic much. It just falls back to loading from the feed entry URI if the local file URI is not set.

io7m: That in itself is a bug. The borrow task is supposed to grab the book cover image at the end and save it to disk. The `Book` value is then updated to hold a non-null `File` value referencing that on-disk copy...

twaddington: Ah, I see the `BookCoverFetchTask` now. Let me look into that.

Is there a reason we want to cache the cover to disk instead of just fetching from the feed entry? Picasso has a built-in disk cache.

twaddington: So, I'd actually like to merge this fix in as-is and investigate the reason that value isn't being written separately. This is a good safe fix to make anyway and I think the underlying issue isn't critical to get into the release.

io7m: I agree!

io7m:
> Is there a reason we want to cache the cover to disk instead of just fetching from the feed entry? Picasso has a built-in disk cache.

Only that the Picasso disk cache isn't guaranteed to survive having the app be offline for a few days. It's not vital.

PR #767, opened 2020-08-04
[SIMPLY-2934] Cache thumbnail images to disk for fulfilled feed items. 

twaddington:
**What's this do?**
There is some existing code that downloads book cover images to disk for items you have checked out. While not causing any known issues, this code was fragile in places so I've cleaned some of that up. Additionally, covers were being cached but not thumbnails.

Follow-up to https://github.com/NYPL-Simplified/Simplified-Android-Core/pull/762#discussion_r464542858.

n/a

n/a

n/a

@twaddington tested with the vanilla app.

### DIFF
--- a/simplified-books-controller/src/main/java/org/nypl/simplified/books/controller/BookCoverFetchTask.kt
+++ b/simplified-books-controller/src/main/java/org/nypl/simplified/books/controller/BookCoverFetchTask.kt
@@ -8,6 +8,8 @@ import org.nypl.simplified.books.book_database.api.BookDatabaseEntryType
 import org.nypl.simplified.books.book_registry.BookStatusDownloadErrorDetails
 import org.nypl.simplified.books.book_registry.BookWithStatus
 import org.nypl.simplified.books.bundled.api.BundledURIs.BUNDLED_CONTENT_SCHEME
+import org.nypl.simplified.books.controller.BookCoverFetchTask.Type.COVER
+import org.nypl.simplified.books.controller.BookCoverFetchTask.Type.THUMBNAIL
 import org.nypl.simplified.http.core.HTTPAuthType
 import org.nypl.simplified.http.core.HTTPResultError
 import org.nypl.simplified.http.core.HTTPResultException
@@ -20,6 +22,7 @@ import java.io.FileNotFoundException
 import java.io.FileOutputStream
 import java.io.InputStream
 import java.net.URI
+import java.util.Locale
 import java.util.concurrent.Callable
 
 /**
@@ -30,8 +33,17 @@ class BookCoverFetchTask(
   private val services: BookTaskRequiredServices,
   private val databaseEntry: BookDatabaseEntryType,
   private val feedEntry: OPDSAcquisitionFeedEntry,
+  private val type: Type,
   private val httpAuth: OptionType<HTTPAuthType>
 ) : Callable<TaskResult<BookStatusDownloadErrorDetails, Unit>> {
+
+  enum class Type {
+    COVER, THUMBNAIL;
+
+    override fun toString(): String {
+      return super.toString().toLowerCase(Locale.US)
+    }
+  }
 
   private val logger =
     LoggerFactory.getLogger(BookCoverFetchTask::class.java)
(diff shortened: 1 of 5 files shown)


simplified-books-database/src/main/java/org/nypl/simplified/books/book_database/BookDatabase.kt
```diff
@@ -113,8 +128,8 @@ class BookDatabase private constructor(
           Book(
             id = id,
             account = this.owner,
-            cover = null,
-            thumbnail = null,
+            cover = cover,
```
twaddington, line 131: This ensures we don't blow away the cover and thumbnail information when updating a stored entry. I'm unsure if we currently do this anywhere, but it was possible.

simplified-books-database/src/main/java/org/nypl/simplified/books/book_database/BookDatabase.kt
```diff
@@ -268,3 +273,10 @@ class BookDatabase private constructor(
     }
   }
 }
+
+/** Return the file, or null if it does not exist or is not a file. */
+
+private fun fileOrNull(bookDir: File, filename: String) = File(bookDir, filename)
```
twaddington, line 279: This just DRYs up some existing code.

simplified-books-database/src/main/java/org/nypl/simplified/books/book_database/BookDatabaseEntry.kt
```diff
@@ -218,6 +233,8 @@ internal class BookDatabaseEntry internal constructor(
   }
 
   companion object {
+    const val COVER_FILENAME = "cover.jpg"
```
twaddington, line 236: Moved these to const values since they're used in several places.

simplified-books-controller/src/main/java/org/nypl/simplified/books/controller/BookCoverFetchTask.kt
```diff
@@ -61,7 +76,7 @@ class BookCoverFetchTask(
         this.taskRecorder.finishSuccess(Unit)
       }
     } catch (e: Throwable) {
-      this.logger.error("failed to fetch cover: ", e)
+      this.logger.error("failed to fetch ${this.type}: ", e)
```
io7m, line 79: Not sure if you know, but the SLF4J API allows for this:

```
this.logger.error("failed to fetch {}: ", this.type, e)
```

The `{}` syntax is just positional argument formatting. Additionally, if there are `N` occurrences of `{}` in the string, and the `N + 1`th argument is an exception, then the exception stacktrace is printed as part of the message.

Using `{}` is mildly more efficient because SLF4J lazily evaluates `toString()` on each the arguments, and it won't do it if logging is disable for that logger.

I'm only putting this because you might not know about it. It's bordering on micro-optimization, so it really doesn't matter here.

twaddington: Yep, the reason I didn't use formatting args here is because Logger doesn't seem to have a method overload of `error` that accepts varargs and a Throwable.

I've seen other logging APIs get around this by making the Throwable the first argument so they can have varargs at the end. 

```
public static void e(Throwable t,
                     String message,
                     Object... args)
```

io7m: Not sure if I'm misunderstanding. All of the SLF4J methods implicitly accept a `Throwable` as the last argument of the `Object...`. 
I'd link to the actual docs, but the SLF4J site isn't responding. This article covers it, though:

https://www.baeldung.com/slf4j-log-exceptions

> In presence of multiple parameters, if the last argument in a logging statement is an exception, then SLF4J will presume that the user wants the last argument to be treated as an exception instead of a simple parameter:

```
logger.error("{}, {}! An exception occurred!", 
  "Hello", 
  "World", 
  new Exception("Custom exception"));
```

> In the above snippet, the String message will be formatted based on the passed object details. We've used curly braces as placeholders for String parameters passed to the method.

> In this case, the output will be:

```
ERROR packageName.NameOfTheClass - Hello, World! An exception occurred!
java.lang.Exception: Custom exception 
  at packageName.NameOfTheClass.methodName(NameOfTheClass.java:lineNo)
```

twaddington: > In presence of multiple parameters, if the last argument in a logging statement is an exception, then SLF4J will presume that the user wants the last argument to be treated as an exception instead of a simple parameter.

Ah! That's unclear from the API. I would have expected it to throw an exception due to the wrong number of arguments. I will refactor.

PR #768, opened 2020-08-05
[SIMPLY-2943] Be more lenient about content types when creating format handles for book db entries.

twaddington:
**What's this do?**
Fixes an issue where Feedbook audiobooks would fail with a "database entry does not have a format handle for {}" error.

https://jira.nypl.org/browse/SIMPLY-2943

An account for "St. Mary's County Library" should be used.

Ask Leonard or Tristan for test credentials.

n/a

n/a

Yes, @twaddington ran the Vanilla app.

### DIFF
--- a/simplified-books-database/src/main/java/org/nypl/simplified/books/book_database/BookDatabaseEntry.kt
+++ b/simplified-books-database/src/main/java/org/nypl/simplified/books/book_database/BookDatabaseEntry.kt
@@ -252,29 +252,36 @@ internal class BookDatabaseEntry internal constructor(
       contentTypes: Set<MIMEType>
     ) {
       for (contentType in contentTypes) {
-        for (formatDefinition in constructors.keys) {
-          val constructor = constructors[formatDefinition]!!
-          if (formatDefinition.supportedContentTypes().contains(contentType)) {
-            if (!existingFormats.containsKey(constructor.classType)) {
-              val bookID = owner.book.id
-              logger.debug("[{}]: instantiating format {} for content type {}",
-                bookID.brief(),
-                constructor.classType.simpleName,
-                contentType)
+        for ((format, constructor) in constructors) {
+          if (format.supports(contentType)) {
 
-              val params =
-                DatabaseFormatHandleParameters(
-                  context = context,
-                  bookID = owner.book.id,
-                  directory = ownerDirectory,
-                  onUpdated = onUpdate,
-                  entry = owner,
-                  contentType = contentType
-                )
-
-              existingFormats[constructor.classType] = constructor.constructor.invoke(params)
-              return
+            // Skip if handler already exists for type
+            if (existingFormats.containsKey(constructor.classType)) {
+              logger.debug(
+                "[{}]: skipping duplicate format {} for content type {}",
+                owner.book.id.brief(), constructor.classType.simpleName, contentType
+              )
+              continue
             }
+
+            // Add new handler for type
+            logger.debug(
+              "[{}]: instantiating format {} for content type {}",
+              owner.book.id.brief(), constructor.classType.simpleName, contentType
+            )
+
+            val params =
+              DatabaseFormatHandleParameters(
+                context = context,
+                bookID = owner.book.id,
+                directory = ownerDirectory,
+                onUpdated = onUpdate,
+                entry = owner,
+                contentType = contentType
+              )
+
+            existingFormats[constructor.classType] = constructor.constructor.invoke(params)
+            return
           }
         }
       }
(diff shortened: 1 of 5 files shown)


simplified-books-database/src/main/java/org/nypl/simplified/books/book_database/BookDatabaseEntry.kt
```diff
@@ -252,29 +252,36 @@ internal class BookDatabaseEntry internal constructor(
...
-                  directory = ownerDirectory,
-                  onUpdated = onUpdate,
-                  entry = owner,
-                  contentType = contentType
-                )
-
-              existingFormats[constructor.classType] = constructor.constructor.invoke(params)
-              return
+        for ((format, constructor) in constructors) {
+          if (format.supports(contentType)) {
```
twaddington, line 256: The actual fix for SIMPLY-2943 is on this line: Instead of matching on an exact MIMEType using `.contains(type)` we match using the existing `.supports(type)` method (which is lenient and does a comparison on MIMEType.fullType only, excluding any extra parameters).

simplified-books-database/src/main/java/org/nypl/simplified/books/book_database/BookDatabaseEntry.kt
```diff
@@ -252,29 +252,36 @@ internal class BookDatabaseEntry internal constructor(
...
-                  bookID = owner.book.id,
-                  directory = ownerDirectory,
-                  onUpdated = onUpdate,
-                  entry = owner,
-                  contentType = contentType
-                )
-
-              existingFormats[constructor.classType] = constructor.constructor.invoke(params)
-              return
+        for ((format, constructor) in constructors) {
```
twaddington, line 255: Used destructuring to clean this up a little.

simplified-books-database/src/main/java/org/nypl/simplified/books/book_database/BookDatabaseEntry.kt
```diff
@@ -252,29 +252,36 @@ internal class BookDatabaseEntry internal constructor(
...
+          if (format.supports(contentType)) {
+
+            // Skip if handler already exists for type
+            if (existingFormats.containsKey(constructor.classType)) {
+              logger.debug(
+                "[{}]: skipping duplicate format {} for content type {}",
+                owner.book.id.brief(), constructor.classType.simpleName, contentType
+              )
+              continue
```
twaddington, line 264: Eliminated one level of nesting with this `continue`.

simplified-ui-catalog/src/main/java/org/nypl/simplified/ui/catalog/CatalogFragmentBookDetail.kt
```diff
@@ -355,13 +355,13 @@ class CatalogFragmentBookDetail : Fragment() {
...
     val context = this.requireContext()
     this.format.text = when (feedEntry.probableFormat) {
-      null,
       BookFormats.BookFormatDefinition.BOOK_FORMAT_EPUB ->
         context.getString(R.string.catalogBookFormatEPUB)
       BookFormats.BookFormatDefinition.BOOK_FORMAT_AUDIO ->
         context.getString(R.string.catalogBookFormatAudioBook)
       BookFormats.BookFormatDefinition.BOOK_FORMAT_PDF ->
         context.getString(R.string.catalogBookFormatPDF)
+      else -> ""
```
io7m, line 364: I don't like this use of `else`. The reason for actually naming all of the cases specifically instead of using a fallback `else` is so that, when we add a new book format in the future, the compiler will highlight this `when` expression as now being non-total.

simplified-ui-catalog/src/main/java/org/nypl/simplified/ui/catalog/CatalogPagedViewHolder.kt
```diff
@@ -202,13 +202,13 @@ class CatalogPagedViewHolder(
...
 
     this.idleMeta.text = when (item.probableFormat) {
-      null,
       BookFormats.BookFormatDefinition.BOOK_FORMAT_EPUB ->
         context.getString(R.string.catalogBookFormatEPUB)
       BookFormats.BookFormatDefinition.BOOK_FORMAT_AUDIO ->
         context.getString(R.string.catalogBookFormatAudioBook)
       BookFormats.BookFormatDefinition.BOOK_FORMAT_PDF ->
         context.getString(R.string.catalogBookFormatPDF)
+      else -> ""
```
io7m, line 211: Ditto!

simplified-ui-catalog/src/main/java/org/nypl/simplified/ui/catalog/CatalogFragmentBookDetail.kt
twaddington: Makes sense, okay to switch it to an explicit `null -> ""`?

twaddington: Good feedback! Changing to explicit `null -> ""`.

io7m:
You've got some unused imports, but the rest looks fine. :+1:

twaddington:
> You've got some unused imports, but the rest looks fine. 👍

doh 😭

PR #770, opened 2020-08-08
Upgrade bottom navigator and set background color

io7m:
**What's this do?**
This upgrades the bottom navigator to 1.7, and fixes the black
background issue we were seeing.

We were getting an odd black background on the bottom navigator after the recent R2 work.

With your eyes.

None.

N/A

Before: 
device-2020-08-07-193441
After:
device-2020-08-08-092357

### DIFF
--- a/simplified-main/src/main/res/layout/main_tabbed_host.xml
+++ b/simplified-main/src/main/res/layout/main_tabbed_host.xml
@@ -27,6 +27,7 @@
     android:id="@+id/bottomNavigator"
     android:layout_width="0dp"
     android:layout_height="64dp"
+    android:background="?android:attr/windowBackground"
     app:layout_constraintBottom_toBottomOf="parent"
     app:layout_constraintEnd_toEndOf="parent"
     app:layout_constraintStart_toStartOf="parent"
--- a/build.gradle
+++ b/build.gradle
@@ -114,7 +114,7 @@ ext.libraries = [
   androidXTestRunner               : "androidx.test:runner:1.1.0",
   autoValue                        : "com.google.auto.value:auto-value-annotations:1.7",
   autoValueProcessor               : "com.google.auto.value:auto-value:1.7",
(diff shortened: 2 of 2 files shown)


simplified-main/src/main/res/layout/main_tabbed_host.xml
```diff
@@ -27,6 +27,7 @@
     android:id="@+id/bottomNavigator"
     android:layout_width="0dp"
     android:layout_height="64dp"
+    android:background="?android:attr/windowBackground"
```
twaddington, line 30: I'm not sure `windowBackground` is the right thing to use here. Typically this will just be transparent. Can you point me to some documentation that suggests we should set this attribute on the `BottomNavigationView`?

io7m: Unfortunately, I don't have a good answer to this. I looked at various posts online and this seemed to be what people were using. I tried various colors and this was the only one I could get to work correctly.

io7m: I should point out actually that you can see this problem in the UI designer if you set the theme to one of our app themes. The bottom navigator appears to be black if you don't set the color. If you set the bottom navigator color to red, you'll get a red bar with black borders. If you set the bottom navigator to the window background, you'll get a transparent view without borders. Frankly, I'd had enough of going down the rabbit hole at that point. :laughing:

twaddington: Yeah, this seems fine for now.

I see we set a value for windowBackground in styles. Historically, you could set an image for windowBackground (like a splash image) and it would display while your application is spinning up before your first activity's onCreate method is called. I don't think this technique is used so much anymore but it was a thing for a while.

I think the root problem was probably that the navigator just didn't have a background set. You sometimes get that black appearance if transparent is set on a view and there's nothing behind it.

io7m: Only in Android could _white_ + _transparent_ = _black_...

PR #771, opened 2020-08-10
Mitigate a potential crash on startup by waiting for service directory creation to complete.

twaddington:
**What's this do?**
Wait up to 30 seconds for the service directory to finish spinning up. Ideally the service directory is initialized serially, but quickly; or lazily, when needed. Either way we should block until this process finishes. This fix is a bit of a hack but follows a pattern already used elsewhere within the application.

https://console.firebase.google.com/project/simplye-nypl/crashlytics/app/android:org.nypl.simplified.simplye/issues/0da057da8867a50cf3d7de37b1cd6b84

Make sure the app starts up normally.

n/a

n/a

### DIFF
--- a/simplified-main/src/main/java/org/nypl/simplified/main/MainFragment.kt
+++ b/simplified-main/src/main/java/org/nypl/simplified/main/MainFragment.kt
@@ -37,6 +37,7 @@ import org.nypl.simplified.ui.settings.SettingsNavigationControllerType
 import org.nypl.simplified.ui.thread.api.UIThreadServiceType
 import org.nypl.simplified.ui.toolbar.ToolbarHostType
 import org.slf4j.LoggerFactory
+import java.util.concurrent.TimeUnit
 
 /**
  * The main application fragment.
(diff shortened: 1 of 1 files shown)


io7m:
Looks fine.

This is one of those "no matter what we do, Android won't let us do this correctly" situations.

If we initialize services serially, we block the application startup on a blank screen. If we initialize them lazily, that leads to all kinds of nasty problems where you get service errors at the point of use instead of when the application starts up (meaning that problems can go undiscovered for ages). This is yet another consequence of the insane "lifecycle" programming model; literally every other system outside of Android allows you to reliably have some kind of "loading screen" that will always be executed and can't accidentally be bypassed.

twaddington:
> If we initialize services serially, we block the application startup on a blank screen. &mdash; @io7m 

You can actually use `windowBackground` to show a splash image while your application is spinning up. Funny coincidence with the discussion on your pull-request.

https://android.jlelse.eu/the-complete-android-splash-screen-guide-c7db82bce565

PR #773, opened 2020-08-10
Fix a few possible crashes on app startup as seen in Firebase

twaddington:
- https://console.firebase.google.com/project/simplye-nypl/crashlytics/app/android:org.nypl.simplified.simplye/issues/01496df5f6758b6ba270f6ed3f8fc46f
- https://console.firebase.google.com/project/simplye-nypl/crashlytics/app/android:org.nypl.simplified.simplye/issues/43fb299ddf1974a4221a4c69500daeb5
- https://console.firebase.google.com/project/simplye-nypl/crashlytics/app/android:org.nypl.simplified.simplye/issues/f2095f5c847a085b901dcfb64945a4dc

Run the app and make sure it starts up successfully.

n/a

n/a

### DIFF
--- a/simplified-main/src/main/java/org/nypl/simplified/main/MainActivity.kt
+++ b/simplified-main/src/main/java/org/nypl/simplified/main/MainActivity.kt
@@ -337,6 +337,10 @@ class MainActivity :
   }
 
   private fun openCatalog() {
+    // Sanity check; we were seeing some crashes on startup when performing a
+    // transaction after the fragment manager had been destroyed.
+    if (this.isFinishing || this.supportFragmentManager.isDestroyed) return
+
     ViewModelProviders.of(this)
       .get(MainFragmentViewModel::class.java)
       .clearHistory = true
--- a/simplified-main/src/main/java/org/nypl/simplified/main/MainActivity.kt
+++ b/simplified-main/src/main/java/org/nypl/simplified/main/MainActivity.kt
@@ -7,6 +7,7 @@ import android.os.Bundle
 import android.view.View
 import androidx.appcompat.app.AppCompatActivity
 import androidx.appcompat.widget.Toolbar
+import androidx.core.content.ContextCompat
 import androidx.fragment.app.FragmentManager
 import androidx.lifecycle.ViewModelProviders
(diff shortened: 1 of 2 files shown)


io7m:
I don't object to these changes in general, but looking at some of them, I'm wondering if we're papering over a more fundamental issue.

The splash screen code is on life support until the onboarding work, so I don't care too much about what happens to that. But is there something fundamentally wrong with the way I'm using fragments? I know fragments cause lifecycle issues, but our usage of them seems pretty straightforward and idiomatic... Are they really this broken by design?

twaddington:
> I don't object to these changes in general, but looking at some of them, I'm wondering if we're papering over a more fundamental issue.

Yeah, these are definitely bandaid fixes for a more fundamental problem.

> But is there something fundamentally wrong with the way I'm using fragments? I know fragments cause lifecycle issues, but our usage of them seems pretty straightforward and idiomatic... Are they really this broken by design?

So, the way we're using fragments isn't wrong, it's just several years out of date and reflects the age of the app. The issues we're seeing are very common lifecycle issues that crop up when using a fragment based application. They bite new constantly and were a constant source of issues for developers maybe 3-4 years ago.

Android's Architecture components were [released in 2017](https://android-developers.googleblog.com/2017/11/announcing-architecture-components-10.html) specifically to address these kinds of issues. This isn't the only way to solve the problem of course, but it's the "officially blessed" way to do so from the Android core team.

What I think our goal should be in the short to medium term is migrate all the callback based code out of our view layer (Fragments) up into view models. Then, our fragments can subscribe to `LiveData` observers published by the view model. These are lifecycle aware and that will fix a lot of these issues.

We don't have to do away with our use of Futures or or anything. Just move that up out of any code that deals with the Android lifecycle.

io7m:
I think it's more accurate to say it reflects my out-of-date knowledge. This UI code is only six months old... :sweat:

twaddington:
> I think it's more accurate to say it reflects my out-of-date knowledge. This UI code is only six months old... 😓

Once things settle down I'll take a stab at refactoring one or two of the fragments to use LiveData observers and you can see what you think of the pattern.

PR #776, opened 2020-08-11
SIMPLY-2847: Fix overlay issue with toolbar and bookmarks fragment.

twaddington:
https://jira.nypl.org/browse/SIMPLY-2847

n/a

n/a

n/a

@twaddington ran the vanilla app and tested the R2 reader.


twaddington:
@io7m added a commit. I realized there was a bug that I hadn't addressed with the previous code.

PR #777, opened 2020-08-12
[SIMPLY-2971] Fix a format specifier error.

twaddington:
**What's this do?**
Fix an error handling an audio book server failure.

https://jira.nypl.org/browse/SIMPLY-2971

n/a

n/a

n/a

n/a


twaddington:
> We should really have some automated systems in place. My non-android IDEA setup and at least two Maven plugins I use would have caught that one and not allowed the build to proceed.

Yeah, I have no idea why Android Studio did't flag it as an error. I know I've seen errors about incorrect number of format specifier arguments in the past.

PR #778, opened 2020-08-12
Add new simplified-crashlytics-api module with service type and logging appender.

twaddington:
**What's this do?**
Add the necessary interfaces to plug Crashlytics into our Logback configuration.

We'd like to be able to browse non-fatal exceptions in the Firebase Crashlytics web console.

n/a

We'll need to version bump core and also merge an implementation of the service type into SimplyE.

n/a

@twaddington ran both the vanilla and simplye apps.


io7m:
Looks good to me! I assume the actual service implementation isn't here yet?

twaddington:
> I assume the actual service implementation isn't here yet? &mdash; @io7m 

Implementation is up for review here!

https://github.com/NYPL-Simplified/Simplified-Android-SimplyE/pull/18

PR #779, opened 2020-08-12
Rejigger the vanilla build script to explicitly compile x86 architectures for debug builds only.

twaddington:


simplified-app-vanilla/build.gradle
```diff
@@ -53,16 +66,6 @@ android {
       applicationIdSuffix ".with_profiles"
     }
   }
-
-  packagingOptions {
-    exclude "META-INF/LICENSE"
```
twaddington, line 58: This is excluded by default.

simplified-app-vanilla/build.gradle
```diff
@@ -24,6 +24,17 @@ android {
     setProperty("archivesBaseName", "${versionName}-${versionCode}")
   }
 
+  packagingOptions {
+    doNotStrip 'lib/**/*.so'
```
twaddington, line 28: This was added to match the SimplyE build script.

io7m:
I'm unclear what the consequences of this will be. We don't have R1 `x86_64` binaries...

twaddington:
> I'm unclear what the consequences of this will be. 

This shouldn't actually change much of anything 😂 

> We don't have R1 `x86_64` binaries...

You're right, I guess I've never used an `x86_64` emulator image. I removed that.

io7m:
This would work for vanilla builds that won't ever reach the store, but this won't work for SimplyE builds for one reason: Google won't let you publish to the store if you include `x86` binaries but don't include corresponding `x86_64` binaries. I understand what their aim is, but as usual they screwed up the implementation; what if my library _doesn't need_ native binaries for a 64-bit variant of a 32-bit architecture? I've run into this countless times.

twaddington:
> This would work for vanilla builds that won't ever reach the store, but this won't work for SimplyE builds for one reason: Google won't let you publish to the store if you include `x86` binaries but don't include corresponding `x86_64` binaries. I understand what their aim is, but as usual they screwed up the implementation; what if my library _doesn't need_ native binaries for a 64-bit variant of a 32-bit architecture? I've run into this countless times.

twaddington:
oops, hit the wrong button.

twaddington:
> I'm unclear what the consequences of this will be. &mdash; @io7m 

So, right now the vanilla app is building all available architectures by default. This is a bit slow and bloats the APK that is generated. I did two builds, one from `develop` and one from this branch `build-abi` and here's a screenshot comparing the two in the APK analyzer.

[Screen Shot 2020-08-13 at 8.44.02 AM.png](https://user-images.githubusercontent.com/107117/90156408-3fba2d00-dd41-11ea-9a52-3472a55f1de8.png)


The big difference is that the second one, built with an explicit list of ABIs, is 10MB smaller and only includes the architectures we want.

> Google won't let you publish to the store if you include x86 binaries but don't include corresponding x86_64 binaries. I understand what their aim is, but as usual they screwed up the implementation; what if my library doesn't need native binaries for a 64-bit variant of a 32-bit architecture? I've run into this countless times.

That's totally fine because the way this is configured we only build `x86` for debug builds, which are never uploaded to the Play Store. You can see for release builds we still just compile for ARM.

https://github.com/NYPL-Simplified/Simplified-Android-Core/pull/779/files#diff-e1b6dc8da28300b1edb66fa8e7183efcR52

twaddington:
More info about fat binaries and ABI splitting:

- https://developer.android.com/ndk/guides/abis#gc

io7m:
Those poor forgotten MIPS users!  They will never be able to run SimplyE on these:

SGI_Origin3800_300

Let's make these changes.

twaddington:
Thanks Mark! This will make development a smidge faster.

PR #782, opened 2020-08-13
SIMPLY-2981: Stop capitalizing

io7m:
**What's this do?**
This stops the automatic capitalization of the username field in the accounts fragments.
It also fixes a couple of crashes I discovered on the way to fixing this.

https://jira.nypl.org/browse/SIMPLY-2981

Open the account screen for an NYPL account, or any account that uses a barcode/pin combination. Try typing anything into the barcode field and check that Android doesn't automatically capitalize the first letter.

None.

N/A

@io7m did the above.

### DIFF
--- a/simplified-ui-accounts/src/main/res/layout/auth_basic.xml
+++ b/simplified-ui-accounts/src/main/res/layout/auth_basic.xml
@@ -55,7 +55,7 @@
     android:layout_height="wrap_content"
     android:layout_marginEnd="16dp"
     android:importantForAutofill="no"
-    android:inputType="textPersonName"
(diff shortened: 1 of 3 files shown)


simplified-ui-accounts/src/main/res/layout/auth_basic.xml
```diff
@@ -55,7 +55,7 @@
     android:layout_height="wrap_content"
     android:layout_marginEnd="16dp"
     android:importantForAutofill="no"
-    android:inputType="textPersonName"
```
twaddington, line 58: 👍

PR #787, opened 2020-08-17
[SIMPLY-2970]: Trigger BookSyncTask before ProfileFeedTask when reloading a local feed.

twaddington:
**What's this do?**
This patch updates the local 'Books' and 'Holds' feeds to actually kick off the book sync task when reloading instead of just reparsing the cached local feed.

https://jira.nypl.org/browse/SIMPLY-2970

n/a

n/a

n/a

### DIFF
--- a/simplified-accounts-api/src/main/java/org/nypl/simplified/accounts/api/AccountReadableType.kt
+++ b/simplified-accounts-api/src/main/java/org/nypl/simplified/accounts/api/AccountReadableType.kt
@@ -53,6 +53,6 @@ interface AccountReadableType {
       is AccountProviderAuthenticationDescription.COPPAAgeGate -> false
       is AccountProviderAuthenticationDescription.Basic -> true
       is AccountProviderAuthenticationDescription.OAuthWithIntermediary -> true
-      AccountProviderAuthenticationDescription.Anonymous -> false
+      is AccountProviderAuthenticationDescription.Anonymous -> false
     }
 }
--- a/simplified-futures/src/main/java/org/nypl/simplified/futures/FluentFutureExtensions.kt
+++ b/simplified-futures/src/main/java/org/nypl/simplified/futures/FluentFutureExtensions.kt
@@ -63,7 +63,7 @@ object FluentFutureExtensions {
    */
 
   fun <A> fluentFutureOfAll(futures: List<FluentFuture<A>>) =
-    FluentFuture.from(Futures.allAsList(futures))
+    FluentFuture.from(Futures.successfulAsList(futures))
 
   /**
    * A future that completes immediately with the given value.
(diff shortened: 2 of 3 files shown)


io7m:
I'm in two minds about this. I think what you have is 95% the right behaviour, but I think probably not this:

```
    return this.createNewStatus(
      arguments = arguments,
      future = future
      future = syncFuture.flatMap { future }
    )
```

Futures act like error monads in that, given `x.flatMap(y)`, if `x` fails then `y` is not executed. This would then mean that users couldn't view their books without a network connection. I think you want to just ignore the result of `syncFuture` and unconditionally execute `future` when `syncFuture` completes for any reason.

io7m:
In fact... I think it would likely be better if you just ran the sync tasks synchronously. In other words, don't call the controller API from inside the task, just instantiate and execute the task classes yourself. Easier to unit test, and easier to understand.

io7m:
I feel deleted.

twaddington:
Edit: Somehow I managed to edit Mark's original comment instead of reply to it.

> I think you want to just ignore the result of `syncFuture` and unconditionally execute `future` when `syncFuture` completes for any reason. &mdash; @io7m 

Yep! I *think* that's what I have here or, at least, that's what I intended. The `syncFuture` is generated from a call to `fluentFutureOfAll`, which looks like this:

https://github.com/NYPL-Simplified/Simplified-Android-Core/blob/21a2f2fe4b3e924e6806b4964ac896b1a56b4fa0/simplified-futures/src/main/java/org/nypl/simplified/futures/FluentFutureExtensions.kt#L65-L66

The Guava documentation for `successfulAsList` states:

> Creates a new ListenableFuture whose value is a list containing the values of all its successful input futures. The list of results is in the same order as the input list, and if any of the provided futures fails or is canceled, its corresponding position will contain null (which is indistinguishable from the future having a successful value of null).

I just tested this by turning on airplane mode and refreshing the books fragment. After waiting for the network read to timeout I see the `BookSyncTask` fails with an exception but then the `ProfileFeedTask` executes as expected.

io7m:
Ah! I see. I think this is another reason `FluentFuture` is just... horrible. I'm fairly sure it breaks at least one of the monad laws, but I'd need to prove it to be sure.

The API we have composes fairly well at the application level (which almost never actually happens; almost nothing calls the controller and then `flatMap`-combines the call with another call), but it doesn't compose well inside the actual controller implementation. If I was writing this code today, I'd write a purely synchronous API inside the controller, and expose asynchronous entry points to the app.

If it works, let's go with it, but... be warned! I think this code is due to be assassinated fairly soon.

io7m:
We need an "approve nervously" button.

io7m (commented):
I am skeptical.

twaddington:
> If it works, let's go with it, but... be warned! I think this code is due to be assassinated fairly soon. &mdash; @io7m 

👍

PR #788, opened 2020-08-17
Use the private, internal, cache directory for writing logs instead of the external cache dir (world-readable).

twaddington:
**What's this do?**
This patch updates all references of `externalCacheDir` to `cacheDir` (internal).

In practice, `externalCacheDir` and `cacheDir` may return the same location.

> If a shared storage device is emulated (as determined by `Environment#isExternalStorageEmulated(File))`, its contents are backed by a private user data partition, which means there is little benefit to storing data here instead of the private directory returned by `getCacheDir()`.
>
> &mdash; https://developer.android.com/reference/kotlin/android/content/Context#getExternalCacheDir()

However, in some cases `externalCacheDir` may be a world-readable location.

> There is no security enforced with these files. For example, any application holding `android.Manifest.permission#WRITE_EXTERNAL_STORAGE` can write to these files.
> 
> &mdash; https://developer.android.com/reference/kotlin/android/content/Context#getExternalCacheDir()

While we shouldn't be logging any sensitive information to disk anyway, it would be prudent to default to the more secure internal cache directory.

n/a

n/a

n/a

### DIFF
--- a/simplified-main/src/main/java/org/nypl/simplified/main/MainApplication.kt
+++ b/simplified-main/src/main/java/org/nypl/simplified/main/MainApplication.kt
@@ -60,8 +60,8 @@ class MainApplication : Application() {
         while (index.hasNext()) {
           val appender = index.next()
           if (appender is RollingFileAppender<*>) {
-            externalCacheDir?.mkdirs()
-            val path = File(externalCacheDir, "log.txt").absolutePath
+            cacheDir?.mkdirs()
+            val path = File(cacheDir, "log.txt").absolutePath
             (appender as RollingFileAppender<*>).file = path
             appender.start()
           }
--- a/simplified-main/src/main/res/xml/provider_paths.xml
+++ b/simplified-main/src/main/res/xml/provider_paths.xml
@@ -1,9 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <paths>
-  <external-cache-path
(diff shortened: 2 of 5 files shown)


twaddington:
I need to verify this doesn't break the attaching of logs to problem report emails. We may need to grant additional permissions in `Reports.kt` when calling `getUriForFile`.

twaddington:
> I need to verify this doesn't break the attaching of logs to problem report emails. We may need to grant additional permissions in `Reports.kt` when calling `getUriForFile`.

Looks like we already set `FLAG_GRANT_READ_URI_PERMISSION` on the Intent, so no additional permissions are required.

io7m:
This looks fine to me, as long as you're sure we can still get error reports.

In LFA's updater, I think I use the internal cache directory for logs, but I expose everything through a content provider.

twaddington:
> This looks fine to me, as long as you're sure we can still get error reports.
> 
> In LFA's updater, I think I use the internal cache directory for logs, but I expose everything through a content provider.

I double checked that this works as expected and it looks like I need to update the `provider_paths.xml` file.

twaddington:
@io7m I rebased this pull-request on develop and made the necessary updates to `provider_paths.xml`. I tested this by triggering a crash on startup (from the debug options) and sending myself an app report. Works great 👍

PR #789, opened 2020-08-18
When the patron enables debug options set a flag in their profile to persist the preference.

twaddington:


simplified-ui-settings/src/main/res/values/strings.xml
```diff
@@ -54,10 +54,11 @@
   <string name="settingsPlaceholder">Lorem ipsum dolor sit amet, probatus volutpat has at, vis adhuc iuvaret omittantur an, sadipscing conclusionemque his ad. Eam ut simul tempor. Primis consetetur appellantur vim eu, eum an iudico dolorum. Ad vis utamur honestatis, sanctus debitis referrentur an eam.</string>
   <string name="settingsReportIssue">Report an issue…</string>
   <string name="settingsSyncBookmarks">Sync bookmarks</string>
+  <string name="settingsTapToDebug">Debug mode in %1$d&#8230;</string>
   <string name="settingsUserName">User Name</string>
   <string name="settingsVersion">App version</string>
   <string name="settingsVersionBuild">Build</string>
-  <string name="settingsVersionDeveloper">Developer Options</string>
+  <string name="settingsVersionDeveloper">Debug Options</string>
```
twaddington, line 61: Not everyone who uses these options is a developer.

io7m:
Ktlint issues! :yum:

Otherwise, I like it. Been meaning to do something like this for a while.

PR #790, opened 2020-08-18
SIMPLY-2964: Juvenile card creator shows up after login for Brooklyn Public Library

MalcolmMcFly:
**What's this do?**
Adds missing logic to enable/disable buttons and card creator based on account

[SIMPLY-2964: Juvenile card creator shows up after login for Brooklyn Public Library](https://jira.nypl.org/browse/SIMPLY-2964)

Login to an NYPL account and any other library account and ensure that the juvenile card creator is not launched for non-NYPL acccount

n/a

n/a

@MalcolmMcFly

### DIFF
--- a/simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/AccountFragment.kt
+++ b/simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/AccountFragment.kt
@@ -922,6 +922,7 @@ class AccountFragment : Fragment() {
       AsLoginButtonDisabled -> {
         this.loginButton.setText(R.string.accountLogin)
         this.signUpLabel.setText(R.string.accountCardCreatorLabel)
+        this.signUpLabel.isEnabled = true
         this.loginButton.isEnabled = false
       }
       is AsCancelButtonEnabled -> {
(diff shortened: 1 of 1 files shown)


simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/AccountFragment.kt
```diff
@@ -1020,6 +1036,7 @@ class AccountFragment : Fragment() {
     val loginSatisfied = this.determineLoginIsSatisfied()
     this.setLoginButtonStatus(loginSatisfied)
     this.authenticationAlternativesShow()
+    this.signUpButton.isEnabled = true
```
twaddington, line 1039: Do we need a corresponding `this.signUpLabel.isEnabled = true` here?

twaddington:
Looks good! Had one small question.

io7m:
Looks good to me. I don't like seeing "If NYPL then do this..." in the core, but sometimes it's just unavoidable.

killerpuppytails:
> Looks good to me. I don't like seeing "If NYPL then do this..." in the core, but sometimes it's just unavoidable.

Once we have the web-based card creation system up, we can use that and everything will be consistent :)

MalcolmMcFly:
> Looks good to me. I don't like seeing "If NYPL then do this..." in the core, but sometimes it's just unavoidable.

Yea that's what I was thinking. We basically handle every other library the same outside of NYPL that has the juvenile card creator capability.

PR #791, opened 2020-08-19
[SIMPLY-2823] Install a 10MB HTTP response cache.

twaddington:
https://jira.nypl.org/browse/SIMPLY-2823

This will require some longer term testing and dogfooding before I'll feel comfortable shipping it. It's a little magic and I'm wary it could break things in unexpected ways.

We could nest it under a debug flag or just enable it for developer builds for now. Thoughts?

n/a

n/a

### DIFF
--- a/simplified-main/src/main/java/org/nypl/simplified/main/MainApplication.kt
+++ b/simplified-main/src/main/java/org/nypl/simplified/main/MainApplication.kt
@@ -1,6 +1,7 @@
 package org.nypl.simplified.main
 
 import android.app.Application
+import android.net.http.HttpResponseCache
 import android.os.StrictMode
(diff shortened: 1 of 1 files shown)


io7m:
Looks OK. I'm not too enthusiastic about adding more caching in general though.

We could definitely use a debug flag to turn this on/off and possibly a control to clear the cache if possible.

twaddington:
I wrapped it with a debug flag for now, so this will only affect vanilla builds.

io7m:
We need an "abstain" button for GitHub reviews.

twaddington:
> I'm not too enthusiastic about adding more caching in general though. &mdash; @io7m 

I feel the same way but I suppose it's worth trying.

PR #793, opened 2020-08-20
Configure and initialize Logback in MainApplication and remove logback.xml configuration.

twaddington:
**What's this do?**
Configure and initialize Logback in MainApplication and remove logback.xml configuration.

The logback.xml configuration did not reflect the actual
logging configuration due to Android's dynamic filesystem.

This change centralizes logging configuration in one place and makes
future refactoring simpler.

n/a

n/a

n/a

@twaddington ran the Vanilla app and made sure the logging appenders started up as expected.


twaddington:
Very curious what your feelings are about this @io7m.

io7m:
Looks fine to me! I'm used to using `logback.xml` in plain Java applications, but in that world people are able to supply their own config files on the application command line (via a `ch.qos.logback.configurationFile=...` property). In Android, they obviously can't. 

This also has the benefit that programmatic configuration makes the dynamic elements more explicit. I've never been particularly happy about relying on weird environment variables in the config file.

io7m (approved):

twaddington:
> Looks fine to me! I'm used to using `logback.xml` in plain Java applications, but in that world people are able to supply their own config files on the application command line (via a `ch.qos.logback.configurationFile=...` property). In Android, they obviously can't.
> 
> This also has the benefit that programmatic configuration makes the dynamic elements more explicit. I've never been particularly happy about relying on weird environment variables in the config file. &mdash; @io7m 

Woo! I was expecting more attachment to the xml config to be honest 😅

PR #795, opened 2020-08-27
SIMPLY-2400: Wait for book content to load before applying reader preferences.

ray-lee:
**What's this do?**

Wait until Readium indicates that book content has loaded before attempting to apply the user's reader preferences (font and color scheme). Currently, the app attempts to apply the preferences at an arbitrary 300 ms delay after Readium has initialized.

This fixes [SIMPLY-2400](https://jira.nypl.org/browse/SIMPLY-2400).

Readium loads book content into an iframe in its WebView. While the content is loading, the iframe is [hidden with css `opacity: 0.01`](https://github.com/readium/readium-shared-js/blob/master/js/views/reflowable_view.js#L260). When the content is complete (the `load` event of the iframe is fired), Readium [unhides the iframe](https://github.com/readium/readium-shared-js/blob/7f245beba1ed97eaabce0aa5e9cf2f3b23e8f8f6/js/views/reflowable_view.js#L376). Meanwhile, SimplyE is currently [waiting 300 ms](https://github.com/NYPL-Simplified/Simplified-Android-Core/blob/fab8a31da8d19776518c39a0594a39091ee21ce4/simplified-viewer-epub-readium1/src/main/java/org/nypl/simplified/viewer/epub/readium1/ReaderActivity.java#L515-L532) before [calling Readium's `setBookStyles`](https://github.com/NYPL-Simplified/Simplified-Android-Core/blob/fab8a31da8d19776518c39a0594a39091ee21ce4/simplified-viewer-epub-readium1/src/main/java/org/nypl/simplified/viewer/epub/readium1/ReaderReadiumJavaScriptAPI.java#L197) function to set the book styles. That function attempts to modify the document in the iframe by [injecting a `style` element](https://github.com/readium/readium-shared-js/blob/7f245beba1ed97eaabce0aa5e9cf2f3b23e8f8f6/js/helpers.js#L825-L829). On slower devices and/or with larger content, the iframe may not have finished loading in that 300 ms window. The resulting modification of the iframe's document while it is still loading can interfere with the `load` event being fired, so the iframe is never unhidden, and the content fails to appear.

This changes the app to wait for Readium to dispatch its `CONTENT_DOCUMENT_LOADED` event before applying book styles. [`CONTENT_DOCUMENT_LOADED` is dispatched](https://github.com/readium/readium-shared-js/blob/7f245beba1ed97eaabce0aa5e9cf2f3b23e8f8f6/js/views/reflowable_view.js#L303) after the iframe has fired its `load` event.

A slower device is necessary to reproduce the problem. I've been able to do it on the Android emulator on a MacBook, as follows:

- Run the vanilla app
- In the SimplyE collection, open "Rosehead" by Ksenia Anske to the cover
- In Chrome, open chrome://inspect and click the Inspect link on the WebView
- In the inspector, click the reload button (or command/ctrl + R) repeatedly until the cover fails to render (usually within 10 tries)

After applying the fix, the book should appear every time, with the user's last selected font/color scheme. Changing the font/color scheme should continue to work.

n/a

n/a

@ray-lee ran the vanilla app.

### DIFF
--- a/simplified-viewer-epub-readium1/src/main/assets/host_app_feedback.js
+++ b/simplified-viewer-epub-readium1/src/main/assets/host_app_feedback.js
@@ -49,6 +49,7 @@ $(document).ready(function()
             ReadiumSDK.reader.on(ReadiumSDK.Events.MEDIA_OVERLAY_TTS_STOP, this.onMediaOverlayTTSStop, this);
             ReadiumSDK.reader.on(ReadiumSDK.Events.PAGINATION_CHANGED, this.onPaginationChanged, this);
             ReadiumSDK.reader.on(ReadiumSDK.Events.SETTINGS_APPLIED, this.onSettingsApplied, this);
+            ReadiumSDK.reader.on(ReadiumSDK.Events.CONTENT_DOCUMENT_LOADED, this.onContentDocumentLoaded, this);
 
             window.location.href = "readium:initialize";
           },
@@ -81,6 +82,10 @@ $(document).ready(function()
         this.onMediaOverlayTTSStop = function()
         {
           window.location.href = "readium:media-overlay-tts-stop";
+        };
+
+        this.onContentDocumentLoaded = function() {
+          window.location.href = "readium:content-document-loaded";
         };
       }();
 
--- a/simplified-viewer-epub-readium1/src/main/java/org/nypl/simplified/viewer/epub/readium1/ReaderActivity.java
+++ b/simplified-viewer-epub-readium1/src/main/java/org/nypl/simplified/viewer/epub/readium1/ReaderActivity.java
@@ -510,9 +510,7 @@ public final class ReaderActivity extends AppCompatActivity implements
   }
 
   private void applyReaderPreferences(final ReaderPreferences preferences) {
-    LOG.debug("applyReaderPreferences: scheduling task to run later");
-
-    uiThread.runOnUIThreadDelayed(() -> {
+    uiThread.runOnUIThread(() -> {
       LOG.debug("applyReaderPreferences: executing now");
 
       // Get the CFI from the ReadiumSDK before applying the new
@@ -529,7 +527,7 @@ public final class ReaderActivity extends AppCompatActivity implements
 
       this.readium_js_api.getCurrentPage(this);
       this.readium_js_api.mediaOverlayIsAvailable(this);
-    }, 300L);
+    });
   }
 
   @Override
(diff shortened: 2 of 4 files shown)


simplified-viewer-epub-readium1/src/main/assets/host_app_feedback.js
```diff
@@ -82,6 +83,10 @@ $(document).ready(function()
         {
           window.location.href = "readium:media-overlay-tts-stop";
         };
+
+        this.onContentDocumentLoaded = function() {
+          window.location.href = "readium:content-document-loaded";
```
ray-lee, line 88: Following the existing pattern: JavaScript events from Readium are bridged to Java by using the WebView's location as a message bus.

simplified-viewer-epub-readium1/src/main/java/org/nypl/simplified/viewer/epub/readium1/ReaderActivity.java
```diff
@@ -510,9 +510,7 @@ private void onProfileEventPreferencesChanged(
   }
 
   private void applyReaderPreferences(final ReaderPreferences preferences) {
-    LOG.debug("applyReaderPreferences: scheduling task to run later");
-
-    uiThread.runOnUIThreadDelayed(() -> {
+    uiThread.runOnUIThread(() -> {
```
ray-lee, line 513: I'm removing the execution delay here. This function is now called from `onReadiumContentDocumentLoaded`, so a delay is no longer necessary to wait for the content to load. It also continues to be called from `onProfileEventPreferencesChanged`, which happens when the user selects a new font/color scheme. I believe the delay was not needed in this case, and was only making the app less responsive.

simplified-viewer-epub-readium1/src/main/java/org/nypl/simplified/viewer/epub/readium1/ReaderActivity.java
```diff
@@ -804,17 +802,6 @@ public void onReadiumFunctionInitialize() {
...
-      this.applyReaderPreferences(
-        Services.INSTANCE.serviceDirectory()
-          .requireService(ProfilesControllerType.class)
-          .profileCurrent()
-          .preferences()
-          .getReaderPreferences());
-    } catch (final ProfileNoneCurrentException e) {
-      throw new IllegalStateException(e);
-    }
-
```
ray-lee, line 817: This is now called from `onReadiumContentDocumentLoaded`.

io7m:
Amazing work, Ray. I think you're about the fifth person to try to track down this problem and everyone else so far has failed. If this works, LFA in particular are going to be overjoyed (because they would now have a choice between R1 and R2!).

The changes look good to me. That delayed execution has been in there since day 1, I think. Impressive that it's triggered these kinds of problems six years later!

io7m (approved):
This is all good as far as I'm concerned.

Aside from anything; we can hardly make "most people can't read books" worse.

PR #796, opened 2020-08-27
Migrate to Readium2 alpha branch and prepare DRM support

qnga:
**What's this do?**
Use the Readium 2 alpha API with DRM support.

R2 opens publications with a `Streamer` that references `ContentProtection`s which depend on services from `simplified-main`.
Therefore, this streamer must be passed to the `SR2Controller` which is responsible for opening publications.
A special `DRMProtectedFile` must be used to pass along adobe rights path to the `AdobeAdeptContentProtection`.

This may be tested by checking ebooks can still be read.

Add dependencies to some R2 modules.

No.

Me.

See also
https://github.com/NYPL-Simplified/Simplified-R2-Android/pull/7
https://github.com/NYPL-Simplified/DRM-Android-Adobe/pull/8
https://github.com/NYPL-Simplified/DRM-Android-Core/pull/4

### DIFF
--- a/simplified-viewer-epub-readium2/build.gradle
+++ b/simplified-viewer-epub-readium2/build.gradle
@@ -13,4 +13,6 @@ dependencies {
   implementation libraries.nyplR2API
   implementation libraries.nyplR2Vanilla
   implementation libraries.nyplR2Views
+  implementation libraries.r2Shared
+  implementation libraries.r2Streamer
 }
--- a/simplified-viewer-epub-readium2/build.gradle
+++ b/simplified-viewer-epub-readium2/build.gradle
@@ -10,6 +10,7 @@ dependencies {
   implementation libraries.androidXLifecycle
   implementation libraries.androidXLifecycleViewmodel
   implementation libraries.kotlinStdlib
+  implementation libraries.nyplDRMCore
   implementation libraries.nyplR2API
   implementation libraries.nyplR2Vanilla
   implementation libraries.nyplR2Views
--- a/simplified-viewer-epub-readium2/src/main/java/org/nypl/simplified/viewer/epub/readium2/ReaderActivity.kt
+++ b/simplified-viewer-epub-readium2/src/main/java/org/nypl/simplified/viewer/epub/readium2/ReaderActivity.kt
@@ -47,6 +47,9 @@ import org.nypl.simplified.profiles.controller.api.ProfilesControllerType
 import org.nypl.simplified.reader.bookmarks.api.ReaderBookmarkServiceType
 import org.nypl.simplified.reader.bookmarks.api.ReaderBookmarks
 import org.nypl.simplified.ui.thread.api.UIThreadServiceType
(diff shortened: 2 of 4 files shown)


simplified-viewer-epub-readium2/src/main/java/org/nypl/simplified/viewer/epub/readium2/ReaderActivity.kt
```diff
@@ -47,8 +49,12 @@ import org.nypl.simplified.profiles.controller.api.ProfilesControllerType
 import org.nypl.simplified.reader.bookmarks.api.ReaderBookmarkServiceType
 import org.nypl.simplified.reader.bookmarks.api.ReaderBookmarks
 import org.nypl.simplified.ui.thread.api.UIThreadServiceType
+import org.readium.r2.shared.publication.ContentProtection
```
MalcolmMcFly, line 52: This import still being used?

MalcolmMcFly (changes requested):

simplified-viewer-epub-readium2/src/main/java/org/nypl/simplified/viewer/epub/readium2/ReaderActivity.kt
qnga: No, thanks.

simplified-viewer-epub-readium2/build.gradle
```diff
@@ -10,7 +10,10 @@ dependencies {
   implementation libraries.androidXLifecycle
   implementation libraries.androidXLifecycleViewmodel
   implementation libraries.kotlinStdlib
+  implementation libraries.nyplDRMCore
   implementation libraries.nyplR2API
   implementation libraries.nyplR2Vanilla
   implementation libraries.nyplR2Views
+  implementation libraries.r2Shared
```
twaddington, line 17: These two dependencies, `r2Shared` and `r2Streamer`, are provided as transitive dependencies from the `org.librarysimplified.r2.api` module in "Simplified-R2-Android". As a result, I think these are redundant declarations and can be removed.

Shouldn't hurt anything if left here, though.

twaddington:
This looks solid enough. I think we may need to integrate this into `develop` instead of `feature-readium2` but we can do that after it's merged. I'm good with this.

@io7m?

twaddington (approved):

twaddington:
Just a note to follow-up on after this is merged: We'll need to publish a new version of `Simplified-R2-Android` and then integrate this into `develop`. Also, double-check lines 157 and 158 of build.gradle as I think the dep names are swapped for `nyplR2Vanilla` and `nyplR2API`.

simplified-viewer-epub-readium2/build.gradle
qnga: I explicitly added them because some symbols from these modules are directly used in `simplified-viewer-epub-readium2`.
Feel free to remove them if you prefer so.

PR #798, opened 2020-09-02
[SIMPLY-3035] Fix issue with scrollbars briefly appearing on dynamically created buttons.

twaddington:
**What's this do?**
Fixes a visual issue with dynamically created buttons.

https://jira.nypl.org/browse/SIMPLY-3035

You can verify this fix by toggling between the Books and Settings tabs.

n/a

n/a

### DIFF
--- a/simplified-ui-theme/src/main/res/values/styles.xml
+++ b/simplified-ui-theme/src/main/res/values/styles.xml
@@ -108,6 +108,12 @@
     <item name="android:background">@drawable/simplified_button</item>
     <item name="android:textColor">@drawable/simplified_button_text</item>
     <item name="android:textSize">12sp</item>
+
+    <!--
+      The scroll indicator will briefly appear on programmatically created buttons unless the
+      'scrollbars' attribute is set to 'none'.
+    -->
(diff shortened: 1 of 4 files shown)


simplified-ui-theme/src/main/res/values/styles.xml
```diff
@@ -108,6 +108,12 @@
     <item name="android:background">@drawable/simplified_button</item>
     <item name="android:textColor">@drawable/simplified_button_text</item>
     <item name="android:textSize">12sp</item>
+
+    <!--
+      The scroll indicator will briefly appear on programmatically created buttons unless the
+      'scrollbars' attribute is set to 'none'.
+    -->
+    <item name="android:scrollbars">none</item>
```
twaddington, line 116: Turns out when we add buttons dynamically Android somehow thinks they're scrollable. I have no idea why this is happening. I did some searching and read through some of the `View` source code but I couldn't figure out exactly what was happening. However, adding this attribute resolves the issue and I can't imagine why a button would need to be internally scrollable anyway.

io7m (approved):
Looks good!

PR #799, opened 2020-09-02
Show the compiled version of Simplified Core in settings.

twaddington:
**What's this do?**
Makes it easier for testers to discover what version of Simplified Core was used to build their app.

Had a discussion about this with Mark and Jonathan Green in Slack.

Open settings and verify the right version is shown.

Will need to make a small update to the SimplyE build configuration as the interface was updated with a new field.

n/a


io7m:
device-2020-09-02-092013

Nice!

io7m:
You're missing a small change in the test suite. That's what caused the build failure.

twaddington:
> You're missing a small change in the test suite. That's what caused the build failure.

oops, I missed that usage because I had that sandbox module unloaded in android studio.

io7m:
Everything in the sandbox module is open for deletion at any time.

twaddington:
@io7m fixed!

PR #800, opened 2020-09-02
[SIMPLY-2858] Fix ordering of testing libraries, when enabled.

twaddington:
**What's this do?**
The list of testing libraries already includes the production libraries, so only load one or the other. This means the testing libraries are no longer just appended to the end of the list (making them difficult to find).

https://jira.nypl.org/browse/SIMPLY-2858

Make sure enabling testing libraries in debug settings works as expected.

n/a

n/a

### DIFF
--- a/simplified-accounts-registry/src/main/java/org/nypl/simplified/accounts/registry/AccountProviderRegistry.kt
+++ b/simplified-accounts-registry/src/main/java/org/nypl/simplified/accounts/registry/AccountProviderRegistry.kt
@@ -102,6 +102,8 @@ class AccountProviderRegistry private constructor(
   }
 
   override fun clear() {
+    this.descriptions.clear()
+    this.resolved.clear()
     for (source in this.sources) {
       source.clear(this.context)
     }
--- a/simplified-accounts-source-nyplregistry/src/main/java/org/nypl/simplified/accounts/source/nyplregistry/AccountProviderSourceNYPLRegistry.kt
+++ b/simplified-accounts-source-nyplregistry/src/main/java/org/nypl/simplified/accounts/source/nyplregistry/AccountProviderSourceNYPLRegistry.kt
@@ -34,6 +34,7 @@ import org.nypl.simplified.parser.api.ParseResult
 import org.nypl.simplified.taskrecorder.api.TaskResult
 import org.slf4j.LoggerFactory
 import java.io.File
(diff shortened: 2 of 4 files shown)


simplified-accounts-registry/src/main/java/org/nypl/simplified/accounts/registry/AccountProviderRegistry.kt
```diff
@@ -102,6 +102,8 @@ class AccountProviderRegistry private constructor(
   }
 
   override fun clear() {
+    this.descriptions.clear()
```
twaddington, line 105: We only call clear in one place, so this shouldn't negatively affect anything.

simplified-accounts-source-nyplregistry/src/main/java/org/nypl/simplified/accounts/source/nyplregistry/AccountProviderSourceNYPLRegistry.kt
```diff
@@ -244,18 +245,17 @@ class AccountProviderSourceNYPLRegistry(
...
-            this.logger.debug("{} cached production providers", countProduction)
-            this.logger.debug("{} cached testing providers", countTesting)
-            result.result.providers.associateBy(AccountProviderDescription::id).toMap()
+            this.logger.debug("{} cached providers", result.result.providers.size)
+            result.result.providers.associateBy(AccountProviderDescription::id)
           }
         }
       }
+    } catch (e: FileNotFoundException) {
+      this.logger.debug("no cache file exists, skipping")
```
twaddington, line 254: Avoid spamming logcat with an exception for an expected error.

simplified-accounts-registry/src/main/java/org/nypl/simplified/accounts/registry/AccountProviderRegistry.kt
```diff
@@ -102,6 +102,8 @@ class AccountProviderRegistry private constructor(
   }
 
   override fun clear() {
```
twaddington, line 104: We only call clear in one place, so this change shouldn't affect anything unexpected.

io7m (approved):
I like it!

twaddington:
I think the build got stuck so I'm gonna merge.

io7m:
It got there eventually.

PR #801, opened 2020-09-07
SIMPLY-3001: Part 1

io7m:
**What's this do?**
This implements a set of changes for the first part of the borrow task refactoring:

This eliminates the "error" type parameter from the task recorder and
refactors all existing code to use the new lesser-typed interface.
It adjusts the task recorder API such that Presentable attributes
are no longer an aspect of individual task steps but of the recorder
itself (and, as a result, of the task result). It also simplifies the
Presentable interfaces and removes functionality that we weren't using.

And then:

Previously the test suite was structured such that each test class
was actually three classes:
    
      1. An abstract "Contract" class in simplified-tests/src/main/java
         containing the actual tests.
      2. A concrete "Test" class in simplified-tests/src/test/java
         that runs the Contract on the local machine without a real
         or emulated Android device.
      3. A concrete "Test" class in simplified-tests-android/src/test/java
         that runs the Contract on a real or emulated Android device.
    
This was necessary because of the brain-dead way that Android decided
to implement testing: In order for a test to run on a device, the test
must use a proprietary instrumentation API. This means that a test
class must _either_ be a local unit test, _or_ a device unit test,
but cannot be both. The Contract/Test split was created to work around
this stupidity.
    
The problem; maintaining device tests is a full-time job. The test
runner for Android is fragile to the point of being essentially
useless. We didn't maintain the tests, and nobody had run them for
quite some time.
    
This change simply removes the tests-android module. The tests are
still split into Contract/Test classes, but these now all live in
simplified-tests/src/main/java and we only support running tests
locally. New unit tests do not need to have a Contract/Test split.

Altogether, these changes make it possible for me to write the
replacement for the borrow task without drowning in legacy
horrors. The app, after applying this patch, still does exactly
what it always did.

https://jira.nypl.org/browse/SIMPLY-3001
https://jira.nypl.org/browse/SIMPLY-2940

Observe that the test suite still runs. No borrowing/login/logout behaviour
has been changed.

None.

Module descriptions haven't been regenerated yet. That's for part 2.

@io7m ran it a lot. Also used the Vanilla app to borrow and read
books.


io7m:
I'm aware that this is a painfully large amount of code. The borrowing task touched everything, and a lot of files were moved in the process of removing the tests-android module.

I don't expect a serious review, as I've not implemented anything "new" but have just improved what was already there. All of the changes I made have been careful to preserve the existing semantics of the code, but I've changed the APIs in a way that will actually allow me to write the replacement code.

io7m:
Additionally... A lot of the changes that would be reviewed are in code that will be _removed_ as Part 2... :sweat_smile:

io7m:
OK to merge this now that `6.2.0` has gone out?

twaddington:
@io7m I'd still like to take a look at this today. Can you wait one more day?

io7m:
Yep! I'm in no rush other than to potentially avoid conflicts with incoming PRs. I've had to fix a couple of conflicts so far.

simplified-accounts-source-nyplregistry/src/main/java/org/nypl/simplified/accounts/source/nyplregistry/AccountProviderResolutionErrorCodes.kt
```diff
@@ -0,0 +1,45 @@
+package org.nypl.simplified.accounts.source.nyplregistry
+
+import org.nypl.simplified.accounts.api.AccountProviderDescription
+import java.net.URI
+
+/**
+ * Error codes raised during account resolution. Applications MUST NOT depend on these
```
twaddington, line 7: I'm a bit confused by this comment. Looks like you'd intended these to be used for unit tests but I see them used in the `AccountProviderResolution.kt` file?

If they should be used only by tests they should probably be in a test module. Otherwise, maybe the comment needs amending?

twaddington (approved):
Looks alright to me 👍

simplified-accounts-source-nyplregistry/src/main/java/org/nypl/simplified/accounts/source/nyplregistry/AccountProviderResolutionErrorCodes.kt
io7m: Yep, you're right, the comment is misleading.

What I meant was that I _really_ don't want to see code like this in the app:

```
if (taskResult.errorCode == "someSpecificFailure") {
  showSomeSpecificFailureView()
}
```

... because code like that should be written over values of sealed types so that we can be certain at compile time that we've handled all of the special cases. Exposing opaque error codes for unit testing is fine, though.

io7m:
_pulls the pin_

PR #803, opened 2020-09-14
Bandaid a few issues from Crashlytics.

twaddington:
**What's this do?**
Attempt to resolve a few issues from Crashlytics.

- https://jira.nypl.org/browse/SIMPLY-3075
- https://jira.nypl.org/browse/SIMPLY-3076
- https://jira.nypl.org/browse/SIMPLY-3077

n/a

n/a

n/a

### DIFF
--- a/simplified-main/src/main/java/org/nypl/simplified/main/MainServices.kt
+++ b/simplified-main/src/main/java/org/nypl/simplified/main/MainServices.kt
@@ -389,7 +389,7 @@ internal object MainServices {
     return try {
       this.createBundledCredentials(context.assets)
     } catch (e: FileNotFoundException) {
-      this.logger.debug("could not initialize bundled credentials: ", e)
+      this.logger.debug("could not initialize bundled credentials; none found")
       AccountBundledCredentialsEmpty.getInstance()
     } catch (e: IOException) {
       this.logger.debug("could not initialize bundled credentials: ", e)
--- a/simplified-main/src/main/java/org/nypl/simplified/main/MainServices.kt
+++ b/simplified-main/src/main/java/org/nypl/simplified/main/MainServices.kt
@@ -552,11 +552,11 @@ internal object MainServices {
         CardCreatorService.create(stream)
       }
     } catch (e: FileNotFoundException) {
-      this.logger.debug("card creator configuration not present: ", e)
+      this.logger.debug("could not initialize card creator; cardcreator.conf not found")
       null
     } catch (e: IOException) {
       this.logger.debug("could not initialize card creator: ", e)
-      throw IllegalStateException("could not initialize card creator", e)
+      null
     }
   }
 
--- a/simplified-viewer-audiobook/src/main/java/org/nypl/simplified/viewer/audiobook/AudioBookPlayerActivity.kt
+++ b/simplified-viewer-audiobook/src/main/java/org/nypl/simplified/viewer/audiobook/AudioBookPlayerActivity.kt
@@ -412,7 +412,7 @@ class AudioBookPlayerActivity : AppCompatActivity(),
         this.supportFragmentManager
           .beginTransaction()
           .replace(R.id.audio_book_player_fragment_holder, this.playerFragment, "PLAYER")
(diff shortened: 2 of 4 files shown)


MalcolmMcFly:
@twaddington can you create a general ticket and just throw some details in for tracking?

twaddington:
Added Jira tickets and rebased.

simplified-main/src/main/java/org/nypl/simplified/main/MainServices.kt
```diff
@@ -389,7 +389,7 @@ internal object MainServices {
     return try {
       this.createBundledCredentials(context.assets)
     } catch (e: FileNotFoundException) {
-      this.logger.debug("could not initialize bundled credentials: ", e)
+      this.logger.debug("could not initialize bundled credentials; none found")
```
io7m, line 392: At some point in the future, I'm going to try to introduce something where you have to state that you meant to include bundled credentials. If you state that you meant to include them and you don't, the app should refuse to start up. If you don't state that you meant to include them, and you actually do include the file, the app should also refuse to start up.

Right now we just print an exception regardless of your actual intention (and now we don't even do that!).

simplified-viewer-audiobook/src/main/java/org/nypl/simplified/viewer/audiobook/AudioBookPlayerActivity.kt
```diff
@@ -148,7 +148,7 @@ class AudioBookPlayerActivity : AppCompatActivity(),
     this.log.debug("entry id:      {}", this.parameters.opdsEntry.id)
 
     this.setTheme(
-      Services.serviceDirectory()
+      Services.serviceDirectoryWaiting(30L, TimeUnit.SECONDS)
```
io7m, line 151: How is it that we're reaching this code?

Is this happening because Android restarts the app with the `AudioBookPlayerActivity` at the top of the stack, and the boot process is therefore happening in the background?

simplified-viewer-audiobook/src/main/java/org/nypl/simplified/viewer/audiobook/AudioBookPlayerActivity.kt
```diff
@@ -412,7 +412,7 @@ class AudioBookPlayerActivity : AppCompatActivity(),
         this.supportFragmentManager
           .beginTransaction()
           .replace(R.id.audio_book_player_fragment_holder, this.playerFragment, "PLAYER")
-          .commit()
+          .commitAllowingStateLoss()
```
io7m, line 415: We didn't need that state anyway. :laughing:

simplified-main/src/main/java/org/nypl/simplified/main/MainServices.kt
twaddington: Yes, that sounds like a nice improvement.

simplified-viewer-audiobook/src/main/java/org/nypl/simplified/viewer/audiobook/AudioBookPlayerActivity.kt
twaddington: I have no idea. The Application instance should be the first thing that starts. If services are initialized asynchronously it's possible there could be a race condition here if `AudioBookPlayerActivity` is at the top of the stack and the app is restarting. Perhaps if a user left the app open or backgrounded and came back a while later. The system could have killed the app process due to memory pressure and is restarting. Worth investigating more but I don't have time right now 😂

simplified-viewer-audiobook/src/main/java/org/nypl/simplified/viewer/audiobook/AudioBookPlayerActivity.kt
twaddington: haha, yeah, I love that method. does the same thing but makes the fragment manager not complain so much. Another one of the dumb things about the Fragment API.

PR #804, opened 2020-09-15
[SIMPLY-2580] Publish the 'BookOpened' event from both R1 and R2 readers.

twaddington:
**What's this do?**
Publish the 'BookOpened' event from both R1 and R2 readers.

https://jira.nypl.org/browse/SIMPLY-2580

n/a

We'll need to add `:simplified-analytics-circulation` as a dependency to SimplyE to properly consume these events.

n/a

### DIFF
--- a/simplified-analytics-circulation/src/main/java/org/nypl/simplified/analytics/circulation/CirculationAnalyticsSystem.kt
+++ b/simplified-analytics-circulation/src/main/java/org/nypl/simplified/analytics/circulation/CirculationAnalyticsSystem.kt
@@ -2,6 +2,7 @@ package org.nypl.simplified.analytics.circulation
 
 import com.io7m.jfunctional.Option
 import org.nypl.simplified.accounts.api.AccountAuthenticatedHTTP
+import org.nypl.simplified.accounts.api.AccountAuthenticationCredentials
 import org.nypl.simplified.analytics.api.AnalyticsConfiguration
 import org.nypl.simplified.analytics.api.AnalyticsEvent
 import org.nypl.simplified.analytics.api.AnalyticsSystem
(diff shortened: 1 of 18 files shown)


simplified-analytics-circulation/src/main/java/org/nypl/simplified/analytics/circulation/CirculationAnalyticsSystem.kt
```diff
@@ -25,47 +26,23 @@ class CirculationAnalyticsSystem(
...
-          this.logger.debug(
-            "no analytics URI available for book {} ({})",
-            event.opdsEntry.id,
-            event.opdsEntry.title)
+        event.targetURI?.let { target ->
+          postURI(target, event.credentials)
         }
       }
+      else -> {
+        // All other events are silently dropped
```
twaddington, line 37: @io7m are all these other events are used by LFA?

simplified-viewer-epub-readium2/src/main/java/org/nypl/simplified/viewer/epub/readium2/ReaderActivity.kt
```diff
@@ -149,6 +164,19 @@ class ReaderActivity : AppCompatActivity(), SR2ControllerHostType {
...
+      this.analyticsService.publishEvent(
+        AnalyticsEvent.BookOpened(
+          timestamp = LocalDateTime.now(),
+          credentials = this.account.loginState.credentials,
+          profileUUID = this.currentProfile.id.uuid,
+          profileDisplayName = this.currentProfile.displayName,
+          accountProvider = this.account.provider.id,
+          accountUUID = this.account.id.uuid,
+          opdsEntry = this.bookEntry.feedEntry,
+          targetURI = this.bookEntry.feedEntry.analytics.getOrNull()
```
twaddington, line 175: I'm pretty sure this is the right value to set here but I saw some funky results for some books. @io7m any ideas?

simplified-viewer-epub-readium2/src/main/java/org/nypl/simplified/viewer/epub/readium2/ReaderActivity.kt
```diff
@@ -149,6 +164,19 @@ class ReaderActivity : AppCompatActivity(), SR2ControllerHostType {
         .beginTransaction()
         .replace(R.id.reader_container, readerFragment)
         .commit()
+
+      this.analyticsService.publishEvent(
+        AnalyticsEvent.BookOpened(
+          timestamp = LocalDateTime.now(),
+          credentials = this.account.loginState.credentials,
+          profileUUID = this.currentProfile.id.uuid,
```
twaddington, line 170: None of these properties seem to be read. Maybe they're used by LFA?

io7m:
I have to rain on the parade a little. I think it might be better to implement this here:

https://github.com/NYPL-Simplified/Simplified-Android-Core/blob/develop/simplified-viewer-api/src/main/java/org/nypl/simplified/viewer/api/Viewers.kt#L31

Publish one event from the code that opens the viewer, rather than one event from each type of viewer. We did this once and then I accidentally removed it (leading to you having to write this PR in the first place). :sweat:

twaddington:
> I have to rain on the parade a little. I think it might be better to implement this here:
> 
> https://github.com/NYPL-Simplified/Simplified-Android-Core/blob/develop/simplified-viewer-api/src/main/java/org/nypl/simplified/viewer/api/Viewers.kt#L31
> 
> Publish one event from the code that opens the viewer, rather than one event from each type of viewer. We did this once and then I accidentally removed it (leading to you having to write this PR in the first place). 😓

That works for me. How should the close book event be handled then? I didn't want to remove that in case it was used by LFA.

twaddington:
@io7m updated.

io7m:
> That works for me. How should the close book event be handled then? I didn't want to remove that in case it was used by LFA.

That is a good question. As far as I can tell... Android makes it fundamentally impossible to handle reliably; it's inevitably going to be tied to a viewer activity closing, and we've no way to know if the activity closed because the user stopped reading, or it closed because that's what Android does.

twaddington:
> > That works for me. How should the close book event be handled then? I didn't want to remove that in case it was used by LFA.
> 
> That is a good question. As far as I can tell... Android makes it fundamentally impossible to handle reliably; it's inevitably going to be tied to a viewer activity closing, and we've no way to know if the activity closed because the user stopped reading, or it closed because that's what Android does.

We could publish the event only from an intentional user interaction like `onBackPressed` or if the toolbar back is pressed.

PR #805, opened 2020-09-15
SIMPLY-3074: Remove unnecessary Readium API calls when a book is opened.

ray-lee:
**What's this do?**

Remove unnecessary calls to the Readium API when a book is opened (`onReadiumContentDocumentLoaded`). Some of the calls are redundant with calls being made in other handlers. The rest can be moved into (`onReadiumFunctionInitialize`).

This fixes [SIMPLY-3074](https://jira.nypl.org/browse/SIMPLY-3074) (page turns not working when a book is opened).

Currently, SimplyE applies reader preferences (viewer settings and book styles) when a book is opened, in the `onReadiumContentDocumentLoaded` handler. This causes a race condition that can lead Readium to believe the book has no pages, so page turns don't work. Moving the calls into Readium initialization (before a book starts loading) allows Readium to have full control over when the settings and styles are applied, without the risk of SimplyE stepping on its feet.

See the comments in the JIRA issue for all the gory details.

When a book is opened, page turning should now consistently work. The user's style preferences (font face, font size, and colors) should continue to be applied, and the book should continue to open to the user's last position.

n/a

n/a

@ray-lee ran the vanilla app.

### DIFF
--- a/simplified-viewer-epub-readium1/src/main/java/org/nypl/simplified/viewer/epub/readium1/ReaderActivity.java
+++ b/simplified-viewer-epub-readium1/src/main/java/org/nypl/simplified/viewer/epub/readium1/ReaderActivity.java
@@ -330,12 +330,7 @@ public final class ReaderActivity extends AppCompatActivity implements
     final ReaderPreferences readerPreferences;
 
     try {
-      readerPreferences =
-        Services.INSTANCE.serviceDirectory()
-          .requireService(ProfilesControllerType.class)
-          .profileCurrent()
-          .preferences()
-          .getReaderPreferences();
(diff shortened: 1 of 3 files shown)


twaddington (approved):
This looks fine to me. I don't know the R1 code at all so I'm unsure what the consequences of these changes might be. I trust that it's better than what we have today.

io7m:
Same from me! It looks good. If it appears to work, then that's likely good enough.

The ~300ms delay was necessary back in 2014 as Readium didn't expose any other kind of usable API for detecting that a book had fully loaded. Good to know we can do better now.

PR #807, opened 2020-09-16
SIMPLY-3086: Various Card Creator Related bugs

MalcolmMcFly:
**What's this do?**
Various bugs have been seen in Crashlytics associated with the card creator. Most of them are related to the user exiting the app and returning and attempting to interact with dialogs. This changes how dialogs are instantiated and dismissed to relieve the crashes. Additionally, default values are setup for the barcode and username in the confirmation screen.

[SIMPLY-3086: Various Card Creator Related bugs](https://jira.nypl.org/browse/SIMPLY-3086)

1. Access the card creator and enter a valid home address in airplane mode. A dialog will be presented.
2. Background the app and turn networking back on.
3. The dialog should be dismissed

n/a

n/a

@MalcolmMcFly @twaddington has also seen this working.


twaddington (approved):
Looks good!

PR #808, opened 2020-09-17
SIMPLY-3088: Fix chapter progress in Readium 1 bookmarks.

ray-lee:
**What's this do?**

Fix chapter progress not being saved in newly created Readium 1 bookmarks, and saved bookmarks incorrectly showing "0% through chapter".

This fixes [SIMPLY-3088](https://jira.nypl.org/browse/SIMPLY-3088).

Create a bookmark that is not on the first page of a chapter. Open the bookmarks list. The bookmark should show the correct chapter progress % (not 0%). The bookmark should open to the correct page (not the first page).

In an older version of the app (before commit 15d8033851ad2bee28406457dcdb8bb58b2d14a4), create a bookmark that is not on the first page of the chapter. In the bookmarks list, verify that the correct chapter progress % is shown. Upgrade the app to a version after that commit. Open the bookmarks list. The bookmark created in the old version should continue to show the correct chapter progress % (not 0%).

Unit tests are included for deserializing saved bookmarks.

n/a

n/a

@ray-lee ran the vanilla app.

### DIFF
--- a/simplified-books-api/src/main/java/org/nypl/simplified/books/api/BookmarkJSON.kt
+++ b/simplified-books-api/src/main/java/org/nypl/simplified/books/api/BookmarkJSON.kt
@@ -60,10 +60,26 @@ object BookmarkJSON {
     node: ObjectNode
   ): Bookmark {
 
+    // Older bookmarks store chapter progress in a top-level property, instead of inside
+    // location.progress.
+
+    val chapterProgress = JSONParserUtilities.getDouble(node, "chapterProgress")
+
+    val deserializedLocation = BookLocationJSON.deserializeFromJSON(
+      objectMapper, JSONParserUtilities.getObject(node, "location"))
+
+    val location =
+      if (deserializedLocation.progress == null && chapterProgress != null)
+        // If this is an older bookmark, move the chapter progress into location.progress. In this
+        // case the chapter index is unknown.
+        deserializedLocation.copy(progress = BookChapterProgress(0, chapterProgress))
+      else
+        deserializedLocation
+
     return Bookmark(
       opdsId = JSONParserUtilities.getString(node, "opdsId"),
       kind = kind,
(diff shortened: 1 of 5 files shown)


twaddington:
I'd like @io7m to take a look at this since he touched that code last.

simplified-books-api/src/main/java/org/nypl/simplified/books/api/BookmarkJSON.kt
```diff
@@ -60,10 +60,26 @@ object BookmarkJSON {
...
+    // Older bookmarks store chapter progress in a top-level property, instead of inside
+    // location.progress.
+
+    val chapterProgress = JSONParserUtilities.getDouble(node, "chapterProgress")
+
+    val deserializedLocation = BookLocationJSON.deserializeFromJSON(
+      objectMapper, JSONParserUtilities.getObject(node, "location"))
+
+    val location =
+      if (deserializedLocation.progress == null && chapterProgress != null)
```
twaddington, line 72: Kotlin coding conventions require braces on any multiline if / else statements.

> If the condition of an if or when statement is multiline, always use curly braces around the body of the statement.

If the expression is short enough to fit on a single line you can omit the braces.

https://kotlinlang.org/docs/reference/coding-conventions.html#formatting-control-flow-statements

Honestly not sure why ktlint didn't flag this.

twaddington: Looks like there's an experimental rule to enforce this.

https://github.com/pinterest/ktlint/blob/c7bdae030dc55c/ktlint-ruleset-experimental/src/main/kotlin/com/pinterest/ktlint/ruleset/experimental/MultiLineIfElseRule.kt

ray-lee: Thanks, fixed.

io7m:
It looks right to me!

io7m (approved):

PR #809, opened 2020-09-17
SIMPLY-3089: Follow 303 (see other) redirects.

ray-lee:
**What's this do?**

Follow 303 (see other) redirects when downloading content.

This is needed to fulfill content from Internet Archive, which sends 303 redirects.

Download a book from Internet Archive, e.g. from Montana State Library (account info available from Lyrasis), "Montana cassette catalog update, June 1993 - September 1995". The download should succeed.

n/a

n/a

@ray-lee ran the vanilla app.


twaddington (approved):
Looks good to me!

PR #810, opened 2020-09-18
Bump version of ktlint; enable --experimental rules.

twaddington:


.editorconfig
```diff
@@ -1,7 +1,7 @@
 [*.kt]
 indent_size=2
 insert_final_newline=true
-disabled_rules=import-ordering
+disabled_rules=import-ordering,experimental:package-name,experimental:argument-list-wrapping
```
twaddington, line 4: I disabled these two rules, which I thought were particularly overzealous. I'm open to disabling others if folks prefer.

https://github.com/pinterest/ktlint#experimental-rules

simplified-accounts-database/src/main/java/org/nypl/simplified/accounts/database/AccountsDatabase.kt
```diff
@@ -228,32 +228,34 @@ class AccountsDatabase private constructor(
...
-          )
-        }
-        if (existingProvider.updated.isAfter(accountProvider.updated)) {
-          logger.warn(
-            "attempted to update provider {} with an older definition",
-            existingProvider.id
-          )
-          return@FunctionType existing
-        }
+          val existingProvider = existing.provider()
```
io7m, line 234: Not sure what's going on here. Another level of indentation?

twaddington: Yeah, it wrapped line 231 to 232, which caused everything below that to be indented one additional level.

twaddington: I think that's due to this:

> Method call formatting
> In long argument lists, put a line break after the opening parenthesis.

https://kotlinlang.org/docs/reference/coding-conventions.html#method-call-formatting

io7m:
I think this is fine. For the currently disabled argument list wrapping rule... Does this rule refer to method _calls_ or method _declarations_? If it refers to declarations, I'd be happy to turn the rule _on_. I think this:

```
fun f(x: Int, y:Int, z:Int): Int { ... }
```

... is objectively worse than this:

```
fun f(
  x: Int,
  y: Int,
  z: Int
): Int {
  ...
}
```

The latter is the recommended style for the Ada programming language. The DoD did a lot of research on "readability" with regards to human eyesight and cognitive processes and their general conclusion was that humans work best with < 80 character text widths, and with things carefully lined up in columns. That research was folded into the recommended formatting rules for the Ada language, and this stuff has sort of trickled into the programming languages we have today, whether anyone realizes it or not. :smile:

twaddington:
> I think this is fine. For the currently disabled argument list wrapping rule... Does this rule refer to method _calls_ or method _declarations_? If it refers to declarations, I'd be happy to turn the rule _on_. I think this:
> 
> ```
> fun f(x: Int, y:Int, z:Int): Int { ... }
> ```
> 
> ... is objectively worse than this:
> 
> ```
> fun f(
>   x: Int,
>   y: Int,
>   z: Int
> ): Int {
>   ...
> }
> ```
> 
> The latter is the recommended style for the Ada programming language. The DoD did a lot of research on "readability" with regards to human eyesight and cognitive processes and their general conclusion was that humans work best with < 80 character text widths, and with things carefully lined up in columns. That research was folded into the recommended formatting rules for the Ada language, and this stuff has sort of trickled into the programming languages we have today, whether anyone realizes it or not. 😄

Yeah, agree with you on readability. That's enforced as part of ktlint core. This experimental rule is for argument formatting at call sites.

So, for example, something like this:

```kotlin
       this.steps.currentStepFailed(
        this.revokeStrings.revokeBookDatabaseLookupFailed, UnexpectedException(e), e
       )
```

Becomes this:

```kotlin
       this.steps.currentStepFailed(
        this.revokeStrings.revokeBookDatabaseLookupFailed,
        UnexpectedException(e),
        e
       )
```

I'm not against enabling it, but wasn't sure folks would like it. It's slightly more aggressive than the default Android Studio formatting. Here's where this rule is defined.

https://github.com/pinterest/ktlint/blob/master/ktlint-ruleset-experimental/src/main/kotlin/com/pinterest/ktlint/ruleset/experimental/ArgumentListWrappingRule.kt#L23-L27

twaddington:
@io7m I enabled the rule and applied the changes in a new commit so you can see what it looks like. Easy to drop if we don't like it. Thoughts?

io7m:
I think maybe it's a bit much. :laughing: 

I think if we're going to move to putting every function call argument on its own line, we might as well mandate named parameters everywhere.

twaddington:
> I think maybe it's a bit much. 😆
> 
> I think if we're going to move to putting every function call argument on its own line, we might as well mandate named parameters everywhere.

haha, yeah, same. I'll revert that commit.

io7m (approved):

io7m:
I'll probably have to rebase the other PR against this.

PR #814, opened 2020-10-01
SIMPLY-3001: Part 2

io7m:
**What's this do?**
This completely refactors the borrowing task into a set of well-defined, single-purpose
subtasks. It effectively fixes numerous issues the app has had for quite a while:

  * The borrowing code is now actually maintainable, instead of being one massive
     2000-line tangled mess.
  * The borrowing code test suite is now clean and maintainable.
  * The app now very accurately filters out books that are not supported (such as
     encrypted PDF files). The code that performs the filtering knows what the
     current application supports, so Vanilla won't show Adobe-encrypted books.
     A flag has been added in the debug settings that allows for enabling/disabling
     filtering.
  * A lot of legacy code has been removed; borrowing string resources, the downloader
    module, etc.
  * The new borrowing code uses Simplified-Android-HTTP so that we can set policies
     in one place.
  * Handling of DRM is no longer format-specific. This paves the way for us to be
    able to support LCP for every format.
  * We now save ACSM files even if the fulfillment of those files fails. This will allow
    people to revoke loans even if they can't actually download an EPUB. This isn't
    actually implemented yet, but it's at least possible now.

Wait for it...

https://jira.nypl.org/browse/SIMPLY-3001
https://jira.nypl.org/browse/SIMPLY-2874
https://jira.nypl.org/browse/SIMPLY-3029
https://jira.nypl.org/browse/SIMPLY-3036
https://jira.nypl.org/browse/SIMPLY-3106
https://jira.nypl.org/browse/SIMPLY-3107
https://jira.nypl.org/browse/SIMPLY-3108
https://jira.nypl.org/browse/SIMPLY-3109
https://jira.nypl.org/browse/SIMPLY-3110
https://jira.nypl.org/browse/SIMPLY-3006
https://jira.nypl.org/browse/SIMPLY-3083
https://jira.nypl.org/browse/SIMPLY-3041

Check that unsupported books no longer appear in feeds (such as encrypted PDFs).
Check that borrowing unprotected EPUB files from the classics collection works.
Check that borrowing encrypted EPUB files from the NYPL collection works.
Check that borrowing an audio book works.

Depends on SIMPLY-3001: Part 1.

The module list has been updated.

@io7m tested every type of book he could find in both Vanilla and SimplyE.


io7m:
If it wasn't obvious: Part 1 should be merged before this one.

twaddington:
Will do a quick review of this in the morning. Looks like most of the changes here are due to "Part 1" not being merged yet.

simplified-books-borrowing/src/main/java/org/nypl/simplified/books/borrowing/internal/BorrowHTTP.kt
```diff
@@ -0,0 +1,216 @@
...
+   * Encode the given problem report as a set of presentable attributes.
+   */
+
+  fun problemReportAsAttributes(
+    problemReport: LSHTTPProblemReport?
+  ): Map<String, String> {
+    return when (problemReport) {
+      null -> mapOf()
+      else -> {
+        val attributes = mutableMapOf<String, String>()
```
twaddington, line 48: You could make this a bit more fluent using the `to` infix operator:

```
else -> mapOf(
  "HTTP problem detail" to problemReport.detail ?: ""
  "HTTP problem status" to problemReport.status.toString()
  // ...
)
```

Little bit of syntactic sugar but otherwise this looks fine as-is.

simplified-books-borrowing/src/main/java/org/nypl/simplified/books/borrowing/internal/BorrowLoanCreate.kt
```diff
@@ -0,0 +1,309 @@
...
+
+  private fun handleHTTPError(
+    context: BorrowContextType,
+    status: LSHTTPResponseStatus.Responded.Error
+  ) {
+    context.taskRecorder.addAttributes(BorrowHTTP.problemReportAsAttributes(status.problemReport))
+
+    val report = status.problemReport
+    if (report != null) {
+      if (report.type == "http://librarysimplified.org/terms/problem/loan-already-exists") {
```
twaddington, line 124: Possible for this to ever be an `https` type and break this equality check?

twaddington (approved):
Lots of changes but I didn't see anything too concerning. I noted a few small issues but nothing that should block this from being merged.

simplified-opds-core/src/main/java/org/nypl/simplified/opds/core/OPDSAvailabilityLoaned.kt
```diff
@@ -0,0 +1,89 @@
...
+
+  val startDate: OptionType<DateTime>,
+
+  /**
+   * @return A URI for revoking the hold, if any
+   */
+
+  val revoke: OptionType<URI>,
+
+  private val end_date: OptionType<DateTime>
```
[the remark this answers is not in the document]
io7m: Accidentally left in when automatically converting from Java.

simplified-books-borrowing/src/main/java/org/nypl/simplified/books/borrowing/internal/BorrowLoanCreate.kt
io7m: Don't think so. It's similar to a namespace identifier in XML in that it's a URL but it doesn't actually describe a real resource. I wish we'd used URNs for these!

twaddington: Perfect. Just checking!

simplified-opds-core/src/main/java/org/nypl/simplified/opds/core/OPDSAvailabilityLoaned.kt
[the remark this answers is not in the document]
twaddington: Figured it was something like that!

PR #815, opened 2020-10-01
SIMPLY-3124: Fix incorrect pagination after changing reader settings.

ray-lee:
**What's this do?**

Remove the page turn transition animation, refactor font handling, and refactor `applyReaderPreferences` to remove unnecessary Readium API calls.

The CSS transition to create the page turn animation is removed. When the transition is present, Readium can operate on the DOM when it is in a transitional state, causing it to calculate an incorrect page number/number of pages.

Font loading and changing are refactored to use Readium APIs instead of side-loading through CSS. This prevents Readium from doing DOM operations before a font has loaded.

`applyReaderPreferences` is refactored to remove unneeded calls that can cause race conditions. Currently, `applyReaderPreferences` does the following: https://github.com/NYPL-Simplified/Simplified-Android-Core/blob/c96815761016d0b4687c6781ead97acc58ae242c/simplified-viewer-epub-readium1/src/main/java/org/nypl/simplified/viewer/epub/readium1/ReaderActivity.java#L524-L538

- `setBookStyles` and `applyViewerColorScheme` only need to be called if the color scheme is changed.
- `updateSettings` only needs to be called if the font family or font scale is changed.
- `getReadiumCFI` and `setReadiumCFI` don't need to be called, because when the font family or scale is changed, `updateSettings` already saves and restores the current reading position; and when the color scheme is changed, it can't cause a change in pagination.
- `getCurrentPage` and `mediaOverlayIsAvailable` don't need to be called here, because if the pagination changes, they will be called in `onReadiumFunctionPaginationChanged`.

Also, the call to `pageHasChanged` is moved from `onReadiumFunctionPaginationChanged` to `onReadiumContentDocumentLoaded`, which is more appropriate. I'm not sure if it was causing any problems, but it was doing extra work, because it only needs to be run when the content document has been loaded, not whenever its page layout changes.

This fixes [SIMPLY-3124](https://jira.nypl.org/browse/SIMPLY-3124).

Open a book to the middle of a chapter. Change the font and font size. In the chapter progress at the bottom of the screen, the current page and number of pages may change, but they should reflect the actual page displayed, and the actual number of pages in the chapter. Change the color scheme. The current page and number of pages should not change.

n/a

n/a

@ray-lee ran the vanilla app.

### DIFF
--- a/simplified-viewer-epub-readium1/src/main/assets/fonts.css
+++ b/simplified-viewer-epub-readium1/src/main/assets/fonts.css
@@ -1,0 +1,5 @@
+@font-face {
+  font-family: 'OpenDyslexic3';
+  src: url('OpenDyslexic3-Regular.ttf');
+  font-weight: normal;
+}
--- a/simplified-viewer-epub-readium1/src/main/assets/host_app_feedback.js
+++ b/simplified-viewer-epub-readium1/src/main/assets/host_app_feedback.js
@@ -92,7 +92,21 @@ $(document).ready(function()
       var opts = {
         needsFixedLayoutScalerWorkAround: true,
         el: "#viewport",
-        annotationCSSUrl: '/readium_Annotations.css'
+        annotationCSSUrl: '/readium_Annotations.css',
+        fonts: [
+          {
+            fontFamily: "serif",
+            url: "/fonts.css"
+          },
+          {
+            fontFamily: "sans-serif",
+            url: "/fonts.css"
+          },
+          {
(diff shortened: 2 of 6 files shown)


io7m:
I can't comment on the changes with respect to R1 as I suspect at this point that you know more about R1 than all of us combined.

The actual code changes look good to me, though!

io7m (approved):

twaddington:
Looks okay to me!

twaddington (approved):

PR #817, opened 2020-10-13
Ensure DPLA audio books are used correctly

io7m:
**What's this do?**
DPLA audio books use a content type that uses a custom profile. These
changes ensure that we handle all of the possible types correctly.

Affects: https://jira.nypl.org/browse/SIMPLY-2301

Ensure that books in the DPLA Audiobook Test collection can be downloaded and read.

None.

N/A

@io7m tested the two books in the collection.


twaddington (approved):
Looks good. Nice and straightforward.

PR #818, opened 2020-10-15
Refactor custom OPDS creation to use new HTTP client

io7m:
**What's this do?**
This updates the custom OPDS creation task to use the new
Simplified-Android-HTTP client. This ensures we follow redirects
correctly, and makes the code a bit nicer. It also removes some
code that would now be duplicated.

https://jira.nypl.org/browse/SIMPLY-2065

Check that a custom OPDS feed can be created by entering a URL.

None.

N/A

@io7m created a QA account

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -79,7 +79,7 @@ ext {
 
   nyplAudioBookAPIVersion = "6.1.0"
   nyplR2Version = "0.0.5"
-  nyplHTTPVersion = "0.0.13"
+  nyplHTTPVersion = "0.0.14"
   irradiaMimeVersion = "1.1.1"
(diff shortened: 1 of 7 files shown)


twaddington:
Thanks Mark, this looks good. Was the original failure that Leonard reported just an issue with redirects?

twaddington (approved):

io7m:
> Thanks Mark, this looks good. Was the original failure that Leonard reported just an issue with redirects?

Yep, I think that may've been all it was. We were getting a 302, not following it, and then parsing the resulting document (HTML) as if it was an OPDS feed...

PR #820, opened 2020-10-21
Fix Adobe deactivations

io7m:
**What's this do?**
This appears to fix the long-standing issue of being unable to deactivate
Adobe DRM devices on Android. It turned out that the code was correct, but
that it was necessary to perform the device deactivation using a fresh
client token. This was discovered after @ettore noticed some odd calls
the iOS client was making, and this lead to a discussion with @leonardr
in the ticket https://jira.nypl.org/browse/SIMPLY-3208 that lead me to
try deactivation with a fresh token. I've so far been able to deactivate
the device every time, with no sign of the dreaded E_ADEPT_LOGIN_ERROR.

Thanks @ettore, @leonardr!

Affects: https://jira.nypl.org/browse/SIMPLY-3208
Affects: https://jira.nypl.org/browse/SIMPLY-3099
Affects: https://jira.nypl.org/browse/SIMPLY-2980

Try logging in to any account that requires Adobe DRM, and then try logging out again.

None.

N/A

@io7m logged in and out of multiple accounts

### DIFF
--- a/simplified-books-controller/src/main/java/org/nypl/simplified/books/controller/ProfileAccountLogoutTask.kt
+++ b/simplified-books-controller/src/main/java/org/nypl/simplified/books/controller/ProfileAccountLogoutTask.kt
@@ -4,6 +4,7 @@ import com.google.common.base.Preconditions
 import com.io7m.jfunctional.Option
 import org.nypl.drm.core.AdobeAdeptExecutorType
 import org.nypl.simplified.accounts.api.AccountAuthenticatedHTTP
+import org.nypl.simplified.accounts.api.AccountAuthenticationAdobeClientToken
 import org.nypl.simplified.accounts.api.AccountAuthenticationAdobePreActivationCredentials
 import org.nypl.simplified.accounts.api.AccountAuthenticationCredentials
 import org.nypl.simplified.accounts.api.AccountLoginState.AccountLoggedIn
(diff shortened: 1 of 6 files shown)


twaddington (approved):
Nice work y'all! It takes a village sometimes.

io7m:
Going to merge this. Whatever the consequences, it can't be worse than the current logout behaviour. This whole login/logout code needs a lot of work (including some UI work). This is a bandaid at best.

simplified-books-controller/src/main/java/org/nypl/simplified/books/controller/ProfileAccountLogoutTask.kt
```diff
@@ -140,6 +158,13 @@ class ProfileAccountLogoutTask(
     }
 
     this.debug("device is activated and DRM is supported, running deactivation")
+    val token = this.handlePatronUserProfile()
+    if (token == null) {
+      this.warn("Patron user profile contained no Adobe DRM client token")
+      val message = "Patron user profile is missing DRM information."
+      this.steps.currentStepFailed(message, "patronUserProfileNoDRM")
+      throw IOException(message)
```
leonardr, line 166: It's possible for a library to exist in the system but not offer any Adobe client token functionality. We don't have any libraries like that right now (apart from open-access libraries, which don't have "logout") but we will in the future. Is this going to prevent "logout" from working for that sort of library?

io7m: Yes, it might. This code is due some heavy refactoring like the borrow operation recently received. This'll need to be folded into that refactoring work.

PR #821, opened 2020-10-30
Make ServiceDirectoryProviderType.serviceDirectory() block

io7m:
**What's this do?**
This switches to a model where asking for the service directory is now
a blocking operation. In theory, we block indefinitely.  In practice,
we place a three minute bound on the operation: If it's taking longer
than this to retrieve the service directory, something is seriously
wrong and we should abort, print a stack trace, and then throw an
exception.

This should fix any and all problems where we get a lifecycle-related
exception, and may also remove the need to use Kotlin `lazy` properties.

I've seen numerous issues over the past few months where we've requested the service directory before it exists. The Android lifecycle is one big design flaw.

No specific tests. If this is working, there should be no noticeable difference in the app.

None.

N/A

I ran the vanilla app and opened a few books.

### DIFF
--- a/simplified-services-api/src/main/java/org/librarysimplified/services/api/Services.kt
+++ b/simplified-services-api/src/main/java/org/librarysimplified/services/api/Services.kt
@@ -2,29 +2,27 @@ package org.librarysimplified.services.api
 
 import com.google.common.util.concurrent.ListenableFuture
 import com.google.common.util.concurrent.SettableFuture
-import java.lang.IllegalStateException
+import org.slf4j.LoggerFactory
 import java.util.concurrent.TimeUnit
 
 object Services : ServiceDirectoryProviderType {
 
+  private val logger = LoggerFactory.getLogger(Services::class.java)
   private val servicesLock: Any = Any()
   private var servicesDirectory: ServiceDirectoryType? = null
   private val servicesFuture: SettableFuture<ServiceDirectoryType> = SettableFuture.create()
 
   override fun serviceDirectory(): ServiceDirectoryType {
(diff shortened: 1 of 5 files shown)


simplified-services-api/src/main/java/org/librarysimplified/services/api/Services.kt
```diff
@@ -2,30 +2,28 @@ package org.librarysimplified.services.api
...
+  private val logger = LoggerFactory.getLogger(Services::class.java)
   private val servicesLock: Any = Any()
   private var servicesDirectory: ServiceDirectoryType? = null
   private val servicesFuture: SettableFuture<ServiceDirectoryType> = SettableFuture.create()
 
   override fun serviceDirectory(): ServiceDirectoryType {
-    return synchronized(this.servicesLock) {
-      this.servicesDirectory ?: throw IllegalStateException("No service directory has been created!")
+    try {
+      return this.servicesFuture.get(3L, TimeUnit.MINUTES)
```
twaddington, line 17: Three minutes is a very long time in the world of mobile apps. I'd feel a little more comfortable sticking with 30-60 seconds. Have we ever seen it taking longer than that?

twaddington (approved):
I think this is a sensible change. My only concern is that three minutes is a pretty long time for it to block if something is wrong.

simplified-services-api/src/main/java/org/librarysimplified/services/api/Services.kt
io7m: Let's go back to 30.

PR #822, opened 2020-10-30
Temporarily fix custom OPDS feeds

io7m:
**What's this do?**
Somehow, I managed to break the custom OPDS feed feature again. This
fixes it until we replace all of the uses of the old HTTP client with
the new one.

Custom OPDS feeds are required for the integration tests.

Try to create a custom OPDS feed in the hidden debug menu.

None.

N/A

I created an account (again) on the NYPL QA server.

### DIFF
--- a/simplified-accounts-source-nyplregistry/src/main/java/org/nypl/simplified/accounts/source/nyplregistry/AccountProviderResolution.kt
+++ b/simplified-accounts-source-nyplregistry/src/main/java/org/nypl/simplified/accounts/source/nyplregistry/AccountProviderResolution.kt
@@ -373,7 +373,7 @@ class AccountProviderResolution(
 
   private fun contentTypeOf(result: HTTPResultError<InputStream>): String {
     val contentTypeHeaders = result.responseHeaders["content-type"]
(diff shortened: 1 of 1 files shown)


simplified-accounts-source-nyplregistry/src/main/java/org/nypl/simplified/accounts/source/nyplregistry/AccountProviderResolution.kt
```diff
@@ -366,6 +371,11 @@ class AccountProviderResolution(
     }
   }
 
+  private fun contentTypeOf(result: HTTPResultError<InputStream>): String {
+    val contentTypeHeaders = result.responseHeaders["content-type"]
+    return contentTypeHeaders ?.firstOrNull() ?: "application/octet-stream"
```
twaddington, line 376: Looks like an extra space snuck in here.

twaddington:
Looks fine, just one little nit.

leonardr:
In the vanilla build based on this PR I can get all the way through creating an account for https://qa-circulation.librarysimplified.org/feed/qa, but that's not the OPDS feed the app actually fetches -- it fetches the server's normal top-level feed instead.

PR #823, opened 2020-11-09
SIMPLY-3140: Remove simplified-http

io7m:
**What's this do?**
This replaces all uses of the `simplified-http` module in the app with our new HTTP client, and removes the `simplified-http` module. In the process, it was necessary to replace the `simplified-documents` module with something a little more modern.

The card creator hasn't been touched, as this uses `okhttp` and `retrofit` directly.

https://jira.nypl.org/browse/SIMPLY-3140

1. Check that feeds can be loaded.
2. Check that bookmarks can be synced.
3. Check that accounts can be created.
4. Check that books can be returned.
5. Check that analytics events make it to the server.
6. Check that, on a fresh install of SimplyE, you _do_ see the EULA on startup. 
7. Check that, on an existing install of SimplyE, you _don't_ see the EULA on startup (assuming that you previously agreed to it).

SimplyE needs to be updated to provide a new document store configuration service (PR incoming!).

The module list was updated.

@io7m tried the Vanilla app, and SimplyE (PR incoming there!)


io7m:
Fixing conflicts now...

simplified-app-vanilla/src/main/java/org/nypl/simplified/vanilla/VanillaDocumentStoreConfiguration.kt
```diff
@@ -0,0 +1,26 @@
...
+
+class VanillaDocumentStoreConfiguration : DocumentConfigurationServiceType {
+
+  override val privacyPolicy: DocumentConfiguration? =
+    null
+
+  override val about: DocumentConfiguration? =
+    DocumentConfiguration(
+      name = "about.html",
+      remoteURI = URI.create("http://localhost/about.html")
```
io7m, line 15: This is just until we have this document actually hosted somewhere. Could probably host it in the GitHub repository.

simplified-books-covers/src/main/java/org/nypl/simplified/books/covers/BookCoverGenerator.java
```diff
@@ -104,7 +104,10 @@ public BookCoverGenerator(
     final SortedMap<String, String> params = new TreeMap<String, String>();
     params.put("title", NullCheck.notNull(title));
     params.put("author", NullCheck.notNull(author));
-    return URIQueryBuilder.encodeQuery(
-      NullCheck.notNull(URI.create("generated-cover://localhost/")), params);
+
+    return LSHTTPURIQueryBuilder.INSTANCE.encodeQuery(
```
twaddington, line 108: You should be able to make this more java friendly with the `@JvmStatic` annotation.

https://kotlinlang.org/docs/reference/java-to-kotlin-interop.html#static-methods

twaddington (approved):
Looks great! So nice to see all those deletions. Left a couple comments to consider but nothing too important.

simplified-app-vanilla/src/main/java/org/nypl/simplified/vanilla/VanillaDocumentStoreConfiguration.kt
io7m: I feel like SimplyE had one and now doesn't show it... We really need to redo the whole set of documents.

simplified-books-covers/src/main/java/org/nypl/simplified/books/covers/BookCoverGenerator.java
io7m: Yep. I didn't bother here given that we have very little Java code left.

PR #824, opened 2020-11-11
Remove boot string resources

io7m:
**What's this do?**
This removes a ton of boot string resources that we'll never need
to translate. This reduces the translation burden for projects like
LFA, and removes a lot of boilerplate code.

It also took the opportunity to clean up the boot process, which
had a couple of instances of submitting services multiple times.

Spring cleaning in the middle of winter. Also, it helps LFA. :smile:

If the application starts up, this works.

Depends on the new HTTP client PR.

N/A

@io7m started the app


twaddington (approved):
Looks good!

PR #826, opened 2020-11-17
Refactor use of Toolbar, Home Action, and menus; removed simpified-ui-toolbar.

twaddington:
**What's this do?**
This refactor makes several important improvements.

- The simplified-ui-toolbar module has been removed in favor of Android's native APIs.
- Fragments now use the proper APIs for showing an options menu.
- Toolbar titles and subtitles are more helpful and better describe what you're looking at.
- Tapping the "up/home" affordance in the Toolbar now takes you to the root of your current tab. This better adheres to the expected behavior for most Android apps.

Incidental changes include:

- Refactoring the Search dialog to be more user friendly.
- Minor fixes for to the book detail layout.
- Addition of specific styles for the Toolbar and ActionBar.

This improves the patron experience when navigating down into deep hierarchies within the catalog.

Please run this code and poke around the app paying special attention to the first run experience; drilling down into catalog feed lanes; and using the search feature.

Note any unexpected behaviors.

n/a

n/a

@twaddington ran the vanilla app and did extensive manual testing.


twaddington:
After this change is approved I will squash merge into develop.

io7m:
*salivates*

I'll be looking at this when the CI work stops churning.

io7m (approved):
I like it! Always glad to see code go away.

I still think having "up" go to the root instead of to the structural parent of the feed is wrong, but I apparently am losing that argument with the rest of the world.

twaddington:
> I like it! Always glad to see code go away.

Thanks for looking!

> I still think having "up" go to the root instead of to the structural parent of the feed is wrong, but I apparently am losing that argument with the rest of the world.

Yeah, it's one of those things that people have strong opinions about. Luckily, if the new behavior is strongly disliked, switching back is a one-line change in `MainActivity`. Just need to switch `controller.popToRoot()` to `controller.popBackStack()` and you're done.

PR #828, opened 2020-11-20
SIMPLY-3259: Open eBooks: Accounts Button

MalcolmMcFly:
**What's this do?**
Here the `showChangeAccountsUi` attribute is added to the `BuildConfigurationAccountsType` to allow for the hiding of the change account option on the Catalog screen.

[SIMPLY-3259: Open eBooks: Accounts Button](https://jira.nypl.org/browse/SIMPLY-3259)

Toggle the `showChangeAccountsUi` attribute true/false to see if the change account screen his hidden/shown on the Catalog screen with in the vanilla app with `VanillaBuildConfigurationService`.

n/a

n/a

@MalcolmMcFly

### DIFF
--- a/simplified-ui-catalog/src/main/java/org/nypl/simplified/ui/catalog/CatalogFragmentFeed.kt
+++ b/simplified-ui-catalog/src/main/java/org/nypl/simplified/ui/catalog/CatalogFragmentFeed.kt
@@ -743,10 +743,12 @@ class CatalogFragmentFeed : Fragment() {
       this.supportActionBar?.apply {
         // Configure whether or not the user should be able to change accounts
         if (configurationService.showChangeAccountsUi) {
-            setHomeAsUpIndicator(R.drawable.accounts)
-            setHomeActionContentDescription(R.string.catalogAccounts)
+          setHomeAsUpIndicator(R.drawable.accounts)
+          setHomeActionContentDescription(R.string.catalogAccounts)
+          setDisplayHomeAsUpEnabled(true)
+        } else {
+          setDisplayHomeAsUpEnabled(false)
         }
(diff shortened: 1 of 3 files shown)


simplified-ui-catalog/src/main/java/org/nypl/simplified/ui/catalog/CatalogFragmentFeed.kt
```diff
@@ -741,9 +741,12 @@ class CatalogFragmentFeed : Fragment() {
...
       this.supportActionBar?.apply {
-        setHomeAsUpIndicator(R.drawable.accounts)
-        setHomeActionContentDescription(R.string.catalogAccounts)
-        setDisplayHomeAsUpEnabled(true)
+        // Configure whether or not the user should be able to change accounts
+        if (configurationService.showChangeAccountsUi) {
+            setHomeAsUpIndicator(R.drawable.accounts)
+            setHomeActionContentDescription(R.string.catalogAccounts)
+        }
+        setDisplayHomeAsUpEnabled(configurationService.showChangeAccountsUi)
```
twaddington, line 749: I think this would read a little better as an if/else, but it's the same logic:

```kotlin
if (configurationService.showChangeAccountsUi) {
  setHomeAsUpIndicator(R.drawable.accounts)
  setHomeActionContentDescription(R.string.catalogAccounts)
  setDisplayHomeAsUpEnabled(true)
} else {
  setDisplayHomeAsUpEnabled(false)
}
```

twaddington (approved):
Looks good!

io7m (approved):
Looks good! :+1:

io7m:
I pushed a fix to `develop` that corrects the build failure. I was accidentally requiring signing properties even when signing was turned off.

PR #829, opened 2020-11-20
Simplify code for loading account icons.

twaddington:
**What's this do?**
This is a small refactor of the `ImageAccountIcons` code to eliminate some unnecessary work.

Picasso is trying to fetch every image resource in the account list, even if they're null. This is a lot of unnecessary work and can cause the account list to become choppy when scrolling around. This was easily fixed by just doing nothing if the uri is null.

Additionally, use the default error and placeholder functions of Picasso for setting the default image.

n/a

n/a

n/a

### DIFF
--- a/simplified-ui-images/src/main/java/org/nypl/simplified/ui/images/ImageAccountIcons.kt
+++ b/simplified-ui-images/src/main/java/org/nypl/simplified/ui/images/ImageAccountIcons.kt
@@ -1,14 +1,10 @@
 package org.nypl.simplified.ui.images
 
-import android.view.View
 import android.widget.ImageView
 import androidx.annotation.DrawableRes
-import com.squareup.picasso.Callback
 import com.squareup.picasso.Picasso
-import com.squareup.picasso.RequestCreator
 import org.nypl.simplified.accounts.api.AccountProviderDescription
 import org.slf4j.LoggerFactory
-import java.net.URI
 
 /**
  * Functions to efficiently load account logos.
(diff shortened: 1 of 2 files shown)


simplified-ui-images/src/main/java/org/nypl/simplified/ui/images/ImageAccountIcons.kt
```diff
@@ -31,28 +27,13 @@ object ImageAccountIcons {
...
-
-        override fun onError(e: Exception) {
-          LOG.error("failed to load account icon: ", e)
-          iconView.setImageResource(defaultIcon)
-          iconView.visibility = View.VISIBLE
-        }
-      }
-    )
+    val uri = account.logoURI?.hrefURI ?: return
+    this.logger.debug("configuring account logo: {}", uri)
```
io7m, line 31: We've got to adjust or eliminate this logging at some point. This log statement in particular ends up creating a lot of entropy in logs that reduces our ability to gzip them for error reports.

io7m (approved):
I like it!

simplified-ui-images/src/main/java/org/nypl/simplified/ui/images/ImageAccountIcons.kt
twaddington: Yeah, we could probably move a lot of these to trace level logs. Is that what you were thinking?

io7m: Yeah, something along those lines.

PR #831, opened 2020-11-20
SIMPLY-3305: Explicitly send empty passwords for Overdrive

io7m:
**What's this do?**
This adjusts various calls to the audio book API so that we explicitly
signal that a password is not required. This is used by backends such
as Overdrive to send slightly different token requests.

https://jira.nypl.org/browse/SIMPLY-3305

Get an audio book from Kalama Public Library. They use Overdrive but do not use passwords.

None.

N/A

@io7m downloaded and listened to a book from Kalama


twaddington (approved):
Looks good to me!

PR #836, opened 2020-12-02
Remove version code from archive name; fix version name when run on github actions.

twaddington:


io7m:
Current things we have as versions:

* Semantic version in `VERSION_NAME`. This is exposed via the UI, via "Application Info..." in Android, and visibly on the Play Store. `-SNAPSHOT` until we tag a release and increment the minor number.
* Android version code. Tied to `VERSION_NAME`.
* Git commit. This is visible in the app, and in the `android-binaries` repos. Changes on every commit, obviously. There are two commits visible in SimplyE due to the separate core and there's an `M:N` relationship between SimplyE commits and Core commits.
* `GITHUB_RUN_NUMBER`. Visible in the `android-binaries` repos. Changes on every CI run.

I don't know which one QA are going to look at... Or what people are going to refer to when they send us screenshots with bugs.

:weary:

twaddington:
Yeah, we should have a chat with QA and talk about all these different versions. If we're giving a release candidate to QA for testing shouldn't we be incrementing the version name (semvar)?

PR #838, opened 2020-12-07
SIMPLY-2936: Implement announcements

io7m:
**What's this do?**
This provides code to parse announcements from library authentication
documents, to display announcements on particular account/profile
events, and to remember which announcements have been acknowledged.

Announcements are checked whenever an account is updated or profile
preferences are updated. In practice, this happens when we sync accounts
and when we switch accounts in the UI. Acknowledged announcements
are recorded in the preferences for each account. I'm unsure if this is the
behaviour we want, but it seems OK. The announcements are fairly
spartan looking `AlertDialog` views, and multiple announcements are
paged into a single dialog box.

https://jira.nypl.org/browse/SIMPLY-2936

1. Switch to the NYPL QA server account.
2. Note that you get an alert dialog popping up with a frazzled-looking snowman in it and the message "Mark is great!"
3. Close the dialog.
4. Note that you don't see the dialog ever again.
5. Click the "Forget All Announcements" button in the developer debug menu to forget any acknowledged announcements and potentially re-show them on each account.

None.

Regenerated the module list.

@io7m tested against the QA server


simplified-ui-announcements/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+
+<resources>
+  <string name="announcementTitle">%1$s (%2$d of %3$d)</string>
```
twaddington, line 4: Since these resource names aren't prefixed you should make them "private" by including a `res/values/public.xml` file with the contents:

```
<?xml version="1.0" encoding="UTF-8" ?>
<resources>
  <public />
</resources>
```

This effectively hides these resources from other modules and prevents them from showing up in autocomplete dialogs within Android Studio.

simplified-ui-announcements/src/main/java/org/nypl/simplified/ui/announcements/AnnouncementsController.kt
```diff
@@ -0,0 +1,176 @@
...
+
+    val title =
+      context.getString(
+        R.string.announcementTitle,
+        account.provider.displayName,
+        indexNumber + 1,
+        announcements.size
+      )
+
+    val dialog =
```
twaddington, line 133: I don't think a dialog is an ideal ux for this. Instead, I'd suggest we show a snackbar with the text `%d unread announcements`. When the user taps the snackbar we push a new fragment onto the stack that displays the announcements.

twaddington (commented):
This looks fine to me. I left a suggestion for an improved announcements-ui; but that can be handled as a follow-on task.

simplified-ui-announcements/src/main/res/values/strings.xml
twaddington: I've been doing this in new modules but we should honestly mark all module resources private except those meant to be exported like in `ui-theme` or other `ui-` prefixed modules.

twaddington: Went ahead and did it 😅

simplified-ui-announcements/src/main/java/org/nypl/simplified/ui/announcements/AnnouncementsController.kt
io7m: I agree, I'd rather have a separate UI entirely that allows for reading a scrolling list of announcements. They just said dialog box, though... :slightly_frowning_face:

io7m:
Fixing these merge conflicts now.

PR #830, opened 2020-11-20
SIMPLY-3255: Show Report Issue in account views

io7m:
**What's this do?**
This re-enables the Report Issue item in the account view. When
selected, the user's default email client is opened in with the
published support email address.

**Why are we doing this? (w/ JIRA link if applicable)**
https://jira.nypl.org/browse/SIMPLY-3255

**How should this be tested? / Do these changes have associated tests?**
Visit the account screen for a collection that you know publishes a support email. Check that the Report Issue item is there and that it offers to open an email client.
Visit the account screen for a collection that you know _doesn't_ publish a support email. Check that the Report Issue item is gone.

**Dependencies for merging? Releasing to production?**
None.

**Has the application documentation been updated for these changes?**
N/A

**Did someone actually run this code to verify it works?**
@io7m tested various accounts with support email addresses, but couldn't find a collection without one.

PR #839, opened 2020-12-07
Display email in accounts screen

io7m:
**What's this do?**
This displays the email address used for support requests in the
account screen.

Affects: https://jira.nypl.org/browse/SIMPLY-3255

Same as for https://github.com/NYPL-Simplified/Simplified-Android-Core/pull/830
Unfortunately, some devices just don't have working email apps. This adds an email display so someone can conceivably send an email manually instead.

None.

N/A

@io7m

### DIFF
--- a/simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/AccountFragment.kt
+++ b/simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/AccountFragment.kt
@@ -1,7 +1,6 @@
 package org.nypl.simplified.ui.accounts
 
 import android.app.Activity
-import android.app.AlertDialog
 import android.content.Intent
 import android.net.Uri
 import android.os.Bundle
@@ -17,6 +16,7 @@ import android.widget.ProgressBar
 import android.widget.Switch
 import android.widget.TextView
 import androidx.annotation.UiThread
+import androidx.appcompat.app.AlertDialog
 import androidx.constraintlayout.widget.ConstraintLayout
 import androidx.fragment.app.Fragment
 import androidx.lifecycle.ViewModelProviders
@@ -572,9 +572,19 @@ class AccountFragment : Fragment() {
       this.reportIssueGroup.setOnClickListener {
         val emailIntent =
           Intent(Intent.ACTION_SENDTO, Uri.fromParts("mailto", email, null))
-        startActivity(
-          Intent.createChooser(emailIntent, resources.getString(R.string.accountReportIssue))
-        )
+        val chosenIntent =
+          Intent.createChooser(emailIntent, this.resources.getString(R.string.accountReportIssue))
+
+        try {
+          this.startActivity(chosenIntent)
+        } catch (e: Exception) {
+          this.logger.error("unable to start activity: ", e)
+          val context = this.requireContext()
+          AlertDialog.Builder(context)
+            .setMessage(context.getString(R.string.accountReportFailed, email))
+            .create()
+            .show()
+        }
       }
     } else {
       this.reportIssueGroup.visibility = View.GONE
--- a/simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/AccountFragment.kt
+++ b/simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/AccountFragment.kt
@@ -119,13 +119,14 @@ class AccountFragment : Fragment() {
   private lateinit var loginTitle: ViewGroup
   private lateinit var parameters: AccountFragmentParameters
   private lateinit var profilesController: ProfilesControllerType
+  private lateinit var reportIssueEmail: TextView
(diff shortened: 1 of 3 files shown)


simplified-ui-accounts/src/main/res/layout/account.xml
```diff
@@ -274,18 +274,32 @@
...
 
+      <TextView
+        android:id="@+id/accountReportIssueEmail"
+        android:layout_width="0dp"
+        android:layout_height="wrap_content"
+        android:layout_marginStart="16dp"
+        android:layout_marginEnd="16dp"
+        android:layout_marginBottom="16dp"
+        android:fontFamily="monospace"
+        android:text="someone@example.com"
```
twaddington, line 298: Please use `tools:text` here. Any attributes with this namespace are stripped on compile.

simplified-ui-accounts/src/main/res/layout/account.xml
```diff
@@ -258,11 +258,11 @@
     </androidx.constraintlayout.widget.ConstraintLayout>
 
     <androidx.constraintlayout.widget.ConstraintLayout
-      android:visibility="gone"
       android:id="@+id/accountReportIssue"
       android:layout_width="match_parent"
       android:layout_height="wrap_content"
-      android:background="?attr/selectableItemBackground">
+      android:background="?attr/selectableItemBackground"
+      android:visibility="visible">
```
twaddington, line 265: Visible is the default so you can just remove this attribute entirely.

simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/AccountFragment.kt
```diff
@@ -565,7 +568,8 @@ class AccountFragment : Fragment() {
     val email = this.account.provider.supportEmail
     if (email != null) {
       this.reportIssueGroup.visibility = View.VISIBLE
-      this.reportIssueItem.setOnClickListener {
+      this.reportIssueEmail.text = email.removePrefix("mailto:")
+      this.reportIssueGroup.setOnClickListener {
         val emailIntent =
           Intent(Intent.ACTION_SENDTO, Uri.fromParts("mailto", email, null))
         startActivity(
```
twaddington, line 575: It'd be good to wrap this with a call to `resolveActivity` so it doesn't crash on devices lacking an email client:

```kotlin
    if (intent.resolveActivity(packageManager) != null) {
        startActivity(intent)
    }
```

https://developer.android.com/guide/components/intents-common#Email

simplified-ui-accounts/src/main/res/layout/account.xml
io7m: Oops! :+1:

simplified-ui-accounts/src/main/res/layout/account.xml
io7m: What're the practical consequences of that? Is that just so that the `someone@example.com` string doesn't actually appear in the app, but _does_ appear in the layout editor?

simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/AccountFragment.kt
io7m: Hm! What we're actually seeing in practice is that devices that lack a configured email app just seem to silently fail here (hence the need to display the email address). I'll wrap it anyway though.

io7m: Actually, having worked through the chain of method calls, I'm not sure about this one. `Intent.createChooser` can't fail in any way that yields an exception. The `startActivity` call _can_ fail by raising an `ActivityNotFound` exception though...

simplified-ui-accounts/src/main/res/layout/account.xml
twaddington: > Is that just so that the `someone@example.com` string doesn't actually appear in the app, but does appear in the layout editor? &mdash; @io7m 

Yep! That's exactly right.

https://developer.android.com/studio/write/tool-attributes

simplified-ui-accounts/src/main/res/layout/account.xml
twaddington: Sorry, total nitpick 😅

simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/AccountFragment.kt
twaddington: Yeah, it's not essential, this would be pretty rare in practice, but could happen on a managed device like the iPads that were sent out earlier in the year. Just a best practice sort of thing.

You did exactly what I've done in the past, which is just show a dialog in the case of a failure here.

twaddington (approved):
Looks good!

PR #840, opened 2020-12-09
SIMPLY-3362: Delete caches

io7m:
**What's this do?**
This makes two main changes:

  1. Responses returned from the R1 web server have Cache-Control
     headers to prevent caching of data.
  2. The web view cache is deleted on starting and ending.

https://jira.nypl.org/browse/SIMPLY-3362

Check that your webview cache remains generally empty when reading books.

None.

N/A

@io7m tested that the cache stayed mostly empty...

cache0
cache1

The screenshots show that only a couple of internal javascript files are cached. No HTML pages appear.

### DIFF
--- a/simplified-viewer-epub-readium1/src/main/java/org/nypl/simplified/viewer/epub/readium1/ReaderHTTPServerAAsync.java
+++ b/simplified-viewer-epub-readium1/src/main/java/org/nypl/simplified/viewer/epub/readium1/ReaderHTTPServerAAsync.java
@@ -145,6 +145,7 @@ public final class ReaderHTTPServerAAsync
   {
     try {
       final ReaderNativeCodeReadLock read_lock = ReaderNativeCodeReadLock.get();
+      disableCache(response);
 
       /**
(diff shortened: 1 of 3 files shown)


io7m:
Fixing test failures now. Amusingly, I didn't realize we still had working tests for this code. :laughing:

MalcolmMcFly (approved):

simplified-viewer-epub-readium1/src/main/java/org/nypl/simplified/viewer/epub/readium1/ReaderHTTPServerAAsync.java
```diff
@@ -283,6 +284,10 @@ public static ReaderHTTPServerType newServer(
     }
   }
 
+  private Headers disableCache(AsyncHttpServerResponse response) {
+    return response.getHeaders().add("Cache-Control", "no-store");
```
twaddington, line 288: 🎉

twaddington:
I'll likely need to make these changes for R2 as well.

PR #844, opened 2020-12-28
Add "SimplyE" and "Open eBooks" as subprojects

twaddington:
This change adds the two subprojects `simplified-app-simplye` and `simplified-app-openebooks`.

There are several advantages to this change.

- Redundant build configuration has been consolidated into the root `build.gradle` file.
- Dependency versions can be pinned across all projects within `libraries.gradle`.
- App versions will track Core and always be up to date.
- Refactoring across projects becomes simpler and requires only a single pull-request.
- Less app versions for QA to track.

If this change is adopted the GitHub repositories for [SimplyE](https://github.com/NYPL-Simplified/Simplified-Android-SimplyE) and [Open eBooks](https://github.com/NYPL-Simplified/Simplified-Android-Open-eBooks) can be deprecated and removed.

The two projects were migrated using the following script:

https://gist.github.com/twaddington/45be011e6d65f230aa8b26883c91c703


twaddington:
Currently this change will pin all app versions to the version of Core. The nice thing about this is there's only one version to increment whenever we do a release. Also, all apps will have the same version code so it's easy to know what's in which app. However, we can easily go back to the old way and have discrete versions for each published app if we like that better.

twaddington:
Currently the pull-request build is failing because `google-services.json` is missing. We should just check that into our public repository. It's not sensitive information and can be easily extracted from our published .apk files.

io7m:
I'm still on vacation so I've not looked at this yet, but I have a couple of requirements with this:

* Someone with no credentials whatsoever still needs to be able to build the code.
* Reinforcing the above: I'd want to see a complementary CI workflow that builds the code DRM-free with no credentials, so that we're forced to keep the promise that the code will always be buildable DRM-free.

io7m:
Additionally: I'd like to show both the application version and commit, and the core version and commit in the settings screen. The exception to this would be that if both versions and commits are identical, one of those could be hidden in the UI. This would result in NYPL builds (SimplyE and Open eBooks) always showing just one version. LFA builds would continue to show two versions.

io7m:
Yes, I know, stop working and be on vacation. :relaxed:

twaddington:
Hey @io7m thanks for taking a quick look! Was planning on letting this hang out until you were back.

> Someone with no credentials whatsoever still needs to be able to build the code.

This should be possible right now, though we have a lot of different possible configurations to test. I'm going to keep poking at it this week to make sure I'm not missing something.

I also have a follow-on branch that makes a few improvements to the way we're handling credentials.

> LFA builds would continue to show two versions.

Yep! Right now the apps will still show two versions, but NYPL builds will just have the same app version and core version. We can make a small change to hide the core version if it's the same.

> Yes, I know, stop working and be on vacation. ☺️

Go relax!

twaddington:
Making some additional changes that I think will be pretty cool. Gonna close this pull-request and reopen when I'm done.

PR #846, opened 2020-12-30
Rename src/ to docs/ and update README

twaddington:


twaddington:
Ah, I actually see this `src/site/resources` pattern is in use across all the other repos. It seemed a bit weird to me, but at least it's consistent so I'm gonna close this out 😅

PR #849, opened 2021-01-06
Migrate to consolidated secrets.conf file.

twaddington:
**What's this do?**
This updates the projects to use a consolidated `secrets.conf` file instead of individual config files in differing formats.

Make configuring projects and managing secrets easier.

Ensure release builds compile correctly.

Depends on https://github.com/NYPL-Simplified/Certificates/pull/16

Yes

Yes


twaddington:
No rush on merging this one.

simplified-app-simplye/src/main/resources/META-INF/services/org.nypl.simplified.viewer.audiobook.AudioBookFeedbooksServiceType
```diff
@@ -1 +0,0 @@
-org.nypl.simplified.simplye.FeedbooksConfigurationService
```
twaddington, line 1: Looks like this service is unnecessary due to the `MainFeedbooksSecretService` in Core.

Unless I'm misunderstanding something?

io7m: This looks like a vestigial organ. I think the service just got defined twice, but the app only looks at the first one it encounters anyway.

io7m:
I'm fine with doing this.

PR #852, opened 2021-01-20
Intercept calls to non-local URIs

io7m:
**What's this do?**
This adds code to reject any calls made to non-localhost URIs in
Readium 1.

This is desired by LFA, who want to prevent the limited
bandwidth available to classrooms being wasted by accidental redirects
to real servers on the internet from inside EPUB files.

Check that books still work. :smile: 

None.

N/A

@io7m Tested a couple of books in the SimplyE collection.

### DIFF
--- a/simplified-viewer-epub-readium1/src/main/java/org/nypl/simplified/viewer/epub/readium1/ReaderWebViewClient.java
+++ b/simplified-viewer-epub-readium1/src/main/java/org/nypl/simplified/viewer/epub/readium1/ReaderWebViewClient.java
@@ -11,31 +11,33 @@ import com.io7m.jnull.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.io.ByteArrayInputStream;
 import java.io.InputStream;
 import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.HashMap;
+import java.util.Objects;
 
 /**
  * The web client responsible for overriding the requests for certain
  * resources.
  */
 
-final class ReaderWebViewClient extends WebViewClient
-{
+final class ReaderWebViewClient extends WebViewClient {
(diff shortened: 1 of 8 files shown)


io7m:
This is one I would like a few eyes on, because touching anything inside Readium 1 makes me extremely nervous!

simplified-viewer-epub-readium1/src/main/java/org/nypl/simplified/viewer/epub/readium1/ReaderWebViewClient.java
```diff
@@ -101,6 +114,22 @@ public ReaderWebViewClient(
...
+          "text/plain",
+          "UTF-8",
+          403,
+          "FORBIDDEN",
+          new HashMap<>(),
+          new ByteArrayInputStream(new byte[0])
+        );
+      }
+    } catch (URISyntaxException e) {
+      LOG.error("invalid URI: ", e);
```
io7m, line 130: This strikes me as a silly answer to a silly question.

twaddington:
Looks okay to me. Were you able to find a book to test it with?

io7m:
Yeah, most books seem to have links to a publisher web site, and this seems to prevent navigating to them.

io7m:
> this seems to prevent navigating to them.

To be clear... That's a good thing.

io7m:
Need to extend this with a flag that allows for SimplyE to still open links. Ideally in something that isn't the same web view as R1... :sweat:

io7m:
I've gone with the simpler option for now: If a build allows links, then the reader behaves the same way it always did. If they don't allow links... Nothing happens.

In the future (likely for R2 instead), we should open an external web view for URLs like this.

killerpuppytails:
> we should open an external web view for URLs like this.

Absolutely yes to this.

twaddington (approved):

PR #856, opened 2021-01-27
Re-add the ability to show books from all accounts

io7m:
**What's this do?**
At some point, we lost the ability for builds to have Books/Holds views
that show books from all accounts. This change adds a build-time flag
that allows builds to opt-in to showing all books.

LFA use this in their builds.

Check that SimplyE builds have the same behaviour they always did (showing
books from only the relevant account). Vanilla has been changed to show books
from all accounts.

None.

N/A

@io7m checked SimplyE and Vanilla

### DIFF
--- a/simplified-app-vanilla/src/main/java/org/nypl/simplified/vanilla/VanillaBuildConfigurationService.kt
+++ b/simplified-app-vanilla/src/main/java/org/nypl/simplified/vanilla/VanillaBuildConfigurationService.kt
@@ -15,6 +15,8 @@ class VanillaBuildConfigurationService : BuildConfigurationServiceType {
     get() = true
   override val showHoldsTab: Boolean
     get() = true
+  override val showBooksFromAllAccounts: Boolean
+    get() = true
   override val vcsCommit: String
     get() = BuildConfig.GIT_COMMIT
   override val simplifiedVersion: String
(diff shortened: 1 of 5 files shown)


simplified-app-vanilla/src/main/java/org/nypl/simplified/vanilla/VanillaBuildConfigurationService.kt
```diff
@@ -15,6 +15,8 @@ class VanillaBuildConfigurationService : BuildConfigurationServiceType {
     get() = true
   override val showHoldsTab: Boolean
     get() = true
+  override val showBooksFromAllAccounts: Boolean
+    get() = true
```
twaddington, line 19: Can we make this false so it works more like SimplyE?

twaddington (approved):
Looks good

simplified-app-vanilla/src/main/java/org/nypl/simplified/vanilla/VanillaBuildConfigurationService.kt
io7m: The point was to make it _not_ like SimplyE so that you could see the difference. :laughing:

twaddington: Maybe we could make this a debug preference

io7m: Deja vu!

twaddington: Fine as-is though! Let's just get it merged.

PR #857, opened 2021-01-28
SIMPLY-3476: Explicit auth return

io7m:
**What's this do?**
This adjusts the behaviour of the account fragment in that the "closeOnLoginSuccess"
flag now explicitly switches back to the catalog tab. This is a bit of a stopgap
measure to specifically support Open eBooks and should be replaced when we have
a better viewmodel backing the catalog (the viewmodel would subscribe to account
events, and would refresh itself when an account logs in).

https://jira.nypl.org/browse/SIMPLY-3476

1. Open the Open eBooks collection in any app capable of seeing it. The collection (assuming you're not logged in) will redirect you to the settings screen to log in.
2. Log in.
3. After a brief pause (two seconds, currently), observe that you're redirected back to the catalog.

None.

N/A

@io7m Tried the Open eBooks collection in SimplyE and Vanilla.

### DIFF
--- a/simplified-main/src/main/java/org/nypl/simplified/main/MainActivity.kt
+++ b/simplified-main/src/main/java/org/nypl/simplified/main/MainActivity.kt
@@ -521,6 +521,10 @@ class MainActivity :
         override fun openSettingsAccountRegistry() {
           throw UnreachableCodeException()
         }
+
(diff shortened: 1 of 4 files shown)


twaddington (approved):
Looks good 👍

io7m:
Zoooom! :train:

PR #858, opened 2021-01-28
Set the Crashlytics user id parameter.

twaddington:
**What's this do?**
If the current account is authenticated then set the Crashlytics user id
to the md5 hash of the account username.

This allows us to search Crashlytics by username (barcode) when we get
bug reports from specific patrons.

To match the behavior of iOS and enable partners (e.g. Lyrasis) to look up crashes.

n/a

n/a

n/a

### DIFF
--- a/simplified-main/src/main/java/org/nypl/simplified/main/MainFragmentViewModel.kt
+++ b/simplified-main/src/main/java/org/nypl/simplified/main/MainFragmentViewModel.kt
@@ -38,16 +38,16 @@ class MainFragmentViewModel : ViewModel() {
     this.services.requireService(ProfilesControllerType::class.java)
   }
   private val crashlytics by lazy {
-    services.optionalService(CrashlyticsServiceType::class.java)
+    this.services.optionalService(CrashlyticsServiceType::class.java)
   }
 
   init {
     this.disposables.add(
-      profilesController.accountEvents()
+      this.profilesController.accountEvents()
         .subscribe(this::onAccountEvent)
     )
     this.disposables.add(
-      profilesController.profileEvents()
+      this.profilesController.profileEvents()
         .subscribe(this::onProfileEvent)
     )
   }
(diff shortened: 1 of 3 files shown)


simplified-main/src/main/java/org/nypl/simplified/main/MainFragmentViewModel.kt
```diff
@@ -11,13 +29,150 @@ class MainFragmentViewModel : ViewModel() {
...
+  private val disposables = mutableListOf<Disposable>()
+  private val services = Services.serviceDirectory()
+  private val accountProviders by lazy {
+    this.services.requireService(AccountProviderRegistryType::class.java)
+  }
+  private val profilesController by lazy {
+    this.services.requireService(ProfilesControllerType::class.java)
+  }
+  private val crashlytics by lazy {
+    services.optionalService(CrashlyticsServiceType::class.java)
```
io7m, line 41: `this.`

simplified-main/src/main/java/org/nypl/simplified/main/MainFragmentViewModel.kt
```diff
@@ -11,13 +29,150 @@ class MainFragmentViewModel : ViewModel() {
 
   private val logger = LoggerFactory.getLogger(MainFragmentViewModel::class.java)
 
-  /**
-   * `true` if the history of tabs should be cleared.
-   */
+  private val disposables = mutableListOf<Disposable>()
```
io7m, line 32: There's actually a `CompositeDisposable` class in `rxjava` for this. Best to use that instead. It acts like a thread-safe mutable collection of `Disposable` values that is itself a `Disposable` value.

simplified-main/src/main/java/org/nypl/simplified/main/MainFragmentViewModel.kt
```diff
@@ -11,13 +29,150 @@ class MainFragmentViewModel : ViewModel() {
...
+          }
+        }
+      }
+    }
+
+  override fun onCleared() {
+    super.onCleared()
+
+    // Clear any disposables before the view model is destroyed
+    disposables.forEach { it.dispose() }
```
io7m, line 77: With the `CompositeDisposable`, this then just becomes `disposables.dispose()`

twaddington: Nice. That's cleaner.

PR #859, opened 2021-02-04
Clear web cache once readium has initialized.

ray-lee:
**What's this do?**

Clear the WebView cache as soon as Readium has initialized. This removes cached JS files.

This is a follow-on to [SIMPLY-3362](https://jira.nypl.org/browse/SIMPLY-3362)/[#840](https://github.com/NYPL-Simplified/Simplified-Android-Core/pull/840), to add a bit of insurance that we will pass security checks. Apparently it's not possible to prevent WebView from caching JS files altogether, so this removes them as soon as possible. The problem with cached files is that they reveal the port number of the embedded http server. With this knowledge, one could use a browser to get decrypted content directly from the http server.

Open a book in Readium. Browse the filesystem at /data/data/org.nypl.simplified.vanilla/cache/WebView/Default/HTTP Cache/Code Cache. There should be no JS files cached.

n/a

n/a

@ray-lee ran the vanilla app.


io7m:
Looks good to me! Hard to argue with a change like this. :laughing:

PR #860, opened 2021-02-16
Redo Crashlytics control

io7m:
**What's this do?**
This moves the Crashlytics handling from the main fragment view model
into the profile controller. This prevents a couple of crashes that
showed up in LFA builds. This also adds a small section to the debug
setting UI that shows the user their current Crashlytics ID.

LFA builds were crashing due to a `lazy` race condition. Looking at it, I realized that the Crashlytics handling didn't need to be in a view model at all.

Check Crashlytics. :smile:

None.

N/A

@io7m Tested on multiple accounts.


MalcolmMcFly (approved):
Looks good

PR #861, opened 2021-02-18
SIMPLY-1506: Accessibility events

io7m:
**What's this do?**
This defines a new accessibility service that announces book status changes to screen readers.

Specifically it does a number of things:

* Upgrades book status events to contain the current and previous status of any given book.
* Uses the new status information to drastically simplify the implementation of the current Announcements code as it no longer needs to maintain its own cache of book statuses. There was no real way to avoid doing this; there were compilation errors otherwise. The new announcements code passes all of the tests that the old code did.
* Publishes Toast messages when book statuses change in specific ways as listed in the referenced ticket. Toasts are only published if a screen reader is active.
* Adds a small debug switch that allows for _always_ publishing Toast messages even if a screen reader is disabled.

https://jira.nypl.org/browse/SIMPLY-1506

Check that all of the requested announcements in `SIMPLY-1506` actually happen.

None.

N/A

@io7m Tested downloads and returns, but couldn't test a couple of the Hold-related cases.


simplified-accessibility/src/main/java/org/nypl/simplified/accessibility/AccessibilityServiceType.kt
```diff
@@ -0,0 +1,35 @@
...
+import androidx.lifecycle.Lifecycle
+import androidx.lifecycle.LifecycleObserver
+import androidx.lifecycle.LifecycleOwner
+import androidx.lifecycle.OnLifecycleEvent
+
+/**
+ * The interface exposed by accessibility services.
+ */
+
+interface AccessibilityServiceType : LifecycleObserver {
```
twaddington, line 12: Nice to see LifecycleObserver getting some love here

simplified-accessibility/src/main/java/org/nypl/simplified/accessibility/AccessibilityToasts.kt
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * The default implementation of the [AccessibilityToastsType] interface that just displays
+ * a toast using the given context.
+ */
+
+class AccessibilityToasts(
+  private val context: Context
+) : AccessibilityToastsType {
+  override fun show(message: String) {
+    Toast.makeText(this.context, message, Toast.LENGTH_LONG).show()
```
twaddington, line 15: You might see some glitchy behavior here when multiple messages are shown in rapid succession.

Best practice is to retain the Toast object that's created and update it instead of creating a new one for each message. You'll want to do something like this:

```kotlin
  this.toast = this.toast ?: Toast(context).apply { setDuration(Toast.LENGTH_LONG) }
  this.toast.setText(message)
  if (!this.toast.view.isShown()) {
    this.toast.show()
  }
```

twaddington:
Looks good to me! Had one piece of feedback.

simplified-accessibility/src/main/java/org/nypl/simplified/accessibility/AccessibilityToasts.kt
io7m: Is this just a visual issue, or would it affect the screen reader too? Remember the sole purpose of these is _not_ to be seen. :laughing:

twaddington: Well, if you don't want this to actually be visible then I suggest sending an accessibility event to announce instead.

You'll need a view reference to do this though.

```kotlin
view.announceForAccessibility(message)
```

> Source: https://developer.android.com/reference/android/view/View#announceForAccessibility(java.lang.CharSequence)

This is just shorthand for something like:

```kotlin
val manager = context.getSystemService(Context.ACCESSIBILITY_SERVICE) as AccessibilityManager

if (manager.isEnabled()) {
    AccessibilityEvent e = AccessibilityEvent.obtain().apply {
        setEventType(AccessibilityEvent.TYPE_ANNOUNCEMENT)
        setClassName(context.getClass().getName())
        setPackageName(context.getPackageName())
        getText().add(message)
    }
    manager.sendAccessibilityEvent(e)
}
```

I haven't tested this myself so you'll want to verify it works properly.

io7m: :weary: 

I sat there for about half an hour looking for a way to explicitly "announce" messages to the screenreader. _Of course_ it's defined on `View`. Of course.

Won't use toasts at all for this now.

twaddington: Hah! Yeah, it's not super well documented. I think the accessibility team at Google was only a couple people at one point. I only know cause I've had to do this once before :P

simplified-accessibility/src/main/java/org/nypl/simplified/accessibility/AccessibilityService.kt
```diff
@@ -0,0 +1,159 @@
...
+  @OnLifecycleEvent(Lifecycle.Event.ON_RESUME)
+  override fun onViewAvailable(owner: LifecycleOwner) {
+    this.lifecycleOwner = owner
+    this.subscriptions.add(this.bookRegistry.bookEvents().subscribe(this::onBookEvent))
+  }
+
+  @OnLifecycleEvent(Lifecycle.Event.ON_PAUSE)
+  override fun onViewUnavailable(owner: LifecycleOwner) {
+    this.lifecycleOwner = owner
+    this.subscriptions.dispose()
```
io7m, line 71: For future reference: This is a bug.

Why? Because `dispose()` will mark `subscriptions` as disposed, and any subsequent calls to `add()` will silently be ignored.

io7m:
Now publishing events directly without any toasts and without a lifecycle bug. :sweat_smile:

simplified-accessibility/src/main/java/org/nypl/simplified/accessibility/AccessibilityService.kt
twaddington: Aha, good catch!

PR #862, opened 2021-02-19
Add simplified-drm-core to the monorepo

twaddington:
This change pulls DRM-Android-Core into the monorepo as a subproject named simplified-drm-core.

Since DRM-Android-Core is public, and available on Maven Central, it made sense to pull this into the core monorepo. This allows us to keep build configuration and dependency versions in lockstep.

It's declared as a dependency by 19 other subprojects in the monorepo so it's already a key dependency.


twaddington:
This is failing because I need to also update the deps in Simplified-R2-Android and publish a new version of that. I might consider pulling that into core as well but we can chat about this tomorrow.

PR #864, opened 2021-03-02
Ensure correct adobe credentials are associated with the current account.

ray-lee:
**What's this do?**

When Adobe credentials are activated, the Adobe DRM connector can return a list of multiple credentials, including the newly activated credentials, and any that were previously activated for other accounts. This attempts to ensure that the current account is associated with the correct credentials, by finding the last credentials in the last that are not associated with any other account.

This fixes [SIMPLY-2979](https://jira.nypl.org/browse/SIMPLY-2979). Previously, if activation returned a list of credentials, the first in the list was assumed to be the newly activated credentials for the current account. If other accounts had previously been activated, the first in the list actually belonged to the first account that was activated, not the current account being activated.

Unit tests are included. See the [first comment in SIMPLY-2979](https://jira.nypl.org/browse/SIMPLY-2979?focusedCommentId=52276&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-52276) for testing steps. 

n/a

n/a

@ray-lee ran this in a slightly older commit of the SimplyE app. I can't run it in the latest commit of SimplyE because of [SIMPLY-3529](https://jira.nypl.org/browse/SIMPLY-3529), but I don't think there's any reason to believe it wouldn't work.

### DIFF
--- a/simplified-books-controller/src/main/java/org/nypl/simplified/books/controller/ProfileAccountLoginTask.kt
+++ b/simplified-books-controller/src/main/java/org/nypl/simplified/books/controller/ProfileAccountLoginTask.kt
@@ -391,8 +391,21 @@ class ProfileAccountLoginTask(
         "Must have returned at least one activation"
       )
 
+      /*
+       * Find the newly activated credentials (the one whose user id is not associated with any
+       * account). There should only be one, and it should be the last in the list, but this check
+       * makes sure.
+       */
+
+      val newPostCredentials = postCredentials.last { credentials ->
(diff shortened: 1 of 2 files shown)


io7m:
:star_struck: 

I will review this properly later today. Impressive if this is what the problem has been for all these years!

io7m:
Looks good to me!

Apologies for the number of dependencies that have to be injected in these tests. They are not the most pleasant tests to write.

PR #865, opened 2021-03-02
SIMPLY-3574: Fix data being incorrectly set in error reporting intent.

ray-lee:
**What's this do?**

Fix log files not being correctly set in the intent when reporting errors.

This fixes [SIMPLY-3574](https://jira.nypl.org/browse/SIMPLY-3574).

See description of [SIMPLY-3574](https://jira.nypl.org/browse/SIMPLY-3574).

n/a

n/a

@ray-lee ran the vanilla app.

### DIFF
--- a/simplified-reports/src/main/java/org/nypl/simplified/reports/Reports.kt
+++ b/simplified-reports/src/main/java/org/nypl/simplified/reports/Reports.kt
@@ -102,7 +102,7 @@ object Reports {
       val compressedFiles =
         files.mapNotNull(this::compressFile)
       val contentUris =
-        compressedFiles.map { file -> this.mapFileToContentURI(context, file) }
+        compressedFiles.toSet().map { file -> this.mapFileToContentURI(context, file) }
 
       this.logger.debug("attaching {} files", compressedFiles.size)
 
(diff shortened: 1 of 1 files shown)


simplified-reports/src/main/java/org/nypl/simplified/reports/Reports.kt
```diff
@@ -102,7 +102,7 @@ object Reports {
       val compressedFiles =
         files.mapNotNull(this::compressFile)
       val contentUris =
-        compressedFiles.map { file -> this.mapFileToContentURI(context, file) }
+        compressedFiles.toSet().map { file -> this.mapFileToContentURI(context, file) }
```
ray-lee, line 106: `compressedFiles` can contain duplicates. For example, if `log.txt` is compressed to `log.txt.gz`, and `log.txt.gz` already existed from a previous error report, `log.txt.gz` will appear twice in the list. The `toSet()` call removes any duplicates.

simplified-reports/src/main/java/org/nypl/simplified/reports/Reports.kt
```diff
@@ -112,7 +112,7 @@ object Reports {
           this.putExtra(Intent.EXTRA_EMAIL, arrayOf(address))
           this.putExtra(Intent.EXTRA_SUBJECT, subject)
           this.putExtra(Intent.EXTRA_TEXT, body)
-          this.putExtra(Intent.EXTRA_STREAM, arrayListOf(contentUris))
+          this.putExtra(Intent.EXTRA_STREAM, ArrayList(contentUris))
```
ray-lee, line 115: Changing to `ArrayList()` instead of `arrayListOf()`. Since `contentUris` is a list, `arrayListOf(contentUris)` created a list of lists instead of a list of Uris.

io7m:
Nice. Even more important now that we're about to introduce a major new feature...

PR #866, opened 2021-03-04
SIMPLY-3581: Populate error report emails with error details.

ray-lee:
**What's this do?**

When sending error reports, populate the body with the error details as displayed on the Error Details screen:

This fixes [SIMPLY-3581](https://jira.nypl.org/browse/SIMPLY-3581). This makes error reports more informative, and makes it easier to see what the user is seeing without having to dig through logs.

There are unit tests for rendering the error details as strings.

To test:

- Do something that results in an error, e.g. log in to a library with an incorrect password.
- Tap the Error Details button.
- Tap the Send To Support button.
- Select Share with Gmail, Just once.

An email should be composed, containing the same details (attributes and steps) as were displayed on the Error Details screen.

#865 should be merged first, because the incorrect intent that it fixes can cause gmail to get confused.

n/a

@ray-lee ran the vanilla app.

### DIFF
--- a/simplified-tests/src/test/java/org/nypl/simplified/tests/errorpage/ErrorPageParametersContract.kt
+++ b/simplified-tests/src/test/java/org/nypl/simplified/tests/errorpage/ErrorPageParametersContract.kt
@@ -1,0 +1,244 @@
+package org.nypl.simplified.tests.errorpage
+
+import org.junit.Assert.assertEquals
+import org.junit.Test
+import org.nypl.simplified.taskrecorder.api.TaskStep
+import org.nypl.simplified.taskrecorder.api.TaskStepResolution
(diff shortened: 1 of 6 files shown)


simplified-tests/src/test/java/org/nypl/simplified/tests/errorpage/ErrorPageParametersContract.kt
```diff
@@ -0,0 +1,244 @@
...
+    assertEquals(
+      """
+      There was an error!
+
+      Account:
+      12345
+
+      Steps:
+      1. Doing something...
+      ❌ Nooooooo!
```
io7m, line 239: :smile:

io7m (approved):
Always good when a PR is more unit test than feature! :smile:

PR #869, opened 2021-03-19
SIMPLY-3605: Simplified Bookmarks

io7m:
**What's this do?**

This code uses the latest Simplified-R2-Android snapshot which
switches to using href values for locators rather than integer
chapter indices. In order to reliably implement the new bookmarks
spec [0], this required some changes to how we serialize bookmarks
internally. The BookLocation class has been split into a modern
BookLocationR2 class, and a deprecated BookLocationR1 class. We now
properly version any serialized bookmarks on disk, and there are
tests that demonstrate the app can still understand bookmarks as old
as the 3.* branch formats.

I took the opportunity to use shorter timeout values for HTTP
requests in the test suite, because this was making the test suite
run slowly for the bookmark service tests.

This implements the remainder of the Simplified Bookmarks spec,
including unit tests to demonstrate that the expected test samples
are parsed the correct way.

The bookmarks spec has been imported as a submodule so that the
various included test cases can be used as test suite resources
directly.

We now enforce that all bookmarks must be specified in a UTC
time zone (although they can be displayed to the user in any
time zone).

[0] https://github.com/NYPL-Simplified/Simplified-Bookmarks-Spec

https://jira.nypl.org/browse/SIMPLY-3605
https://jira.nypl.org/browse/SIMPLY-3604

This encapsulates some fairly complex behaviour. Right now, bookmark sharing for Android and iOS for R1 bookmarks will still not work; the two platforms still serialize different subsets of the same data. Bookmark sharing for Android and iOS for R2 _will_ work after the equivalent support is added on iOS.

I think testing this should include:

* Ensure all existing R1 bookmarks continue to work as needed.
* Test that syncing R1 bookmarks to the server continues to work as it did before (might need two Android devices for this).
* Test that R2 bookmarks work in general; bookmarks should survive app restarts and so on.
* Test that syncing R2 bookmarks to the server works (bookmarks created by a previous R2 version are no longer compatible, but no patrons have ever had access to those).

None.

N/A

@io7m tested R1 bookmarks, including bookmarks that came from the server, and even bookmarks that were written by ancient versions of the app (nothing earlier than 3.*). I also tested R2 bookmarks without apparent issue.


io7m:
Seems we have some test failures that I don't see locally. Investigating...

io7m:
The issues were timezone related and were highlighting real bugs. Made some changes to ensure those bugs can't happen again. As a happy side effect, our GitHub Actions now upload archives of the junit test reports. :smile:

MalcolmMcFly (approved):

PR #870, opened 2021-03-23
Implement support for AxisNow DRM

qnga:
**What's this do?**
- Implementing support for AxisNow DRM. This support can be enabled by putting the module `org.librarysimplified.drm.axis:org.librarysimplified.drm.axis.provider` on the classpath. 

- Adapting the UI to show an indeterminate progress bar during fulfillment when progression information is missing.

This is done in order to support the AxisNow DRM, especially in OpenEbooks. See [SIMPLY-3602](https://jira.nypl.org/browse/SIMPLY-3602) and [SIMPLY-3601](https://jira.nypl.org/browse/SIMPLY-3601).

The `BorrowAxisNow` task has an associated test. The branch `feature/axis-test` must be used to test AxisNow support (only the QA circulation manager offers books protected with AxisNow). R2 experimental reader must be enabled for reading.

Must be merged along with https://github.com/NYPL-Simplified/DRM-Android-Core/pull/6 and https://github.com/NYPL-Simplified/DRM-Android-Adobe/pull/9 .

No.

Yes, I did.

**Shortcomings we might be interested to address or not**
- Dowloading big books is not very fast because all resources have to be downloaded. The process is currently sequential, making it parallel might help a bit.
- Seeing from the public code, fulfilment is a unique black-boxed step. This is meant to ensure no detail about the DRM is leaked. Actually, the stored data and exception wrapping still leak some insights about the DRM scheme. This balance might be reconsidered with real security and privacy requirements in mind.
- Error handling is quite basic. For instance, if Axis doesn't allow a patron to download some book because it has been downloaded too many times (usually more than 6 times), we get a `BadRequest` error while attempting to fetch a license. The returned content is currently not parsed, so the error message is something like "cannot download license".
- Byte range requests are not really supported: the whole resource is decrypted every time. Because audio and video resources are deflated, there's no easy way to deal with this.


io7m:
I've updated this PR to point to the latest DRM core/adobe releases so that the CI gets a chance to run.

io7m:
This all looks good to me! We can deal with the points you raised as they come (assuming they do come up in practice :smile:)

PR #872, opened 2021-03-30
SIMPLY-3604: Bookmarks (again)

io7m:
**What's this do?**
This updates to the latest bookmarks spec with the changes @ettore discovered.

https://jira.nypl.org/browse/SIMPLY-3604

The same tests that were applicable to https://github.com/NYPL-Simplified/Simplified-Android-Core/pull/869 are applicable to this. The changes essentially enforce things that we were already doing implicitly, and change the format of the R2 bookmarks. There won't be any R2 bookmarks in the wild yet, because R2 hasn't been officially released.

None.

See the [spec](https://github.com/NYPL-Simplified/Simplified-Bookmarks-Spec).

@io7m ran the test suite and tried a few books.

### DIFF
--- a/simplified-books-api/src/main/java/org/nypl/simplified/books/api/BookmarkKind.kt
+++ b/simplified-books-api/src/main/java/org/nypl/simplified/books/api/BookmarkKind.kt
@@ -31,7 +31,7 @@ sealed class BookmarkKind(
    */
 
   object ReaderBookmarkExplicit :
-    BookmarkKind("https://www.w3.org/ns/oa#bookmarking") {
+    BookmarkKind("http://www.w3.org/ns/oa#bookmarking") {
     override fun toString(): String {
       return "ReaderBookmarkExplicit"
     }
@@ -40,7 +40,7 @@ sealed class BookmarkKind(
   companion object {
 
     val obsoleteBookmarkingURI =
-      "http://www.w3.org/ns/oa#bookmarking"
+      "https://www.w3.org/ns/oa#bookmarking"
 
     fun ofMotivation(motivationURI: String): BookmarkKind {
       return when (motivationURI) {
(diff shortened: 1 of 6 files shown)


simplified-books-api/src/main/java/org/nypl/simplified/books/api/BookmarkKind.kt
```diff
@@ -40,7 +40,7 @@ sealed class BookmarkKind(
   companion object {
 
     val obsoleteBookmarkingURI =
-      "http://www.w3.org/ns/oa#bookmarking"
+      "https://www.w3.org/ns/oa#bookmarking"
```
ettore, line 43: At first I thought having `https` here was a mistake but I see the `obsolete` prefix now, so all good. :)

simplified-books-api/src/main/java/org/nypl/simplified/books/api/BookmarkKind.kt
```diff
@@ -31,7 +31,7 @@ sealed class BookmarkKind(
    */
 
   object ReaderBookmarkExplicit :
-    BookmarkKind("https://www.w3.org/ns/oa#bookmarking") {
+    BookmarkKind("http://www.w3.org/ns/oa#bookmarking") {
```
ettore, line 34: wondering if it might be better to have a constant for this fixed value somewhere, unless.... this is actually it (did I mention I don't know kotlin?)

ettore (approved):
looks good to me!

simplified-books-api/src/main/java/org/nypl/simplified/books/api/BookmarkKind.kt
io7m: That's actually the constant there.

PR #876, opened 2021-04-09
Added step to post to slack after successful build

Dumblydore:
**What's this do?**
Fastlane will post to slack after app is built and deployed to firebase

To make it easier for QA to know when a new Android build is available

After merge see if a slack post is made

None

I tried it on a test slack channel but more tests would be nice!

### DIFF
--- a/.ci-local/credentials.sh
+++ b/.ci-local/credentials.sh
@@ -71,7 +71,7 @@ EOF
 
 #------------------------------------------------------------------------
 # Addding slack webhook to environment
-SLACK_WEBHOOK_URL=$(</Users/medwar40/Workspace/NYPL/Certificates/SimplyE/slack-webhook.url) || 
+SLACK_WEBHOOK_URL=$(<.ci/credentials/SimplyE/slack-webhook.url) || 
   fatal "Slack Webhook url not found."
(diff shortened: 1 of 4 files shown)


.ci-local/credentials.sh
```diff
@@ -68,3 +68,11 @@ org.librarysimplified.nexus.password=${NYPL_NEXUS_PASSWORD}
 org.librarysimplified.app.assets.openebooks=${OPENEBOOKS_CREDENTIALS}
 org.librarysimplified.app.assets.simplye=${SIMPLYE_CREDENTIALS}
 EOF
+
+#------------------------------------------------------------------------
+# Addding slack webhook to environment
+SLACK_WEBHOOK_URL=$(</Users/medwar40/Workspace/NYPL/Certificates/SimplyE/slack-webhook.url) ||
```
io7m, line 74: Fairly sure this line isn't going to work on CI!

The right path would be `.ci/credentials/SimplyE/slack-webhook.url`.

Dumblydore: Ooops, nice catch!

io7m (approved):
Let's go!

PR #877, opened 2021-04-12
Refactor/rewrite MainActivity, MainFragment and the splash module

qnga:
**What's this do?**
- Refactor/rewrite `MainActivity`, `MainFragment` and the splash module
- Move the onboarding to a dedicated module
- Fix some confusion in the navigation stuff (there was a mismatch in the meaning of some methods in the different implementations of `NavigationControllerType`)

The code I rewrote was confused, not maintainable and buggy. Many crashes were possible at startup, see [SIMPLY-3473](https://jira.nypl.org/browse/SIMPLY-3473). This PR introduces a new architecture for UI related code that prevents crashes on orientation changes.

Launch SimplyE and VanillaWithProfiles, possibly with `simplified-migration-fake` on the classpath. Switch your device orientation on any screen.

As it's going to be moved, I didn't touch the age gate. It's probably still buggy.

Yes, I think so.

Yes, I did.


io7m:
This all looks good to me, and I'm happy to see that awful splash code going away. I need a bit more time to clone and run this though, so I can get an understanding of how the code executes now.

io7m (approved):
Let's go! I'm looking forward to the number of crashes being reduced. :smile:

PR #878, opened 2021-04-13
SIMPLY-3582: Introduce a noise reduction policy

io7m:
**What's this do?**
This moves logging configuration into its own class, and reduces the
logging levels of some of the noisier parts of the app. This appears
to improve the UI performance of the app somewhat, but I'm willing
to accept that it might be a placebo effect.

https://jira.nypl.org/browse/SIMPLY-3582

Nothing to test really. The app logs less to logcat than it used to.

None.

N/A

@io7m ran it a lot.


io7m:
I'll sort out the conflict shortly. Was introduced after merging the last PR.

MalcolmMcFly (approved):
Looks good.

PR #880, opened 2021-04-16
Changed age gate to only appear for the SimplyE catalog

Dumblydore:
**What's this do?**
Changed age gate to only appear for the SimplyE catalog.
Also Fixed crash when user would click done before selecting birth year

https://jira.nypl.org/browse/SIMPLY-3690

Verify that the age gate dialog only appears when user selects "Add Library Later" during onboarding

None

Yes


MalcolmMcFly (approved):
This looks good. Seems to address the robo test crash

qnga:
I believed we wanted to remove the neutral age gate and go back to what there was before. To achieve this, the age gate should be completely moved from the MainActivity or onboarding to the catalog. Otherwise, if a patron selects "Find my library" on the onboarding and then switches to the SimplyE collection, he'll get the old age gate.

I also believed RxJava had been preferred to Kotlin channels and other coroutine stuff in SimplyE. If you really want to use Kotlin channels, be careful, events might be received between the stopped and destroyed states, leading to a crash if you try to modify the Fragment's state after it has been saved. In other words, I think you should explicitly cancel the observing coroutine in `onStop`.
You can see an example at https://github.com/readium/r2-testapp-kotlin/blob/develop/r2-testapp/src/main/java/org/readium/r2/testapp/utils/EventChannel.kt

killerpuppytails:
> I believed we wanted to remove the neutral age gate and go back to what there was before. To achieve this, the age gate should be completely moved from the MainActivity or onboarding to the catalog. Otherwise, if a patron selects "Find my library" on the onboarding and then switches to the SimplyE collection, he'll get the old age gate.

Sorry, I was not clear. I did not want to remove the neutral age gate, as Google will not approve an app with the old version. I wanted the neutral age gate to only appear on the library called SimplyE Collection.

qnga:
Sorry, I may have misunderstood what "neutral" referred to and expressed myself poorly. I understood we need to keep the new style age gate, which doesn't explicitly state the age boundary, but it should protect only the SimplyE collection, as it did before. If this is true, I keep believing both `MainActivity` and onboarding are wrong places to put the age gate in.

killerpuppytails:
> Sorry, I may have misunderstood what "neutral" referred to and expressed myself poorly. I understood we need to keep the new style age gate, which doesn't explicitly state the age boundary, but it should protect only the SimplyE collection, as it did before. If this is true, I keep believing both `MainActivity` and onboarding are wrong places to put the age gate in.

Yes then you are absolutely correct :)

qnga:
Great, we'll get there! To be specific, the age gate which is at the moment shown in `MainActivity` (in `develop`) should be shown in `CatalogFeedFragment` instead of the old one. Nothing related to the age gate should be left both in `MainActivity` and the onboarding. This way, the same age gate will be shown whatever the path followed to open the SimplyE collection for the first time.

simplified-main/src/main/java/org/nypl/simplified/main/MainActivity.kt
```diff
@@ -236,29 +228,6 @@ class MainActivity :
...
-        this.synthesizeDateOfBirthDescription(description, 14)
-      }
-    } else {
-      this.profilesController.profileUpdate { description ->
-        this.synthesizeDateOfBirthDescription(description, 0)
-      }
-    }
-  }
-
   private fun synthesizeDateOfBirthDescription(
```
qnga, line 231: I think the following two methods are not used any more.

simplified-profiles-api/src/main/java/org/nypl/simplified/profiles/api/ProfilePreferences.kt
```diff
@@ -43,5 +43,8 @@ data class ProfilePreferences(
 
   /** @return `true` if the debug settings should be visible. */
 
-  val showDebugSettings: Boolean = false
+  val showDebugSettings: Boolean = false,
+
+  /** @return `true` if user selects SimplyE collection. */
+  val requiresAge: Boolean = false,
```
qnga, line 49: I'm not sure this flag is useful.

`(authentication is AccountProviderAuthenticationDescription.COPPAAgeGate ||
        buildConfig.showAgeGateUi) && preferences.dateOfBirth == null` is a good criterion, isn't it?

qnga (commented):
Looks better! I've left two comments. Also, there are plenty of crashes, as there were before. Not sure it is worth (if ever possible) to try to find a working solution within the `runOnUiThread` paradigm though. I could fix this at the same time as the rest of the module, if you like.

simplified-main/src/main/java/org/nypl/simplified/main/MainActivity.kt
Dumblydore: Nice catch, removed them!

qnga:
I tested it again and it looks like there are some issues:
- the age gate is shown before showing NYPL's catalog
- the age is not restored upon configuration changes
- even if one selected it again, clicking the `Done` button randomly works

simplified-profiles-api/src/main/java/org/nypl/simplified/profiles/api/ProfilePreferences.kt
qnga: Sorry, I made a stupid copy/paste from your code. The right criterion is likely to be `authentication is AccountProviderAuthenticationDescription.COPPAAgeGate` **&&** `buildConfig.showAgeGateUi && preferences.dateOfBirth == null`. That should fix your issue.

qnga (approved):

PR #881, opened 2021-04-19
Refactor and fix simplified-ui-accounts, simplified-ui-settings and simplified-ui-profiles

qnga:
**What's this do?**
- Introduce ViewModels where there were missing.
- Post background events to the UI thread in ViewModels.
- Don't use `runOnUIThread` any more in Fragments to ensure there's no crash on configuration changes.
- Indicate whether the build supports AxisNow in the debug settings.

To ensure that configuration changes don't cause a crash.

This can be tested trying every screen from the relevant modules.

N/A

N/A

Yes, I did.


simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/AccountFragment.kt
```diff
@@ -894,29 +841,22 @@ class AccountFragment : Fragment() {
           view,
           object : com.squareup.picasso.Callback {
             override fun onSuccess() {
-              this@AccountFragment.uiThread.runOnUIThread {
-                onSuccess.invoke()
-              }
+              onSuccess.invoke()
```
io7m, line 844: Are the Picasso callbacks guaranteed to be called on the UI thread? It's been a very long time, I'm unsure what guarantees they provide.

simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/AccountListRegistryViewModel.kt
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  private fun onAccountEvent(event: AccountEvent) {
+    this.accountCreationEvents.onNext(event)
+  }
+
+  override fun onCleared() {
+    super.onCleared()
+    this.subscriptions.clear()
+    this.backgroundExecutor.shutdown()
```
io7m, line 60: :+1: 

So many people forget to shut down exectors. They extend `AutoCloseable` in modern Java, but ... you know... Android...

simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/AccountFragment.kt
qnga: I couldn't find any guarantee in the doc. Looking at the code, it is so though. Everything happens on the same thread on completion: the view image is set and then the callback is called. I think it's unlikely to change in the future.

simplified-ui-settings/src/main/java/org/nypl/simplified/ui/settings/SettingsFragmentDebug.kt
```diff
@@ -166,7 +86,33 @@ class SettingsFragmentDebug : Fragment() {
...
       view.findViewById(R.id.settingsVersionCrashlyticsID)
 
-    return view
+    this.drmTable.addView(
+      this.createDrmSupportRow("Adobe Acs", this.viewModel.adeptSupported)
+    )
+    this.drmTable.addView(
+      this.createDrmSupportRow("AxisNow", this.viewModel.axisNowSupported)
+    )
+
```
io7m, line 95: Nice to see this! This has been something I've been meaning to do for a while. We have quite a few optional DRM components that could be added here, but they don't need to go in right now.

simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/AccountFragment.kt
io7m: Yeah, I doubt they would change this.

io7m:
I like it!

io7m (approved):

qnga:
Thanks!

PR #883, opened 2021-04-22
SIMPLY-3696: New APIs for Library Finder

io7m:
**What's this do?**
This replaces the existing library registry parser with a real OPDS
2.0 parser. This also necessitated upgrading to the latest AudioBook
API to avoid a dependency conflict. It then adds a basic API for querying
the library registry. 

Currently, the API merely calls the existing `/libraries` endpoint, and annotates
each entry with fake location and distance data. This will be replaced
with a call to a different endpoint when the server work is completed.

https://jira.nypl.org/browse/SMA-7

Test that the library registry currently behaves identically to the way it used to behave.

None.

The module list was updated.

@io7m Has been trying it.


simplified-opds2-r2/src/main/java/org/nypl/simplified/opds2/r2/internal/OPDS2ParserR2.kt
```diff
@@ -0,0 +1,51 @@
...
+  override fun parse(): ParseResult<OPDS2Feed> {
+    return try {
+      val data = OPDS2Parser.parse(stream.readBytes(), uri.toURL())
+      ParseResult.Failure(
+        warnings = listOf(),
+        errors = listOf(
+          ParseError(
+            source = this.uri,
+            message = "Not implemented yet!",
+            line = 0,
```
io7m, line 26: I've left this parser in place in order to keep our options open with regards to OPDS 2.0 parsers. I'd like to finish this at a slightly later date and do a competitive analysis between the R2 and Irradia parsers. Right now, we can't use R2 for the accounts registry as it doesn't support the custom roles and other things the accounts registry feeds use.

simplified-tests/src/test/java/org/nypl/simplified/tests/opds2/OPDS2ParserContract.kt
```diff
@@ -0,0 +1,53 @@
...
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import org.nypl.simplified.opds2.parser.api.OPDS2ParsersType
+import org.nypl.simplified.parser.api.ParseResult
+import org.slf4j.Logger
+import java.io.FileNotFoundException
+import java.io.InputStream
+import java.net.URI
+
+abstract class OPDS2ParserContract {
```
qnga, line 13: This is interesting. I used to use parametrized tests in such cases, is there any specific rationale behind your choice?

qnga (approved):
Looks fine!

simplified-tests/src/test/java/org/nypl/simplified/tests/opds2/OPDS2ParserContract.kt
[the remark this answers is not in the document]
io7m: Hah, didn't think of that! We only just moved from junit 4 to junit 5, and I guess I was still thinking in terms of junit 4 (without extensions)... JUnit 5 has an API in the core for injecting different values into tests.

PR #887, opened 2021-04-29
SMA-90: Fixed Age Gate Crash

Dumblydore:
**What's this do?**
Added check to see if AgeGateDialog is visible before attempting to dismiss

https://jira.nypl.org/browse/SMA-90

Repeat steps from jira ticket

No

No documentation changes needed

I did


qnga:
The code is a bit confusing actually. I think `onAgeUpdateSuccess` should be called only when... age has been updated. I guess it is not the case. You could check in `onProfileEvent` that the old age and the new age differ. Just be sure that age change is the only reason to reload the feed.

Dumblydore:
I agree, i added a check and changed the code a little bit to be more readable.

qnga:
Looks good! I would just like @io7m to confirm there is no reason to reload the feed if the profile is updated, other than age changes.

io7m:
No reason that I'm aware of. :+1:

qnga (approved):

PR #890, opened 2021-05-04
SMA-24: Allow for metrics in library finder

Dumblydore:
**What's this do?**
Implements metric events when adding/removing libaries

https://jira.nypl.org/browse/SMA-24

Verify that events displayed in the Firebase console

None

Added documentation for new modules

I did


build.gradle
```diff
@@ -15,7 +15,7 @@ buildscript {
 
   dependencies {
     classpath 'com.android.tools.build:gradle:4.1.1'
-    classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
+    classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:1.4.32"
```
io7m, line 18: This looks like an IDE-inflicted mistake. :slightly_smiling_face:

simplified-books-controller/src/main/java/org/nypl/simplified/books/controller/ProfileAccountCreateTask.kt
```diff
@@ -35,6 +38,7 @@ class ProfileAccountCreateTask(
       this.logger.debug("creating account for provider {}", this.accountProviderID)
       val accountProvider = this.resolveAccountProvider()
       val account = this.createAccount(accountProvider)
+      metrics?.logMetric(MetricEvent.LibraryAdded(account.id.uuid.toString()))
```
io7m, line 41: Do we have a spec somewhere of what the various metric events are? I'm guessing what you probably want here is actually `account.provider.id...`. The provider ID is a publically known ID that tells us which account was created (NYPL, SimplyE Collection, etc). The `account.id` is a randomly generated unique ID and so likely has no meaning in terms of metrics.

simplified-books-controller/src/main/java/org/nypl/simplified/books/controller/ProfileAccountDeleteTask.kt
```diff
@@ -51,6 +54,7 @@ class ProfileAccountDeleteTask(
       val profile = this.profiles.currentProfileUnsafe()
       val account = profile.deleteAccountByProvider(this.accountProviderID)
 
+      this.metrics?.logMetric(MetricEvent.LibraryRemoved(account.uuid.toString()))
```
io7m, line 57: Same as for `LibraryAdded`.

simplified-main/src/main/java/org/nypl/simplified/main/MainServices.kt
```diff
@@ -943,6 +945,12 @@ internal object MainServices {
       serviceConstructor = { return@addService AudioBookManifests }
     )
 
+    addService(
+      message = strings.bootingGeneral("metrics service"),
+      interfaceType = MetricServiceType::class.java,
+      serviceConstructor = { MetricService(context) }
+    )
+
```
io7m, line 953: This makes the metrics service mandatory for all builds. We should really be looking up an implementation via `ServiceLoader` and adding it optionally.

simplified-metrics/README.md
```diff
@@ -0,0 +1,4 @@
+org.librarysimplified.metrics
+=============================
+
+The `org.librarysimplified.metrics` module provides a Frirebase impl.
```
io7m, line 4: Typo. :yum:

simplified-metrics/src/main/java/resources/META-INF/services/org.nypl.simplified.crashlytics.api.CrashlyticsServiceType
```diff
@@ -0,0 +1 @@
+org.nypl.simplified.metrics.api.MetricServiceType
```
io7m, line 1: This file isn't right. The name of the file must be the name of the interface you're passing to `ServiceLoader`, and the line(s) within the file name implementations of that interface. What you're saying here is that you want to call: `ServiceLoader.load(CrashlyticsServiceType.class)` and you want to system to call `new MetricServiceType()` as a result. Probably not what you intended. :smile: 

The file should be called `org.nypl.simplified.metrics.api.MetricServiceType`, and it should contain one like: `org.nypl.simplified.metrics.MetricService`.

There's a further problem here in that the `MetricService` implementation requires an argument to the constructor, and `ServiceLoader` can't handle that. You need to expose a separate factory.

Let me know if you need a hand with this.

io7m (changes requested):
Mostly good, but needs some adjustments. :+1:

README.md
```diff
@@ -85,7 +85,7 @@ support the use of any other IDE at the moment.
 
 #### JDK
 
-Install a reasonably modern JDK: Java 11 is the current long-term support (LTS) release of Java. We
+Install a reasonably modern JDK: Java 11 is the curt long-term support (LTS) release of Java. We
```
io7m, line 88: Typo. :yum:

simplified-main/src/main/java/org/nypl/simplified/main/MainServices.kt
```diff
@@ -945,7 +945,7 @@ internal object MainServices {
       serviceConstructor = { return@addService AudioBookManifests }
     )
 
-    addService(
+    addServiceFromServiceLoaderOptionally(
       message = strings.bootingGeneral("metrics service"),
       interfaceType = MetricServiceType::class.java,
       serviceConstructor = { MetricService(context) }
```
io7m, line 951: I think I realize where the confusion is now.

The purpose of `ServiceLoader` is decouple the _provider_ of an interface from the _consumer_ of the interface. Essentially it's a way of saying "Give me an implementation of the `com.example.Thing` interface, and I don't care where it comes from!". When you call:

```ServiceLoader.load(com.example.Thing::class.java)```

... what happens is that the `ServiceLoader` class looks for that `META-INF/services/com.example.Thing` file that you created. It reads each line of the file, interprets each line as the fully-qualified name of a class, and then it calls the public no-arg constructor of those classes using reflection. It then returns the list of instances to you. You, as the caller, just see a `List<com.example.Thing>`, and you don't know or care where those instances came from. Effectively, you're the _consumer_ of the classes, and the _providers_ of those classes are hidden from you.

We use this in the app to find optional components. For example, take a look at the Crashlytics instantiation above:

```
addServiceFromServiceLoaderOptionally(
  message = strings.bootingGeneral("Crashlytics"),
  interfaceType = CrashlyticsServiceType::class.java
)
```
The `CrashlyticsServiceType` interface is defined in the `simplified-crashlytics-api` which is _always_ present on the classpath (it's a mandatory dependency). If you follow the definition of `addServiceFromServiceLoaderOptionally`, you'll see this call turns into:

```ServiceLoader.load(CrashlyticsServiceType::class.java)```

If the `simplified-crashlytics` module is present on the classpath (it will be in some builds and not others), the call to `ServiceLoader.load(...)` will find this file in the resulting APK file:

https://github.com/NYPL-Simplified/Simplified-Android-Core/blob/develop/simplified-crashlytics/src/main/resources/META-INF/services/org.nypl.simplified.crashlytics.api.CrashlyticsServiceType

You can see that file contains a single line: `org.nypl.simplified.crashlytics.CrashlyticsService`. `ServiceLoader` will look up the class `org.nypl.simplified.crashlytics.CrashlyticsService` by name and call the constructor, and return the new instance to you. If the `simplified-crashlytics` module _isn't_ present on the classpath, `ServiceLoader` won't find that file, and therefore won't know the names of any implementations classes, and will therefore return `null`. This is how we expose optional features in the map.

The way you've got things configured right now, you've created a file called `org.nypl.simplified.metrics.api.MetricServiceType` (which is correct), but the first line of that file reads `org.nypl.simplified.metrics.api.MetricServiceType`. `org.nypl.simplified.metrics.api.MetricServiceType` is an interface type, there's no constructor to call. The name you _should_ put in there is your implementation class `org.nypl.simplified.metrics.MetricService`.

The last problem: Assuming you _do_ put that class name in there, you'll get an error. The `MetricService` class has one constructor and that needs an argument passed in. What you need to expose via `ServiceLoader` is a `MetricServiceFactory` class in the implementation module with a corresponding `MetricServiceFactoryType` interface class in the API module. Something like this:

```
interface MetricServiceFactoryType {
  fun create(context: Context): MetricServiceType
}
```

```
class MetricServiceFactory {
  override fun create(context: Context): MetricServiceFactoryType {
    return MetricService(context)
  }
}
```

In the `MainServices` module, you would call `ServiceLoader.load(MetricServiceFactoryType::class.java)` and if there's an implementation of it, call the `create(...)` method on it and call `addService` on the resulting instance.

Please let me know if this is all gibberish. I'm aware that `ServiceLoader` is unfamiliar territory, and there's a little confusion here because we have `ServiceLoader` services, and then a completely unrelated service directory the app uses.

io7m:
This looks right to me! :+1:

io7m (approved):

PR #891, opened 2021-05-05
Implement and use a new navigation system

qnga:
**What's this do?**
This implements a system to quite safely communicate from child to parent fragments. It is used to replace the existing navigation API. Each fragment exposes a sealed class of events it can send to its parent fragment or activity. Each fragment or activity hosting child fragments claims its listening capabilities through a `ListenerRepository`. When a fragment wants to send an event, it recursively searches a listener able to handle it through the fragment/activity hierarchy (whatever their states are). When a hosting fragment or activity enters a state appropriate to do everything, in particular using its `FragmentManager`, it registers with the `ListenerRepository` which will call the provided handler every time a new event is received, until the callback is unregistered. Events received when no handler is registered are queued up.

As implementations of exiting `NavigationControllerType` store references to `FragmentManager`s, they must be recreated along with the Activity or parent Fragment on configuration changes. Therefore, the ability for fragments to navigate (without crashing) is tied to the lifecycle of their parent Fragment or Activity.

Besides, the next destination was hard-coded inside fragments. To properly separate concerns and be able to have different navigation graphs for onboarding and main fragments, it should be the responsibility of the parent to decide it. Not every fragment of a module is used in all navigation graphs, so it also looks better to separate the "navigation" needs of each fragment.

Comprehensively testing navigation through all possible paths.

It currently depends on a custom fix on the `BottomNavigator` library that will hopefully be merged upstream soon.

I think so.

Yes, I did.


io7m (approved):
I think it's important that we get this into testing ASAP. Potentially fixing the top ten app crashes is huge!

PR #892, opened 2021-05-10
SMA-123: Added Back Button to Card Creator Screen

Dumblydore:
**What's this do?**
Adds a back button to toolbar & added more checks around signup button being enabled to prevent crashes

https://jira.nypl.org/browse/SMA-123

Verify that the card creator back button behavior is correct

None

No documentation updates needed

I have


io7m (approved):
This looks right to me, but Malcolm knows this particular bit of the code better than I do.

MalcolmMcFly (approved):

PR #893, opened 2021-05-13
Make BookController API uniformly use TaskResult and not raise exceptions

qnga:
**What's this do?**
Make BookController API uniformly use TaskResult and not raise exceptions

This enables to simplify the calling code in UI module. Some methods returned a `TaskResult` while anyway possibly raising exceptions, which is confusing and difficult to work with.

Tests had been modified to match the new API.

None

N/A

Yes, I did.


simplified-books-controller/src/main/java/org/nypl/simplified/books/controller/AbstractBookTask.kt
```diff
@@ -0,0 +1,129 @@
...
+      result
+    } catch (e: Exception) {
+      this.logger.error("task failed with unhandled exception: ", e)
+      val msg = e.message ?: e.javaClass.name
+      this.taskRecorder.currentStepFailedAppending(msg, BorrowErrorCodes.unexpectedException, e)
+      val result = this.taskRecorder.finishFailure<Unit>()
+      this.onFailure(result)
+      result
+    }
+  }
```
io7m, line 74: :smile:

io7m (approved):
I like it!

PR #896, opened 2021-05-14
Fix an SLF4J crash

io7m:
**What's this do?**
I recently removed the SLF4J version resolution adjustment thinking
that it was no longer necessary. Turns out that it's still necessary. Fixes
a crash on startup.

Affects: https://jira.nypl.org/browse/SMA-176

Check the app starts up!

None.

N/A

@io7m ran it, with a sense of shame.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -68,6 +68,17 @@ ext {
   }
 }
 
+subprojects {
+  /*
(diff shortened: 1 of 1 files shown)


killerpuppytails (approved):
No need for shame! 😆 We caught it before deploy 👍

PR #897, opened 2021-05-17
Fix feedIsRoot for custom OPDS feeds

qnga:
**What's this do?**
Move `feedIsRoot` method from `AccountProviderType` to `AccountReadableType` and add a check for the root URI provided when the account was created.

With custom OPDS feeds, `AccountProviderType.feedIsRoot` failed to identify the catalog root because two different URIs coexist: one provided by the user when the account was created (stored in `AccountPreferences`), and another one from the authentication document.

Check that the account picker is shown on the catalog of a custom OPDS feed.

N/A

N/A

Yes, I did.


io7m (approved):
Yep, looks right!

PR #898, opened 2021-05-18
Removed Library Finder UI

MalcolmMcFly:
**What's this do?**
[Description of what this PR does goes here]

[Quick blurb about why the code is needed and Jira link goes here / Do these changes meet the business requirements of the story?]

[Description of any tests that need to be performed once merged goes here]

[Description of any watchouts, dependencies, or issues we should be aware of goes here]

### DIFF
--- a/simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/AccountListRegistryViewModel.kt
+++ b/simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/AccountListRegistryViewModel.kt
@@ -59,11 +59,21 @@ class AccountListRegistryViewModel(private val locationManager: LocationManager)
       this.profilesController.accountEvents()
         .observeOn(AndroidSchedulers.mainThread())
         .subscribe(this::onAccountEvent),
-      Observable.combineLatest(queries, locationUpdates) { query, _ -> createQuery(query) }
-        .switchMapCompletable(this::executeQuery)
-        .subscribeOn(Schedulers.io())
+      /**
+       * Here we are hiding Library Finder functionality
+       */
+      queries.map {
+        createQuery(it)
+      }.switchMapCompletable {
+        executeQuery(it)
+      }.subscribeOn(Schedulers.io())
         .onErrorComplete()
         .subscribe()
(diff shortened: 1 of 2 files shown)


Dumblydore (approved):
Looks good!

simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/AccountListRegistryViewModel.kt
```diff
@@ -59,11 +59,21 @@ class AccountListRegistryViewModel(private val locationManager: LocationManager)
       this.profilesController.accountEvents()
         .observeOn(AndroidSchedulers.mainThread())
         .subscribe(this::onAccountEvent),
-      Observable.combineLatest(queries, locationUpdates) { query, _ -> createQuery(query) }
-        .switchMapCompletable(this::executeQuery)
-        .subscribeOn(Schedulers.io())
+      /**
+       * Here we are hiding Library Finder functionality
+       */
+      queries.map {
```
io7m, line 65: This seems kind of weird. Why has the above statement been unrolled like this? It doesn't seem to actually change behaviour...

MalcolmMcFly: @Dumblydore @io7m check latest commit

simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/AccountListRegistryViewModel.kt
```diff
@@ -59,11 +59,14 @@ class AccountListRegistryViewModel(private val locationManager: LocationManager)
...
-        .onErrorComplete()
-        .subscribe()
+      /**
+       * Here we are hiding Library Finder functionality
+       */
+//      Observable.combineLatest(queries, locationUpdates) { query, _ -> createQuery(query) }
+//        .switchMapCompletable(this::executeQuery)
+//        .subscribeOn(Schedulers.io())
+//        .onErrorComplete()
+//        .subscribe()
```
io7m, line 69: Just want to confirm that commenting this out doesn't completely remove any library registry queries...

If you're confident that it doesn't, then this is fine.

PR #899, opened 2021-05-20
Disable location check

MalcolmMcFly:
**What's this do?**
Disables location check added in Library Finder

n/a

Launch the app and select "Find Your Library". There should be no location request.

n/a

n/a

@MalcolmMcFly


Dumblydore:
I might be worthwhile to also comment out the Location permission in the manifest

MalcolmMcFly:
> I might be worthwhile to also comment out the Location permission in the manifest.

Is that the only place we are using location? I know its used in the Card Creator but that is in a separate module.

io7m:
This seems OK to me. The card creator will bring in the location permission when the manifests are merged anyway.

PR #904, opened 2021-06-01
Refactor simplified-ui-catalog and some related backend stuff

qnga:
**What's this do?**
- should remove most of - if not every - possible crashes due to configuration changes by getting rid of runOnUIThread in the catalog
- navigation commands are now launched from ViewModels in the catalog module
- the BookRegistry now keeps track of availability of known books until the app is closed
- the loans feed (locally generated) is not reloaded every time patrons switch to the Books tab any more. It's instead reloaded only every time it looks necessary to
- fragments talking about the same book (for instance book detail opened in both Catalog and Books tab) have synchronized status
- not-patron-specific book status are fetched from the server after logging out when the server allows unauthenticated requests Otherwise, availability doesn't make sense any more and UI should be changed to reflect this in a future work.
- hopefully any other status-related issue after some operation triggered by the user on the device should be fixed

- orientation changes were not handled properly
- book status management was basic and very confusing
- status was not (always) shared between fragments talking about the same book

- do any loan management operation on a book when multiple fragments refer to this book, check availability is synchronized
- log in/log out and check book availability is updated in an appropriate way

N/A

N/A

Yes, I did. There are lots of changes though, so testing it again and again would be useful.


simplified-books-controller/src/main/java/org/nypl/simplified/books/controller/BookRevokeTask.kt
```diff
@@ -72,7 +73,6 @@ class BookRevokeTask(
 
   private lateinit var databaseEntry: BookDatabaseEntryType
   private val adobeACS = "Adobe ACS"
-  private var databaseEntryInitialized: Boolean = false
```
io7m, line 75: Glad to see this gone.

simplified-feeds-api/src/main/java/org/nypl/simplified/feeds/api/FeedLoader.kt
```diff
@@ -39,14 +36,10 @@ import java.util.concurrent.atomic.AtomicBoolean
 
 /**
  * The default implementation of the [FeedLoaderType] interface.
- *
- * This implementation caches feeds. A feed is ejected from the cache if it has
- * not been accessed for five minutes.
```
io7m, line 44: :+1:

Forgot to remove that comment when I removed the cache.

simplified-ui-catalog/src/main/java/org/nypl/simplified/ui/catalog/CatalogFeedFragment.kt
```diff
@@ -20,108 +20,106 @@ import android.widget.RadioButton
...
-import org.nypl.simplified.ui.catalog.CatalogFeedState.CatalogFeedLoadFailed
-import org.nypl.simplified.ui.catalog.CatalogFeedState.CatalogFeedLoaded.CatalogFeedEmpty
-import org.nypl.simplified.ui.catalog.CatalogFeedState.CatalogFeedLoaded.CatalogFeedNavigation
-import org.nypl.simplified.ui.catalog.CatalogFeedState.CatalogFeedLoaded.CatalogFeedWithGroups
-import org.nypl.simplified.ui.catalog.CatalogFeedState.CatalogFeedLoaded.CatalogFeedWithoutGroups
-import org.nypl.simplified.ui.catalog.CatalogFeedState.CatalogFeedLoading
-import org.nypl.simplified.ui.errorpage.ErrorPageParameters
-import org.nypl.simplified.ui.images.ImageLoaderType
+import org.nypl.simplified.ui.catalog.CatalogFeedState.*
+import org.nypl.simplified.ui.catalog.CatalogFeedState.CatalogFeedLoaded.*
```
io7m, line 48: Star imports won't pass `ktlint`.

simplified-main/src/main/java/org/nypl/simplified/main/MainFragmentListenerDelegate.kt
```diff
@@ -213,10 +205,6 @@ internal class MainFragmentListenerDelegate(
         this.openErrorPage(event.parameters)
         state
       }
-      is CatalogBookDetailEvent.DownloadWaitingForExternalAuthentication -> {
```
io7m, line 216: I may not have gotten to it yet, but why is this no longer present?

In the original code, these branches were only reachable when you clicked a book that required a SAML login. They stayed in that state until the login had been completed.

simplified-ui-catalog/src/main/java/org/nypl/simplified/ui/catalog/CatalogFeedFragment.kt
qnga: Why did they, both locally and on CI?

simplified-main/src/main/java/org/nypl/simplified/main/MainFragmentListenerDelegate.kt
qnga: I moved this logic to the `MainFragment` because even if several fragments refer to the same book, the action should be performed only once.

simplified-ui-catalog/src/main/java/org/nypl/simplified/ui/catalog/CatalogFeedFragment.kt
io7m: Great, seems like `ktlint` isn't being run at all.

The build script contains these lines:

```
subprojects {
  assemble.dependsOn ktlint
}
```

So, naturally, they just fail silently and `ktlint` never runs.

io7m: I'm going to OK this, because I'm going to submit a PR afterwards that makes `ktlint` run again. I can fix style issues then.

io7m (approved):
I don't see anything objectionable in here. Once again, it's going to take me a bit of time to get an intuitive sense of how it works, but less so than the previous navigation update.

PR #906, opened 2021-06-02
SMA-183: Remove R1

io7m:
**What's this do?**
This removes R1 from the app. The dependency on the R1 library is
still present, because the Adobe DRM connector requires libepub3.so
to be present in the AAR, even though nothing actually calls it now.

The UI to switch to R2 is gone, and all EPUB files now go through R2.

Fix: https://jira.nypl.org/browse/SMA-183

Check that opening Adobe encrypted books works.
Check that opening open access books works.
Check that opening audio books works.

None.

The module list was regenerated.

@io7m Checked a few Adobe and open access books in SimplyE.


io7m:
Going to merge this, as there's essentially nothing to review.

Good riddance to R1!

PR #908, opened 2021-06-03
Remove MainActivityViewModel and (nowadays) useless clearHistory workaround

qnga:
**What's this do?**
Remove MainActivityViewModel.

clearHistory workaround is now useless.

I believe this should be tested by checking that deleting an account still clears navigation history.

None

N/A

Yes, I did.


io7m (approved):
Looks good.

PR #911, opened 2021-06-07
Introduce UIExecutor to fix a bug occuring when removing an account

qnga:
**What's this do?**
Introduce the class `UIExecutor` that should enable us to safely use Futures returned by the Controller in ViewModels. As opposed to events, Futures have the advantage of being owned by one ViewModel instance which can be the unique listener to its completion.

This is used to fix a bug in simplified-ui-catalog.

Deleting an account could cause a crash in case there were two library accounts: the SimplyE collection and another one.
When deleting the latter account while it was shown in the Catalog tab, the CatalogFeedFragment associated with it happened to receive the callback onAgeUpdated meant to the CatalogFeedFragment showing the SimplyE collection before being deleted. Then, as the account had been deleted, it couldn't be found any more and the app crashed.

Add a library account via "Find my library now", and remove it. Check the app doesn't crash and everything happens as expected.

None

N/A

Yes, I did.

### DIFF
--- a/simplified-futures/src/main/java/org/nypl/simplified/futures/FluentFutureExtensions.kt
+++ b/simplified-futures/src/main/java/org/nypl/simplified/futures/FluentFutureExtensions.kt
@@ -5,12 +5,24 @@ import com.google.common.util.concurrent.AsyncFunction
 import com.google.common.util.concurrent.FluentFuture
 import com.google.common.util.concurrent.Futures
 import com.google.common.util.concurrent.MoreExecutors
+import java.util.concurrent.Executor
 
 /**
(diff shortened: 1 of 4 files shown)


io7m:
Are we at the point where we can get rid of the rest of the `ui-thread` module? This looks largely compatible with the existing UI thread code, except that it's lifecycle-aware.

io7m (approved):
Looks fine.

qnga:
I think this `Executor` can be used safely only in ViewModels, because we don't want tasks to be cancelled on orientation changes. So I think there is still some refactoring work to be done before being able to get rid of the UIThreadService.

PR #912, opened 2021-06-14
Fix announcements UI

qnga:
**What's this do?**
 Rewrite the announcements UI in a proper lifecycle-aware and rotation-aware manner

ProfileType.mostRecentAccount becomes non-nullable, and the profile database format is updated accordingly.
A profile already had to have at least one account, so now the most recent account is set to be this account when the profile is created.

- The announcements were all considered as acknowledged in case of configuration change.
- The announcements were all considered as acknowledged if only some of them have been checked by the patron.
- Some announcements might be lost because of wrong handling of lifecycles.
- We want to keep on getting rid of runOnUIThread.

Add a custom OPDS feed for the QA circulation manager which has currently two announcements set.

None

N/A

Yes, I did.


io7m (approved):
Looks good to me!

PR #913, opened 2021-06-24
Fix several issues with the AccountDetailFragment

qnga:
**What's this do?**
- Refactor a bit the AccountDetailFragment
- Modernize and simplify CardCreator's API
- Remove the EULA checkbox because it was meaningless
- Actually close the AccountDetailFragment once logged in if the catalog was waiting for the operation's completion

- Login button status was wrong after having switched tab from the AccountDetailFragment and back.
- Barcode and PIN used to be lost on orientation changes.

- Try to log in and log out, switching device orientation.
- Leave the AccountDetailFragment switching tab and go back to it, check the status is correct.
- Check the CardCreator is still working.

None

N/A

Yes, I did.

### DIFF
--- a/simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/CardCreatorContract.kt
+++ b/simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/CardCreatorContract.kt
@@ -1,0 +1,73 @@
+package org.nypl.simplified.cardcreator
+
+import android.app.Activity
+import android.content.Context
+import android.content.Intent
+import android.os.Bundle
+import androidx.activity.result.contract.ActivityResultContract
+import androidx.core.os.bundleOf
+
+class CardCreatorContract(
+  private val username: String,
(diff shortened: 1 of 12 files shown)


simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/CardCreatorContract.kt
```diff
@@ -0,0 +1,73 @@
...
+import android.os.Bundle
+import androidx.activity.result.contract.ActivityResultContract
+import androidx.core.os.bundleOf
+
+class CardCreatorContract(
+  private val username: String,
+  private val password: String,
+  private val clientId: String,
+  private val clientSecret: String
+) : ActivityResultContract<CardCreatorContract.Input, CardCreatorContract.Output>() {
```
io7m, line 15: Interesting! This is the first time I've seen this.

io7m (approved):
Yep, looks good!

PR #915, opened 2021-07-07
SMA-186: Make the account bookmark sync switch reflect reality

io7m:
**What's this do?**
This adjusts the account bookmark sync switch so that it works based
off of an observable status value. When the switch is toggled, the
reader bookmark service tracks the progress of the operation and can
publish appropriate status values when asked. A small indefinite
progress indicator has been added to show that a long-running
operation is occurring when the switch is toggled, and the switch is
marked as disabled (non-interactable) when a sync setting change is
in progress. This prevents users from being able to queue up dozens
of switch changes by hammering the toggle repeatedly.

Affects: https://jira.nypl.org/browse/SMA-186

Ensure that the switch reflects what you know to be true of an existing account (ie, the account either has syncing enabled or disabled when you're logged in). Ensure that you can't just quickly and repeatedly tap the switch and end up with it toggling backwards and forwards long after you stopped tapping.

None.

N/A

@io7m Tried it lots with an NYPL account.


Dumblydore (approved):
Looks good to me 👍

PR #917, opened 2021-07-15
Feat/sma 211 - Update CI to use App Bundles Instead of APKs

Dumblydore:
**What's this do?**
- Adds `firebase-bundle.conf` to SimplyE & OpenEBooks app modules which point to app bundle (.aab) (see https://github.com/NYPL-Simplified/Simplified-Android-CI/pull/1 for changes to CI script)

https://jira.nypl.org/browse/SMA-211

Verify that CI build is successful and that firebase receives an .aab file instead of an .apk

Requires this PR: https://github.com/NYPL-Simplified/Simplified-Android-CI/pull/1 to be merged as well

Yes

No, can only really be tested once it's merged in unfortunately

### DIFF
--- a/simplified-app-openebooks/firebase-bundle.conf
+++ b/simplified-app-openebooks/firebase-bundle.conf
@@ -1,0 +1,4 @@
+build/outputs/bundle/release/openebooks-release.aab
+#
+# The first line of this file is the APK that will be deployed to Firebase.
+#
--- a/simplified-app-simplye/fastlane/Fastfile
+++ b/simplified-app-simplye/fastlane/Fastfile
@@ -34,11 +34,12 @@ platform :android do
         next
       end
       gradle(task: "clean")
-      gradle(task: "assembleRelease")
+      gradle(task: "bundleRelease")
        firebase_app_distribution(
          app: "1:220408294534:android:594dde26dd9d35c301f507",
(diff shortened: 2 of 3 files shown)


io7m:
I forgot to mention: I made some progress on this a few days ago, and the process to get the build system to produce _both_ APK and AAB files is trivial. I think what we might end up doing is applying some of the changes you've made here, but without removing the publications of APK files. I've tried it out and AAB files don't make the build take any longer than normal, so I'm guessing the tools are already producing them internally anyway and they just produce APK files at the last minute from the AABs.

MalcolmMcFly:
> I forgot to mention: I made some progress on this a few days ago, and the process to get the build system to produce _both_ APK and AAB files is trivial. I think what we might end up doing is applying some of the changes you've made here, but without removing the publications of APK files. I've tried it out and AAB files don't make the build take any longer than normal, so I'm guessing the tools are already producing them internally anyway and they just produce APK files at the last minute from the AABs.

So @io7m do you think we still want to use this?

io7m:
We will want to use part of it! It can stay here for now.

MalcolmMcFly:
@io7m should we close this?

io7m:
This can be closed now.

PR #918, opened 2021-07-23
SMA-226: Fix a crash related to stale account IDs

io7m:
**What's this do?**
* Catch any possible exception raised when configuring Crashlytics.
* Fix broken account IDs when initializing profiles.

Affects: https://jira.nypl.org/browse/SMA-226

Uncertain, because we don't know how the accounts got into this state in the first place. See Slack!

None.

N/A

@io7m Tried breaking the app, and wrote tests.


Dumblydore (approved):
Looks good to me 👍

PR #920, opened 2021-07-28
SMA-229: Add a toggle to enable/disable publisher CSS

io7m:
**What's this do?**
This switches to a version of the R2 Android component that can toggle publisher CSS.

Affects: https://jira.nypl.org/browse/SMA-229

Unfortunately, I'm not sure, because I don't know what the original problem that required this was. I suspect testing is going to mean trying out the books that were fixed by this change on iOS, and see if the same thing works on Android.

We're still pointing at a snapshot version. We should tag a new R2 release before releasing this to the store.

N/A

@io7m tested that this turned on/off publisher CSS.

### DIFF
--- a/build_libraries.gradle
+++ b/build_libraries.gradle
@@ -56,7 +56,7 @@ ext.versions = [
   nypl_overdrive                : '1.0.3',
   nypl_pdf                      : '0.1.0',
   nypl_readium                  : '0.30.0',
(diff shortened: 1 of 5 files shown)


io7m:
device-2021-07-28-185330
device-2021-07-28-185344

Dumblydore (approved):
Looks good to me!

PR #921, opened 2021-07-30
OE-149: Disable book returns in OE

io7m:
**What's this do?**
This adds a (deprecated) method to the catalog build configuration
service that allows for disabling book returns in applications.
This is only used by OE, and probably won't be used for all that long.

Affects: https://jira.nypl.org/browse/OE-149

Check that you don't see Return buttons in OE in the book detail pages.

None.

N/A

@io7m checked for a lack of buttons.


MalcolmMcFly (approved):
Looks good, will test on a few devices

PR #922, opened 2021-07-30
OE-182: Filter unsupported books locally

io7m:
**What's this do?**
This adds the extra support required to filter out books that
were supported when they were fulfilled, but may be unsupported
now (due to a DRM system becoming unsupported).

https://jira.nypl.org/browse/OE-182

1. Use an older build of OE that supported Adobe DRM.
2. Borrow/fulfill a few Adobe-encrypted books.
3. Upgrade to this version of OE.
4. Check that those previously-fulfilled Adobe books don't appear in your Books tab.

None.

N/A

@io7m Checked that Adobe books don't appear.


MalcolmMcFly (approved):
very straightforward

PR #923, opened 2021-08-02
Check snapshot dependencies in non-snapshot builds

io7m:
**What's this do?**
This implements a check in the build system that refuses to allow
non-SNAPSHOT builds of the app to build with -SNAPSHOT dependencies.
Essentially, this enforces reproducibility for release builds.

Fix: https://jira.nypl.org/browse/SIMPLY-3634

Try setting the version number to a version that doesn't end in -SNAPSHOT, and then seeing if the app will refuse to build with -SNAPSHOT dependencies.

None.

N/A

success

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -66,6 +66,36 @@ ext {
         ^ required by 'org.librarysimplified.drm.enabled'\
       """.stripIndent())
   }
+
+  //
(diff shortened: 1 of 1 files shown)


Dumblydore (commented):
Looks good to me!

Dumblydore (approved):

PR #924, opened 2021-08-18
OE-140: Implement an OE QA toggle

io7m:
**What's this do?**
This implements a toggle that will add and remove an Open eBooks
QA account in the debug menu. Hardcoding the account information,
unfortunately, appears to be the only way to implement this as
there isn't a registry present in OE that seems to know about the OE
QA server.

Fix: https://jira.nypl.org/browse/OE-140

Try pressing the button in the debug menu to add/remove an Open eBooks QA
account.

None.

N/A

@io7m tried it.


Dumblydore (approved):
Looks good 👍

PR #925, opened 2021-08-19
Fix some lifecycle issues in Reader2Activity

qnga:
**What's this do?**
Fix some lifecycle issues in Reader2Activity, notably set the fragment factory at the right place.

https://jira.nypl.org/browse/SMA-236

It can't be tested easily because bugs used to happen in edge situations.

None

N/A

I checked the reader was still working.


io7m (approved):
I see nothing wrong with this. :+1:

PR #928, opened 2021-08-31
Switch to Simplified-R2-Android 0.0.24-SNAPSHOT

qnga:
**What's this do?**
Switch to Simplified-R2-Android 0.0.24-SNAPSHOT and the new API.

**Why are we doing this? (w/ JIRA link if applicable)**
Mainly https://jira.nypl.org/browse/SIMPLY-3802

**How should this be tested? / Do these changes have associated tests?**
Check page numbers are correct with linear resources and absent with non-linear resources.

**Dependencies for merging? Releasing to production?**
None

**Have you updated the changelog?**
I did.

**Has the application documentation been updated for these changes?**
N/A

**Did someone actually run this code to verify it works?**
Yes, I did.

PR #929, opened 2021-09-01
SMA-200: Trim whitespace around usernames and passwords in AccountDetailFragment

qnga:
**What's this do?**
Trim whitespace around usernames and passwords in AccountDetailFragment.

https://jira.nypl.org/browse/SMA-200

-  Open SimplyE
-   Navigate to the Accounts section
-   Add a library that you are not already logged into for which you have valid credentials (e.g. not SimplyE Collection)
-   Enter your user ID and deliberately add a space at the end
-   Enter your password or PIN
-   Try to log in

Depends on #928 in order to have an up-to-date changelog.

Yes, I did.

N/A

Yes, I did.

### DIFF
--- a/simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/AccountAuthenticationViewBindings.kt
+++ b/simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/AccountAuthenticationViewBindings.kt
@@ -252,11 +252,11 @@ sealed class AccountAuthenticationViewBindings {
     }
 
     fun getPassword(): AccountPassword {
-      return AccountPassword(this.pass.text.toString())
+      return AccountPassword(this.pass.text.toString().trim())
     }
 
     fun getUser(): AccountUsername {
-      return AccountUsername(this.user.text.toString())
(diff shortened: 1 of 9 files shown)


simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/AccountAuthenticationViewBindings.kt
```diff
@@ -252,11 +252,11 @@ sealed class AccountAuthenticationViewBindings {
     }
 
     fun getPassword(): AccountPassword {
-      return AccountPassword(this.pass.text.toString())
+      return AccountPassword(this.pass.text.toString().trim())
```
io7m, line 255: Sad to think that this had been the problem for all of this time...

io7m (approved):
:+1:

simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/AccountAuthenticationViewBindings.kt
qnga: Hopefully!

PR #932, opened 2021-09-13
SMA-226: Account identifier disappears

MalcolmMcFly:
**What's this do?**
Here a new custom exception has been added to `BottomNavigators.kt`. When the most recent account is accessed and it is stale for whatever reason, `StaleAccountException` is thrown and as a result the `AccountListRegistryFragment()` is returned so the user has to pick their account again.

[SMA-226: Account identifier disappears](https://jira.nypl.org/browse/SMA-226)

This fix can be tested by modifying the SharedPreferences for the app.
1. Navigate to `/data/data/org.nypl.simplified.simplye/files/profiles/*/profile.json`
View > Tool Windows > Device File Explorer or click the Device File Explorer
2. Delete `mostRecentAccount` entry.
3. Restart the app.

Desired Result: Whenever the user gets to a point where the account identifier is blank resulting from the most recent account being null, the user is prompted to select an account

n/a

Done.

n/a

TBD

### DIFF
--- a/README-CHANGES.xml
+++ b/README-CHANGES.xml
@@ -222,6 +222,11 @@
             <c:ticket id="SMA-237"/>
           </c:tickets>
         </c:change>
+        <c:change date="2021-09-13T12:27:18+00:00" summary="Account identifier disappears">
+          <c:tickets>
+            <c:ticket id="SMA-226"/>
+          </c:tickets>
+        </c:change>
       </c:changes>
     </c:release>
   </c:releases>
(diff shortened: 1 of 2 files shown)


qnga (commented):
The patch doesn't make much sense IMHO: several bottom navigation tabs will point to the `AccountListRegistryFragment`, even after the user has picked up an account again.

This bug looks like a thread synchronisation issue and may deserve a deeper solution if there's one.

MalcolmMcFly:
@qnga Thanks for the feedback here. I am going to test this a bit more, but my theory is that once the account is picked, the navigation would update automatically in response to picking a new account.

PR #934, opened 2021-09-22
SMA-210: Update rules for passwords in the Card creator

qnga:
**What's this do?**
Update rules for passwords in the Card creator

https://jira.nypl.org/browse/SMA-210

Check that good passwords allow to go further while bad ones don't.

None

Yes, I have.

N/A

Yes, I did.

### DIFF
--- a/README-CHANGES.xml
+++ b/README-CHANGES.xml
@@ -200,7 +200,7 @@
         </c:change>
       </c:changes>
     </c:release>
-    <c:release date="2021-09-22T07:44:18+00:00" is-open="true" ticket-system="org.nypl.jira" version="7.0.2">
+    <c:release date="2021-09-22T08:26:45+00:00" is-open="true" ticket-system="org.nypl.jira" version="7.0.2">
       <c:changes>
         <c:change date="2021-08-31T00:00:00+00:00" summary="Switch to a global byte-based page numbering">
           <c:tickets>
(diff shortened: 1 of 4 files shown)


MalcolmMcFly (approved):

build.gradle
```diff
@@ -98,17 +109,6 @@ ext {
...
-  /*
-   * We currently have transitive dependencies that specify dependencies on newer versions
-   * of SLF4J. These are subtly incompatible with Logback-Android, and it will be a while before
-   * new versions appear of both artifacts. Remove this when both are updated!
-   */
-  configurations.all {
-    resolutionStrategy.force libraries.slf4j
-  }
-}
-
```
MalcolmMcFly: Why do we no longe need this? Are there no more transitive dependencies ?

qnga: No idea. I guess it's @io7m 's work , as almost everything outside the card creator. I don't understand why the diff is showing this.

qnga:
Oh! Looks like I succeeded to get rid of Mark's work.

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/ui/ReviewFragment.kt
```diff
@@ -82,143 +79,117 @@ class ReviewFragment : Fragment() {
...
-      )
+        )
+      }
+    viewModel.createJuvenileCard(identifierParent)
+  }
+
+  private fun handleJuvenilePatronResponse(response: JuvenilePatronResponse) {
+    when (response) {
+      is JuvenilePatronResponse.JuvenilePatronData -> {
+        Toast.makeText(activity, "Card created", Toast.LENGTH_SHORT).show()
```
MalcolmMcFly: This probably needs to be in strings.xml

PR #935, opened 2021-09-24
Update release build process

MalcolmMcFly:
**What's this do?**
Updated release document to include library dependency checks before publishing builds to QA.

This tickets was in response to builds being pushed to QA without the correct versions of dependencies.

Yes

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -82,6 +82,21 @@ This creates a new `release/99.0.0` branch to which various commits
 may be made to increment version numbers, update change logs, run
 any last test builds, etc.
 
+#### Verify Library Dependencies
+
+SimplyE uses several remote repositories as library dependencies published to Maven Central. Verify that each library dependency is referencing the latest version in the `build_libraries.gradle` file. The main libraries to focus on here are:  
+[nypl_audiobook_api](https://github.com/NYPL-Simplified/audiobook-android)
+[nypl_drm_adobe](https://github.com/NYPL-Simplified/DRM-Android-Adobe)
+[nypl_drm_axis](https://github.com/NYPL-Simplified/DRM-Android-Axis)
+[nypl_drm_core](https://github.com/NYPL-Simplified/DRM-Android-Core)
+[nypl_findaway](https://github.com/NYPL-Simplified/audiobook-audioengine-android)
+[nypl_http](https://github.com/NYPL-Simplified/Simplified-Android-HTTP)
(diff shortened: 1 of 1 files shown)


RELEASING.md
```diff
@@ -82,6 +82,21 @@ This creates a new `release/99.0.0` branch to which various commits
 may be made to increment version numbers, update change logs, run
 any last test builds, etc.
 
+#### Verify Library Dependencies
+
+SimplyE uses several remote repositories as library dependencies published to Maven Central. Verify that each library dependency is referencing the latest version in the `build_libraries.gradle` file. The main libraries to focus on here are:
```
killerpuppytails, line 87: There will be some times when the latest version of a thing isn't necessarily what we want to release. I'd suggest something like "Verify that each library dependency is referencing the latest version in the `build_libraries.gradle` file or, if not, check with Product to see if exclusion is approved."

RELEASING.md
```diff
@@ -82,6 +82,21 @@ This creates a new `release/99.0.0` branch to which various commits
...
+[nypl_audiobook_api](https://github.com/NYPL-Simplified/audiobook-android)
+[nypl_drm_adobe](https://github.com/NYPL-Simplified/DRM-Android-Adobe)
+[nypl_drm_axis](https://github.com/NYPL-Simplified/DRM-Android-Axis)
+[nypl_drm_core](https://github.com/NYPL-Simplified/DRM-Android-Core)
+[nypl_findaway](https://github.com/NYPL-Simplified/audiobook-audioengine-android)
+[nypl_http](https://github.com/NYPL-Simplified/Simplified-Android-HTTP)
+[nypl_overdrive](https://github.com/NYPL-Simplified/audiobook-android-overdrive)
+[nypl_pdf](https://github.com/NYPL-Simplified/pdfreader-android)
+[nypl_readium2](https://github.com/NYPL-Simplified/Simplified-R2-Android)
+
```
killerpuppytails, line 97: Missing card creator?

killerpuppytails (changes requested):

RELEASING.md
MalcolmMcFly: So on iOS the card creator is a remote dependency, but on Android its built into core.

killerpuppytails: Ooh we should change that in the future, but for now that's fine.

killerpuppytails (approved):

PR #943, opened 2021-11-02
SMA-277: Upgrade to API 30

io7m:
**What's this do?**
This upgrades the build setup to target API 30. This involved
correcting a few methods to handle parameters that are now marked as
non-nullable instead of nullable.

Affects: https://jira.nypl.org/browse/SMA-277

The card creator was affected, but the code that was removed should have been unreachable. Maybe just check the card creator still works.

None.

N/A

N/A

@io7m Ran the app and poked at the settings screen.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -29,9 +29,9 @@ plugins {
 
 ext {
   android_build_tools_version = "30.0.2"
-  android_compile_sdk_version = 29
+  android_compile_sdk_version = 30
   android_min_sdk_version = 21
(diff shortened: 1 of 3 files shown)


MalcolmMcFly (approved):
These changes are sound.

MalcolmMcFly:
@io7m am I correct in thinking this also applies to OE? https://jira.nypl.org/browse/OE-255

io7m:
Yep.

PR #944, opened 2021-11-03
SMA-33: Change the way book IDs are computed

qnga:
**What's this do?**
- Change the way book IDs are computed. They used to be hashed versions of OPDS identifiers. They are now "x"+ hash(hash(opdsIdentifier) + hash(accountId.toString())).
- Old IDs are migrated when the database is opened.
- Remove the legacy module simplified-migration-from3master.

https://jira.nypl.org/browse/SMA-33

This makes changes at very different locations. Check everything in the app works as usual.

None

Yes

N/A

Yes, I did.


simplified-books-database/src/main/java/org/nypl/simplified/books/book_database/BookDatabase.kt
```diff
@@ -42,38 +42,38 @@ class BookDatabase private constructor(
    * A thread-safe map exposing read-only snapshots of database entries.
    */
 
-  private class BookMaps internal constructor() {
+  private class BookMaps() {
 
-    internal val mapsLock: Any = Any()
+    val mapsLock: Any = Any()
```
io7m, line 47: Why is this public now? It's dangerous to expose an internal lock in this way.

qnga: The modifier was redundant because the whole class is private.

io7m: Ah, I see!

io7m (approved):
Looks good! Glad to see the migration go away.

PR #948, opened 2021-11-16
SMA-209: Always parse dates as UTC

io7m:
**What's this do?**
This attempts to ensure that dates in OPDS feeds are always parsed in UTC timezones (unless they explicitly specify a timezone).

Affects: https://jira.nypl.org/browse/SMA-209

See if the original OPDS feed (in the ticket) can still fail to load. Unfortunately, this one is likely to be hard to test as I couldn't reproduce it.

None.

N/A

@io7m Ran the app and the test suite.

### DIFF
--- a/README-CHANGES.txt
+++ b/README-CHANGES.txt
@@ -1,0 +1,1 @@
+This space intentionally left blank.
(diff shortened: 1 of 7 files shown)


io7m:
Investigating test failures. They don't fail locally...

MalcolmMcFly (approved):
Looks good

PR #949, opened 2021-11-18
SMA-288: Fix many (hopefully all) lifecycle issues and add support to landscape mode in the epub viewer

qnga:
**What's this do?**
Fix many (hopefully all) lifecycle issues and add support to landscape mode in the epub viewer

https://jira.nypl.org/browse/SMA-288

- Check switching the device orientation actually switches the reader orientation without any crash.
- Check moving the app to the background and back to the foreground doesn't raise any crash or issue.

https://github.com/NYPL-Simplified/Simplified-R2-Android/pull/20

N/A

Yes, I did.


io7m (approved):
Nice.

PR #950, opened 2021-12-01
Fix IA logins

io7m:
**What's this do?**
This makes the changes necessary to log into Archive.org.  This
basically meant making the patron user profile parser more permissive,
because they don't include information that we'd essentially set as
required in the parser (despite the information being optional).

Affects: https://jira.nypl.org/browse/SMA-171

Check that you can log in to the internet archive using an email address and password.

None.

N/A

@io7m logged in:

device-2021-12-01-151452

### DIFF
--- a/simplified-patron/src/main/java/org/nypl/simplified/patron/PatronUserProfileParser.kt
+++ b/simplified-patron/src/main/java/org/nypl/simplified/patron/PatronUserProfileParser.kt
@@ -221,7 +221,8 @@ internal class PatronUserProfileParser(
   private fun parseAuthorization(root: ObjectNode): PatronAuthorization? {
     return try {
       val identifier =
-        JSONParserUtilities.getString(root, "simplified:authorization_identifier")
+        JSONParserUtilities.getStringOrNull(root, "simplified:authorization_identifier")
+          ?: return null
       val expires =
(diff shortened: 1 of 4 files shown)


qnga (approved):
I suppose it's ok, though I don't have enough context.

io7m:
For a bit of context: The Internet Archive have been trying to get support going for their collection going in the Library Simplified clients for quite some time. However, they don't use a circulation manager on their side - they're using all custom software for generating and serving feeds, authentication, etc. This means that whenever we try to talk to Archive.org, we discover all kinds of not-quite-valid assumptions we've made in our own code, because our code only ever talks to the same kind of server software where those assumptions normally always hold.

This PR fixes one of those faulty assumptions. We're erroneously making something a hard requirement in some parser code, when it's actually optional data.

qnga:
Sorry, I was thinking of code context. Why did you add a test on an OPDS entry while you fixed the user patron profile parser?

io7m:
...

Wow, that was entirely accidental. I accidentally committed the two files I was using during the investigation of SMA-83. They aren't related to this PR at all. Sorry for the noise!

The good news is that we probably want to keep them anyway; they're evidence that SMA-83 is now "won't do" as the code is already working correctly (due to all the borrow refactoring work that happened earlier in the year).

qnga:
Ok! Now I can understand this PR.

PR #952, opened 2021-12-07
Prevent library account deletion when only one account present

robbyice:
**What's this do?**

Changes the AccountListAdapter to conditionally show/hide the popup options menu based on the number of elements in the adapter when items are initially drawn. 

It's worth noting that the visibility of the menu won't update when items are removed/added to the list, it is only set as the items are bound. Currently on account deletion users are routed back to the catalog, so it seemed like there was no need to 'refresh' the options menu visibility for items in the adapter.

Also added as part of this PR is the infrastructure necessary to run Robolectric tests, and an initial small test for the adapter that runs using Robolectric. There is room for improvement around even this minimal test but it's a good first step.

Some other small things...

Renaming:
account_list_item -> account_list_registry_item
account_list_item_old -> account_list_item

Mockito:
Updated our mockito-kotlin dependency to use the official mockito-kotlin repo (they absorbed the nhaarman library)

DataBinding:
Using databinding for the changes to the adapter items' layout requires databinding being enabled in simplified-ui-accounts as well as any app that consumes and uses that module: openEbooks, simplyE, and vanilla.

https://jira.nypl.org/browse/SMA-173

Users should not be given the option to delete accounts from a device when there is only one account remaining. The app shows an error currently when deletion on the final account happens. Now, there should be no options menu present when only one account is registered.

Load up the OpenEbooks, SimplyE, or Vanilla app and check the visibility of the options menu with one/multiple accounts registered.

There are some light unit tests in AccountListAdapterTest - these run using Robolectric.

This PR depends on significant changes to the dependencies module. Here is a PR for all the dependency-related changes:
https://github.com/NYPL-Simplified/Simplified-Android-Platform/pull/2/

No changes to documentation

Tested it with SimplyE and Vanilla on my Pixel2 device.


simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/AccountListAdapter.kt
```diff
@@ -29,9 +28,9 @@ class AccountListAdapter(
     viewType: Int
   ): AccountViewHolder {
     val inflater = LayoutInflater.from(parent.context)
-    val itemView = inflater.inflate(R.layout.account_list_item_old, parent, false)
+    val binding = AccountListItemBinding.inflate(inflater, parent, false)
```
qnga, line 31: I've always seen some end of life management with view bindings. Isn't it necessary?

simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/AccountListAdapter.kt
```diff
@@ -41,83 +40,57 @@ class AccountListAdapter(
...
-      this.itemView.setOnClickListener {
-        this.accountItem?.let { account ->
-          this.onItemClicked.invoke(account)
+      if (hasOptionsMenu) {
+        binding.popupMenuIcon.setOnClickListener {
+          showPopupMenu(account)
         }
+      } else {
+        binding.popupMenuIcon.visibility = View.GONE
       }
```
qnga, line 74: _I_ would have disabled the menu or just the Delete button instead of hiding it. I know Mark suggested the opposite and that's not a big deal though.

simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/AccountListAdapter.kt
robbyice: I believe that should only be necessary when using LiveData in your bindings, so not needed here.

qnga (approved):
Looks good!

PR #954, opened 2021-12-16
SMA-290: Catalog Search button requires characters in search input

robbyice:
- Added Databinding to most CatalogFragment UI components
- Added Robolectric isolated fragment tests for the CatalogFragment
- Refactored CatalogFragment

**What's this do?**

1) Shared Test Dependency Configuration
To make configuring test dependencies across multiple modules easier, test dependency configuration has been extracted into a separate gradle file that can be applied on a per-module basis. If need be, at some point we could break apart this build-test.gradle file into different components to enable modules to apply test dependencies more modularly.

2) DataBinding in the CatalogFeedFragment
There was a significant amount of volume of code in the CatalogFeedFragment dealing with toggling on/off the visibility of views included in the fragment's layout according to current state. This logic is now done through DataBinding and custom BindingAdapters. It's worth noting that there are still a few UI components included in this fragment that have not yet had DataBinding applied. 

3) Isolated Fragment Testing
There are now isolated fragment unit tests for the CatalogFeedFragment. These run using Robolectric to enable unit testing of Android components (Fragments, Activities) in the JVM - no instrumented device required. Setting up the isolation of the fragment for these tests was a bit challenging and the result is not as "clean" as I would like, but it is a good first step.

4) New mocking library
Some creative mocking was required in order to get reasonable tests set up. Particularly, accommodating our dependency loading mechanism `Services`, the `FragmentFinderListener` utilized by the lazy delegate `findFragmentListeners`, and the fragment viewmodels. This mocking is enabled by MockK, which is a dedicated Kotlin mocking library with support for mocking constructor invocation and Kotlin singleton objects. Overall I am impressed by this library and will be using it over Mockito in the future.

5) CatalogFeedFragment Refactor
As part of adding DataBinding to the fragment, the CatalogFeedFragment has been changed. Many of the view references are no longer needed and some of the state-driven `reconfigureUI` logic has been reduced and consolidated. References to adapters are no longer maintained directly, and instead accessed through the view binding. There is definitely still room for improvement in this fragment. In the interest of not keeping this ticket open for too long I avoided refactoring certain pieces of code that I felt required additional attention. Namely, the dynamic "Facet" view building and the Toolbar configuration. Despite more work needing to be done I feel this is a good first step. 

https://jira.nypl.org/browse/SMA-290

Currently, when you perform a search with an empty search text the app displays an error. Instead of this, we should conditionally enable/disable the search button in the popup dialog. The button should require at least 2 characters input in order to be enabled.

Load up the Catalog in any version of the app and open up the search dialog. Check that the search submit button is disabled unless you've input at least 2 characters.

This behavior is explicitly tested inside the CatalogFeedFragmentTest class.

Dependent on the additional changes to dependencies in https://github.com/NYPL-Simplified/Simplified-Android-Platform/pull/3

No changes to documentation

Tested on an emulator and locally on my device.


simplified-ui-catalog/src/main/java/org/nypl/simplified/ui/catalog/CatalogFeedFragment.kt
```diff
@@ -44,23 +45,22 @@ import org.nypl.simplified.ui.catalog.CatalogFeedOwnership.CollectedFromAccounts
...
 /**
  * A fragment displaying an OPDS feed.
  */
 
-class CatalogFeedFragment : Fragment(R.layout.feed), AgeGateDialog.BirthYearSelectedListener {
+class CatalogFeedFragment : Fragment(), AgeGateDialog.BirthYearSelectedListener {
 
   companion object {
 
-    private const val PARAMETERS_ID =
```
qnga, line 63: I think this should be internal. You could add a comment "Visible for testing" as well.

simplified-ui-onboarding/build.gradle
```diff
@@ -8,7 +8,7 @@ dependencies {
   api project(":simplified-ui-errorpage")
 
   implementation libs.androidx.app.compat
-  implementation libs.androidx.lifecycle
+  implementation libs.androidx.lifecycle.ext
   implementation libs.androidx.fragment
```
qnga, line 12: Why did you add this dependency to all UI modules?

simplified-ui-catalog/src/main/java/org/nypl/simplified/ui/catalog/CatalogFeedBindingAdapters.kt
```diff
@@ -0,0 +1,46 @@
+package org.nypl.simplified.ui.catalog
+
+import android.view.View
+import androidx.databinding.BindingAdapter
+
+@BindingAdapter("showWhenFeedAgeGate")
+fun View.showWhenFeedAgeGate(feedState: CatalogFeedState) {
+  visibility = if (feedState is CatalogFeedState.CatalogFeedAgeGate)
```
qnga: I suppose all these functions could be internal.

qnga (commented):
Looks good to me. Though I personally agree with the choices you made, we used to have some discussions before adopting new libraries or techniques. If every one uses his favourite test library, things could become difficult. I might have missed those discussions though.

simplified-ui-catalog/src/main/java/org/nypl/simplified/ui/catalog/AutoClearedValue.kt
```diff
@@ -0,0 +1,47 @@
...
+import kotlin.reflect.KProperty
+
+/**
+ * https://github.com/android/architecture-components-samples/blob/master/GithubBrowserSample/app/src/main/java/com/android/example/github/util/AutoClearedValue.kt
+ *
+ * A lazy property that gets cleaned up when the fragment's view is destroyed.
+ *
+ * Accessing this variable while the fragment's view is destroyed will throw NPE.
+ */
+class AutoClearedValue<T : Any>(val fragment: Fragment) : ReadWriteProperty<Fragment, T> {
```
qnga: I think this could be moved into the module `simplified-android-ktx`.

simplified-ui-catalog/src/main/java/org/nypl/simplified/ui/catalog/AutoClearedValue.kt
```diff
@@ -0,0 +1,47 @@
...
+
+  override fun setValue(thisRef: Fragment, property: KProperty<*>, value: T) {
+    _value = value
+  }
+}
+
+/**
+ * Creates an [AutoClearedValue] associated with this fragment.
+ */
+fun <T : Any> Fragment.autoCleared() = AutoClearedValue<T>(this)
```
qnga: I dislike the use of the view's lifecycle being implicit here. We could name this `viewAutoCleared` or add a boolean argument `viewLifecycle` and make it working with the fragment's lifecycle too.

simplified-ui-onboarding/build.gradle
robbyice: Fair question - this is just renaming the dependency to more fully reflect the name of the actual artifact we are pulling in: `androidx.lifecycle:lifecycle-extensions`

Here's the commit onto platform with the change to `build_libraries.toml`: https://github.com/NYPL-Simplified/Simplified-Android-Platform/commit/3021b352c28f541347769142797bcce2435508b3

simplified-ui-catalog/src/main/java/org/nypl/simplified/ui/catalog/CatalogFeedFragment.kt
robbyice: I don't understand the need for it to be internal but I've only exposed it for testing purposes, so using `@VisibleForTesting` here makes sense

simplified-ui-catalog/src/main/java/org/nypl/simplified/ui/catalog/CatalogFeedBindingAdapters.kt
robbyice: Makes sense, will update this.

simplified-ui-catalog/src/main/java/org/nypl/simplified/ui/catalog/AutoClearedValue.kt
robbyice: Makes sense, I will move it

simplified-ui-catalog/src/main/java/org/nypl/simplified/ui/catalog/CatalogFeedFragment.kt
robbyice: Didn't even need the annotation, done!

simplified-ui-catalog/src/main/java/org/nypl/simplified/ui/catalog/AutoClearedValue.kt
robbyice: I also wasn't a huge fan of the name `autoCleared` so I'm happy to change it to better reflect what it is actually doing. 

I'm thinking `viewLifecycleAware()` or something like that.

robbyice:
Super fair. So I know better in the future - which libraries/techniques in particular do you think warranted discussion prior to use? Happy to chat more about anything here.

simplified-ui-onboarding/build.gradle
qnga: Sounds good.

simplified-ui-catalog/src/main/java/org/nypl/simplified/ui/catalog/CatalogFeedFragment.kt
qnga: Just to prevent use from dependent modules.

simplified-ui-catalog/src/main/java/org/nypl/simplified/ui/catalog/AutoClearedValue.kt
qnga: Agreed.

qnga (approved):

MalcolmMcFly (approved):

PR #955, opened 2022-01-06
SMA-295: When creating a library card, send an appropriate value for the "location" field

qnga:
**What's this do?**
- When creating a library card, send an appropriate value for the "location" field
- Suppress the fake location in the UI. It had been disabled and was unlikely to be ever enabled again.

This is the worst code I've ever written. The flow was difficult to keep track of in `LocationFragment` so I made minimal changes hoping it will work. I didn't want to entirely rewrite the UI because we're supposed to get rid of the native card creator soon or later.

https://jira.nypl.org/browse/SMA-295

Check you can get a permanent card both when you're in New York City, and when you're outside of it but in the New York State.
The card duration can be checked on nypl.org

None

N/A

I did with fake locations. This should also be tested in real conditions as the code slightly diverges.

### DIFF
--- a/simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/ui/LocationFragment.kt
+++ b/simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/ui/LocationFragment.kt
@@ -21,6 +21,7 @@ import androidx.activity.addCallback
 import androidx.appcompat.app.AlertDialog
 import androidx.core.content.ContextCompat
 import androidx.fragment.app.Fragment
+import androidx.fragment.app.activityViewModels
 import androidx.navigation.NavController
 import androidx.navigation.NavDirections
 import androidx.navigation.Navigation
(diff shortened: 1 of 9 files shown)


simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/ui/LocationFragment.kt
```diff
@@ -303,8 +317,8 @@ class LocationFragment : Fragment(), LocationListener {
    * Enables the next button and update header status
    */
   private fun enableNext(enable: Boolean) {
-    isNewYork = enable || locationMock
-    binding.nextBtn.isEnabled = enable || locationMock
+    isNewYork = enable || locationMock != null
+    binding.nextBtn.isEnabled = enable || locationMock != null
```
robbyice, line 321: I'm wondering - why always enable the button when the locationMock is present? Seems odd.

It doesn't really matter too much since the locationMock won't be set in production code and as you mentioned this component will soon be gone in favor of web-based card creation. I just thought it's a weird bit of logic - like if I call enableNext(false) due to some error state but I'm using a locationMock the app will allow me to move to the next step.

robbyice (approved):
Overall looks fine to me. I would advocate for refactoring and adding tests for some of this code but if we're moving away from app-based card creation entirely then it probably isn't worth the effort.

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/ui/LocationFragment.kt
qnga: We have to enable the next button in order to be able to go to the next screen. Otherwise, the app will be waiting for a real location in New York to enable the next button.

qnga:
Thanks for the review, Rob!

MalcolmMcFly (approved):

PR #956, opened 2022-01-07
SMA-299: Fix crash in AudioBookPlayerActivity

qnga:
**What's this do?**
Fix crash in AudioBookPlayerActivity

https://jira.nypl.org/browse/SMA-299

This cannot be tested.

None

N/A

Yes, I did.

### DIFF
--- a/simplified-viewer-audiobook/src/main/java/org/nypl/simplified/viewer/audiobook/AudioBookPlayerActivity.kt
+++ b/simplified-viewer-audiobook/src/main/java/org/nypl/simplified/viewer/audiobook/AudioBookPlayerActivity.kt
@@ -151,6 +151,15 @@ class AudioBookPlayerActivity :
     this.setContentView(R.layout.audio_book_player_base)
     this.playerScheduledExecutor = Executors.newSingleThreadScheduledExecutor()
 
+    /*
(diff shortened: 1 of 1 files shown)


simplified-viewer-audiobook/src/main/java/org/nypl/simplified/viewer/audiobook/AudioBookPlayerActivity.kt
```diff
@@ -151,6 +151,15 @@ class AudioBookPlayerActivity :
...
 
+    /*
+    * Create a new downloader that is solely used to fetch audio book manifests.
+    */
+
+    this.downloadExecutor =
+      MoreExecutors.listeningDecorator(
+        NamedThreadPools.namedThreadPool(1, "audiobook-player", 19)
+      )
+
```
MalcolmMcFly, line 162: @qnga this appears to just be moved up earlier in the activity. Is the problem that it's initialized too late?

qnga: The real problem is probably the `return` on line 206. The move of the property initialisation should act as a quick fix.

robbyice (approved):

PR #957, opened 2022-01-12
SMA-294: Fix layout issues with open keyboard

robbyice:
**What's this do?**
This PR adjusts the layout for the AccountInformationFragment - the screen where users enter a username/password as part of the card creation process.

These changes fix a bug with the layout on the AccountInformation screen where when the keyboard was opened, the password "helper" text would be pushed up and on top of the password form field. Now everything should readjust and display properly as the keyboard is opened/closed.

Navigate to the AccountInformation fragment as part of the card creation process. Open and close the keyboard and observe that the password hint text displays correctly.

None

No documentation changes

I tested it on my Pixel 2 device

### DIFF
--- a/simplified-cardcreator/src/main/res/layout/fragment_account_information.xml
+++ b/simplified-cardcreator/src/main/res/layout/fragment_account_information.xml
@@ -6,117 +6,110 @@
     android:layout_height="match_parent"
     tools:context=".ui.AccountInformationFragment">
 
+    <TextView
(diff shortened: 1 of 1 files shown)


robbyice:
I was wondering - does anyone know why the form fields are nested inside a ScrollView? I left it as is but it seems quite odd.

MalcolmMcFly (approved):
This looks good to me, but CI appears to be failing.

PR #958, opened 2022-01-13
SMA-297: Update NYPL HTTP Android Library

robbyice:
**What's this do?**
This PR bumps the platform submodule to point to a new version using NYPL Http Android 0.0.25-SNAPSHOT

Keeping dependencies up-to-date

https://jira.nypl.org/browse/SMA-297

Ensure the app and its associated tests run/function normally

No required dependencies.

No documentation changes

I ran the app on my Pixel 2 device and bounced around a few screens without issue.

### DIFF
--- a/simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/CardCreatorContract.kt
+++ b/simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/CardCreatorContract.kt
@@ -41,9 +41,9 @@ class CardCreatorContract(
       USER_IDENTIFIER_KEY to input.userIdentifier
     )
 
-  override fun parseResult(resultCode: Int, intent: Intent?): Output? =
+  override fun parseResult(resultCode: Int, intent: Intent?): Output =
     if (intent == null || resultCode != Activity.RESULT_OK) {
-      null
+      Output("", "")
     } else {
       this.parseResult(intent.extras!!)
     }
--- a/build.gradle
+++ b/build.gradle
@@ -29,9 +29,9 @@ plugins {
 
 ext {
   android_build_tools_version = "30.0.2"
-  android_compile_sdk_version = 30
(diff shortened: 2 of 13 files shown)


MalcolmMcFly:
@robbyice so I think what we want to do here is release a new version of NYPL HTTP Android Library first by incrementing the version and publishing 0.0.25. This should help: https://docs.google.com/document/d/1DIuMWKqpt4mNqJ1Cz795fZV5T6N3yYGyNTOY6vlMbQc/edit
Give it a look and let me know if we need to connect to walkthrough it.

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/CardCreatorContract.kt
```diff
@@ -41,9 +41,9 @@ class CardCreatorContract(
       USER_IDENTIFIER_KEY to input.userIdentifier
     )
 
-  override fun parseResult(resultCode: Int, intent: Intent?): Output? =
+  override fun parseResult(resultCode: Int, intent: Intent?): Output =
     if (intent == null || resultCode != Activity.RESULT_OK) {
-      null
+      Output("", "")
```
robbyice, line 46: This gets the code to compile but the implications of losing nullability here probably warrants a review of where/how this contract is used.

robbyice:
In order to resolve the naming collision with transitive dependencies I added an explicit dependency declaration for viewmodel-ktx wherever fragment-ktx is used. Wasn't sure if this "bottom-up" approach is preferable or if we'd rather do something "top-down" - i.e., the individual applications' build.gradle file resolves this (vanilla, simplyE, openEbooks). Either way seems a little iffy.

We could also be more explicit about what we are doing by excluding viewmodel-ktx from the fragment-ktx declaration. Or do something like this: https://docs.gradle.org/current/userguide/dependency_constraints.html

Open to ideas/suggestions here.

PR #960, opened 2022-01-28
SMA-111: Fix isLoginRequired in CatalogBorrowViewModel

qnga:
**What's this do?**
Fix bug when attempting to borrow a book in LogoutFailed state.

https://jira.nypl.org/browse/SMA-111

It's hard to test and not worth it.

None

N/A

Yes, I did, though not in logout failed state.


io7m (approved):
This appears to be correct!

PR #971, opened 2022-03-01
Remove migration API

qnga:
**What's this do?**
Removes migration API.

I removed the last migration a while ago. We won't use this global migration API any more.

Nothing to test.

None

N/A

Yes, I checked nothing was broken.


io7m (commented):
Nice. Really glad to see this finally going away.

MalcolmMcFly (approved):

PR #979, opened 2022-04-27
SMA-367: Card Creator Validation

robbyice:
**What's this do?**
This does some refactoring in a few pieces of the CardCreator and adds client-side validation for passwords to ensure they contain only alphanumeric characters.

https://jira.nypl.org/browse/SMA-367

Go to the account information page as part of the card creator process, and observe that the 'Next' button remains disabled until only alphanumeric characters are entered into the password field (in addition to length requirements)

No dependencies need merging

No documentation changes needed

Tested the validation on my Pixel 5a device


simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/ui/AccountInformationFragment.kt
```diff
@@ -11,33 +9,35 @@ import android.widget.Toast
...
+class AccountInformationFragment(
+  // Needed to enable isolated fragment testing with an activityViewModel
+  activityVMFactory: (() -> ViewModelProvider.Factory)? = null
+) : Fragment() {
 
   private val logger = LoggerFactory.getLogger(AccountInformationFragment::class.java)
 
-  private var _binding: FragmentAccountInformationBinding? = null
-  private val binding get() = _binding!!
+  private var binding by viewLifecycleAware<FragmentAccountInformationBinding>()
```
MalcolmMcFly, line 35: Is `viewLifecycleAware` the recommendation for viewBindings?

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/viewmodel/ConfirmationViewModel.kt
```diff
@@ -0,0 +1,69 @@
...
+  val events = _events.asSharedFlow()
+
+  private val _state = MutableStateFlow(initialState)
+  val state = _state.asStateFlow()
+
+  fun confirmCard() { viewModelScope.launch { _events.emit(CardConfirmed) } }
+  fun prepareToSaveCard() { viewModelScope.launch { _events.emit(SaveCardPermissionsCheck) } }
+
+  // This could probably be its own utility component
+  fun createAndStoreDigitalCard(bitmap: Bitmap) {
```
MalcolmMcFly, line 37: Great idea to move this out of the fragment.

MalcolmMcFly (approved):
Excellent test work!

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/viewmodel/ConfirmationViewModel.kt
```diff
@@ -0,0 +1,69 @@
...
+import java.io.File
+import java.io.FileOutputStream
+
+class ConfirmationViewModel(
+  initialState: ConfirmationData,
+) : ViewModel() {
+  private val logger = LoggerFactory.getLogger(ConfirmationViewModel::class.java)
+
+  private val _events = MutableSharedFlow<ConfirmationEvent>()
+  val events = _events.asSharedFlow()
```
qnga: I'm afraid that might not be correct (If I don't miss anything). I believe `Channels` must be used to pass single events from `ViewModels` to `Fragments`. With a `SharedFlow` without replay cache, you can miss events on configuration changes.

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/ui/AccountInformationFragment.kt
robbyice: I don't think it's an official recommendation, just something I picked up somewhere - I like it as an easy way to deal with managing binding references in a lifecycle-aware manner

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/viewmodel/ConfirmationViewModel.kt
robbyice: Yeah you are right this could lose events with a configuration change. I guess I didn't think it was a big deal given there's no logic being executed - they're just UI events routed through the view model rather than events representing a side effect of work the view model is doing. So worst case the user just hits the button again. Still not a great experience though and probably worth addressing. Also maybe a case for handling those directly in the view rather than through the viewmodel. 

I could probably just add some replay to the SharedFlow and fix most of the issue, though that still introduces some other potential problems. 

I did some reading and it seems like Channels don't fully prevent dropped events either. There's an interesting thread discussing single events that mentions it [here](https://gist.github.com/gmk57/330a7d214f5d710811c6b5ce27ceedaa?permalink_comment_id=3744320#gistcomment-3744320). For the code in question we are sometimes emitting events in a background thread.

Recently Google put forth their own [recommended approach](https://developer.android.com/topic/architecture/ui-layer/events#handle-viewmodel-events) for single events. They advocate exposing events as part of the viewmodel state, rather than a separate stream, and to have the view notify the viewmodel when events have been processed. I'd like to try their recommendation even though I'm sure it'll change again in a year.

[Here's](https://proandroiddev.com/sending-view-model-events-to-the-ui-eef76bdd632c) a good article talking about Google's new recommended approach.

qnga: Interesting reading about Channels ! It looks like there is no issue if events are sent on the main thread though, which is not a big deal. Your new approach looks totally fine too. It's more complex, but more obvious to prove to be correct as well.

qnga (approved):

PR #980, opened 2022-04-28
SMA-368: Keep Screen Awake when EReading

robbyice:
**What's this do?**
This PR makes it so that the screen will stay awake whenever the user is reading an ebook. 

https://jira.nypl.org/browse/SMA-368

Set your device's screen sleep timer to the minimum value (15s on my Pixel w/ Android 12), then try reading any ebook and see if the screen stays on or not after that time has passed. 

No dependencies

No documentation changes

Tested locally on my Pixel5a using 'The Catcher in the Rye'


MalcolmMcFly (approved):
Looks simple enough.

qnga (approved):

PR #984, opened 2022-05-26
OE-525: Re-add Clever to the login flow of Open eBooks

robbyice:
**What's this do?**
Fixes the missing Clever login button from the AccountDetails page for openEbooks. A 'return' was added to the for-loop that dynamically populates the alternative auth views, which bails out of the parent function (not just the for-loop), preventing the app from properly iterating over and potentially creating views for each auth alternative. 

https://jira.nypl.org/browse/OE-525

Open up the Accounts page in OpenEbooks and observe the return of the "Log in with Clever" button

No dependency changes

No documentation changes

Tested locally on my Pixel 5a device


MalcolmMcFly (approved):
The Clever login is back!

https://user-images.githubusercontent.com/15109016/170743976-8571eb5b-f645-4ad1-b155-c6628c08bb54.mp4

PR #990, opened 2022-06-16
SMA-376: Audiobook Player crash

MalcolmMcFly:
**What's this do?**
To resolve the crash, I have updated the `AudioBookPlayer` activity to correctly pass the required parameters to the `PlayerFragment` constructor.

[SMA-376: Audiobook Player crash](https://jira.nypl.org/browse/SMA-376)
There was a spike in audio book player crashes discovered via Crashlytics which prompted this investigation: https://console.firebase.google.com/project/simplye-nypl/crashlytics/app/android:org.nypl.simplified.simplye/issues?time=last-seven-days

Simply play an audiobook and verify it works.

We will need a new fix version in Jira to assign this ticket to.

n/a

@MalcolmMcFly

### DIFF
--- a/simplified-viewer-audiobook/src/main/java/org/nypl/simplified/viewer/audiobook/AudioBookPlayerActivity.kt
+++ b/simplified-viewer-audiobook/src/main/java/org/nypl/simplified/viewer/audiobook/AudioBookPlayerActivity.kt
@@ -43,6 +43,7 @@ import org.librarysimplified.audiobook.views.PlayerFragmentParameters
 import org.librarysimplified.audiobook.views.PlayerPlaybackRateFragment
 import org.librarysimplified.audiobook.views.PlayerSleepTimerFragment
 import org.librarysimplified.audiobook.views.PlayerTOCFragment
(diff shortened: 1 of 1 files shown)


simplified-viewer-audiobook/src/main/java/org/nypl/simplified/viewer/audiobook/AudioBookPlayerActivity.kt
```diff
@@ -408,12 +409,13 @@ class AudioBookPlayerActivity :
       )
       this.supportFragmentManager.fragmentFactory = fragmentFactory
       if (!this.isFinishing && !this.supportFragmentManager.isDestroyed) {
+        val fallbackArgs = bundleOf(Pair(PlayerFragment.parametersKey, null))
```
robbyice: Does this work with a null value in the key? I would think the requireArguments() call in the fragment wouldn't accept it. You could also just pass in PlayerFragmentParameters(null) - the parameters class only takes one nullable argument and isn't used at all in the actual fragment.

If this works though that's fine too :)

robbyice (approved):

PR #994, opened 2022-07-06
OE Login Harmonization

MalcolmMcFly:
**What's this do?**
A copy of `AccountDetailFragment` as `OEAccountDetailFragment` has been made that will show the new login design if `showOELogin` is set in the build configuration service. 

- [x]  [OE-552: Create sign in designs in dark mode for iOS and Android](https://jira.nypl.org/browse/OE-552)
- [x] There is no design for loading screens and errors, now porting old design of this to new login screen.
- [x] [OE-516: Privacy policy page](https://jira.nypl.org/browse/OE-516)
- [x] [OE-515: Terms of Use screen](https://jira.nypl.org/browse/OE-515)
- [x] [OE-517: Redesign the login screen](https://jira.nypl.org/browse/OE-517) - _Logins are currently failing in the new design, now under investigation_
- [x] Clean up all un-needed code in `OEAccountDetailFragment`
- [x] Move hard-coded Gradle dependencies to platforms repo

Launch the Open eBooks app in both light/dark modes and attempt to sign in.

TBD

n/a

TBD


simplified-ui-accounts/src/main/AndroidManifest.xml
```diff
@@ -1,5 +1,13 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="org.nypl.simplified.ui.accounts">
+
     <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
+
+    <application>
+        <activity
+            android:name="org.nypl.simplified.ui.accounts.WebViewActivity"
+            android:exported="false" />
```
robbyice, line 10: What made you go for an activity for the EULA rather than another fragment?

simplified-ui-accounts/build.gradle
```diff
@@ -42,6 +44,9 @@ dependencies {
   implementation libs.rxjava2.extensions
   implementation libs.rxandroid2
   implementation libs.slf4j
+  implementation 'androidx.appcompat:appcompat:1.4.2'
+  implementation 'com.google.android.material:material:1.6.1'
+  implementation 'androidx.constraintlayout:constraintlayout:2.1.4'
```
robbyice: Just a reminder to update build_libraries/platform as part of this PR if you need to

simplified-ui-accounts/build.gradle
```diff
@@ -1,10 +1,12 @@
 apply plugin: 'kotlin-kapt'
 
 apply from: "$rootProject.projectDir/build_test.gradle"
+apply plugin: 'org.jetbrains.kotlin.android'
 
 android {
   buildFeatures {
     dataBinding true
+    viewBinding true
```
robbyice: I believe dataBinding includes anything viewBinding would provide

build_apk.gradle
```diff
@@ -78,6 +78,8 @@ android {
     jvmTarget = "1.8"
   }
   packagingOptions {
+    resources.excludes.add("META-INF/*")
+    resources.excludes.add("**/attach_hotspot_windows.dll")
```
robbyice, line 82: What's going on here?

build.gradle
```diff
@@ -25,6 +25,7 @@ buildscript {
 plugins {
   id "com.github.ben-manes.versions" version "0.39.0"
   id "ru.vyarus.animalsniffer" version "1.5.0"
+    id 'org.jetbrains.kotlin.android' version '1.6.10' apply false
```
robbyice, line 28: What does the 'apply false' do? Never seen that before.

simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/OEAccountDetailFragment.kt
```diff
@@ -0,0 +1,965 @@
...
+    this.firstBookBtn = view.findViewById(R.id.firstbook)
+    this.cleverBtn = view.findViewById(R.id.clever)
+    this.terms = view.findViewById(R.id.terms)
+    this.faq = view.findViewById(R.id.faq)
+    this.loginTrouble = view.findViewById(R.id.loginTrouble)
+    this.firstBookBack = view.findViewById(R.id.firstBookBack)
+    this.firstBookAcccessCode = view.findViewById(R.id.et_access_code)
+    this.firstBookPin = view.findViewById(R.id.et_pin)
+    this.firstBookAcccessCode.filters += InputFilter.AllCaps()
+    this.firstBookSignIn = view.findViewById(R.id.sign_in)
```
robbyice: All these `findViewById` calls can be pretty easily replaced with dataBinding without too much fiddling, would help clean this file up a lot

simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/OEAccountDetailFragment.kt
```diff
@@ -0,0 +1,965 @@
...
+      val request =
+        Basic(
+          accountId = this.viewModel.account.id,
+          description = description,
+          password = accountPassword,
+          username = accountUsername
+        )
+      this.viewModel.tryLogin(request)
+    }
+  }
```
robbyice, line 220: Maybe all of this auth code can be moved inside the ViewModel? (or another component the viewmodel uses)

simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/OEAccountDetailFragment.kt
```diff
@@ -0,0 +1,965 @@
...
+
+    this.logger.debug("broadcasting login state")
+    this.viewModel.account.setLoginState(this.viewModel.account.loginState)
+
+    this.accountIcon.setImageDrawable(null)
+    this.subscriptions.clear()
+  }
+
+  private fun reconfigureAccountUI() {
+    this.authenticationViews.showFor(this.viewModel.account.provider.authentication)
```
robbyice: Does the whole idea of 'reconfiguring' UI even make sense for this OE-specific screen? I would think we wouldn't need this given the screen is only ever going to look the one way - maybe I'm missing something

simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/OEAccountDetailFragment.kt
```diff
@@ -0,0 +1,965 @@
...
+    )
+
+    this.authenticationViews.unlock()
+
+    val loginSatisfied = this.determineLoginIsSatisfied()
+    this.setLoginButtonStatus(loginSatisfied)
+    this.authenticationAlternativesShow()
+  }
+
+  private fun authenticationAlternativesMake() {
```
robbyice: Again I would think we could do away with "auth alternatives" altogether and just have dedicated buttons/flows for this specific screen

simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/WebViewActivity.kt
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * Shows the different outbound account links as WebViews
+ */
+class WebViewActivity : AppCompatActivity() {
+
+  private val logger = LoggerFactory.getLogger(WebViewActivity::class.java)
+
+  companion object {
+    const val PAGE = "page"
```
robbyice, line 19: Might want to use a different value for this 'page' constant since it's intended to be a key for a bundle and not as part of a URL like the other constants. Maybe something like 'PAGE_KEY' or whatnot

simplified-ui-accounts/src/main/res/layout/activity_webview.xml
```diff
@@ -0,0 +1,69 @@
...
+
+    <WebView
+        android:id="@+id/page"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toBottomOf="@id/toolbar" />
+
+    <androidx.constraintlayout.widget.ConstraintLayout
```
robbyice: We could probably unwrap this ProgressBar and move the ID onto the bar itself? Probably not a big deal either way

simplified-ui-accounts/src/main/res/layout/first_book_login.xml
```diff
@@ -0,0 +1,176 @@
...
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:id="@+id/first_book_login"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    android:background="@color/screenBackground"
+    android:visibility="gone"
+    tools:visibility="visible">
+
+    <androidx.constraintlayout.widget.ConstraintLayout
```
robbyice, line 7: I am wondering what the implications are of doing this manually rather than using a customized Action/Toolbar or whatnot.

simplified-ui-accounts/src/main/res/layout/oe_account.xml
```diff
@@ -0,0 +1,320 @@
...
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toBottomOf="@id/accountCustomOPDSText" />
+
+    </androidx.constraintlayout.widget.ConstraintLayout>
+
+  </LinearLayout>
+
+        <include layout="@layout/oe_login_landing" />
+        <include layout="@layout/first_book_login" />
```
robbyice: I think we should be aggressively moving away from the "overlay multiple screens/views on top of each other and toggle them" strategy. The code will be so much easier to work with when it's broken apart into bespoke views/components.

simplified-ui-accounts/src/main/res/values/colors.xml
```diff
@@ -0,0 +1,23 @@
...
+    <color name="textColor">@color/black</color>
+    <color name="linkColor">#0576D3</color>
+    <color name="firstBookBtnPressed">#F5F5F5</color>
+    <color name="firstBookSignIn">#0576D3</color>
+    <color name="firstBookTextColor">@color/firstBookBlue</color>
+    <color name="cleverTextColor">@color/cleverBlue</color>
+    <color name="screenBackground">@color/white</color>
+    <color name="hintTextColor">#BDBDBD</color>
+    <color name="inputLine">#BDBDBD</color>
+    <color name="passwordToggle">#8A8A8A</color>
```
robbyice: I'm wondering about if/how/whether we should be integrating all these new colors into our themes library. Similarly, are there colors we should be using that already exist elsewhere?

simplified-ui-accounts/src/main/res/layout/oe_login_landing.xml
```diff
@@ -0,0 +1,115 @@
...
+            app:layout_constraintEnd_toEndOf="parent"
+            app:layout_constraintStart_toStartOf="parent"
+            app:strokeColor="@color/white" />
+
+        <androidx.constraintlayout.widget.Guideline
+            android:id="@+id/guideline"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:orientation="horizontal"
+            app:layout_constraintGuide_begin="360dp" />
```
robbyice, line 78: Wondering about how this will look on different form factors. Might be worth checking out as part of this PR

simplified-ui-accounts/src/main/java/org/nypl/simplified/ui/accounts/utils/TextViewUtils.kt
```diff
@@ -0,0 +1,41 @@
...
+import android.text.TextPaint
+import android.text.method.LinkMovementMethod
+import android.text.style.ClickableSpan
+import android.view.View
+import android.widget.TextView
+
+/**
+ * Utility to provide the ability to have multiple clickable links in a TextView
+ */
+fun TextView.makeLinks(vararg links: Pair<String, View.OnClickListener>) {
```
robbyice, line 16: It'd be nice to have tests for this, this looks important/decently complex

robbyice:
Tests around the OEAccountDetailFragment are probably a good idea

simplified-ui-accounts/build.gradle
MalcolmMcFly: Yes this was on my to-do list above.
Done.

simplified-ui-accounts/build.gradle
MalcolmMcFly: Didn't know that.

simplified-ui-accounts/src/main/res/layout/oe_account.xml
MalcolmMcFly: Yes, I really hate this, and it is only a stepping stone until we have the time budget to re-write the accounts/settings logic.

PR #998, opened 2022-07-26
OE-605: Login Screens Not Showing Properly on Smaller Screens

MalcolmMcFly:
**What's this do?**
Set button text to show ellipses for longer text and better re-aligned toolbar image.

[OE-605: Login Screens Not Showing Properly on Smaller Screens](https://jira.nypl.org/browse/OE-605)

Run the app on a very small device and examine UI/UX for visibility. 

n/a

n/a

@MalcolmMcFly 

Screenshot_20220726_123639

### DIFF
--- a/simplified-ui-accounts/src/main/res/layout/activity_webview.xml
+++ b/simplified-ui-accounts/src/main/res/layout/activity_webview.xml
@@ -19,6 +19,16 @@
             app:layout_constraintTop_toTopOf="parent">
 
             <ImageView
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:contentDescription="@string/contentDescOpenEbooksLogo"
+                android:background="@drawable/ic_oe_top_bar"
+                app:layout_constraintBottom_toBottomOf="@id/toolbar"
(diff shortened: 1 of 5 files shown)


ettore (commented):
i can’t really comment on the actual changes here since i don’t know much of android, but are they really ok with showing ellipses?

ettore:
> i can’t really comment on the actual changes here since i don’t know much of android, but are they really ok with showing ellipses?

For example, one option would be to tighten the padding on the left, or reduce the font size on smaller screens.

MalcolmMcFly:
> i can’t really comment on the actual changes here since i don’t know much of android, but are they really ok with showing ellipses?

Good call-out @ettore, I am going to see if I can modify that button text size specifically for smaller screens.

ettore:
> > i can’t really comment on the actual changes here since i don’t know much of android, but are they really ok with showing ellipses?
> 
> Good call-out @ettore, I am going to see if I can modify that button text size specifically for smaller screens.

looks a lot better! and it seems a lot easier than iOS

ettore (approved):

ErnestFan (approved):

PR #1005, opened 2022-09-21
SMA-398: Minor Test Upgrades

robbyice:
**What's this do?**
This PR does a few minor upgrades to help improve our Robolectric tests.

Such as...

- Bump Gradle and AGP version
- Upgrade Robolectric to latest version
- Upgrade AndroidX Fragment to 1.4+ for Closeable FragmentScenarios
- Invoke close() for scenarios in After block to help clean up smoothly

These changes are a result of investigating some seemingly slow Robolectric tests. It turns out the particular tests weren't actually slow, but were instead just reporting the overhead cost of starting up the test suite. That overhead cost isn't great but it is expected and seems generally reasonable for Robolectric. 

https://jira.nypl.org/browse/SMA-398

There are no new added tests but existing tests can be run with these changes.

Platform dependencies have already been merged.

No documentation changes needed.

Ran tests locally on my development machine.

### DIFF
--- a/simplified-cardcreator/src/test/kotlin/org/nypl/simplified/cardcreator/ui/ConfirmationFragmentTest.kt
+++ b/simplified-cardcreator/src/test/kotlin/org/nypl/simplified/cardcreator/ui/ConfirmationFragmentTest.kt
@@ -26,8 +26,10 @@ import io.mockk.justRun
 import io.mockk.mockkConstructor
 import io.mockk.verify
 import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.update
+import kotlinx.coroutines.runBlocking
 import kotlinx.coroutines.test.runBlockingTest
 import org.amshove.kluent.fail
 import org.amshove.kluent.shouldBe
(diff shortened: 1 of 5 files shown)


simplified-cardcreator/src/test/kotlin/org/nypl/simplified/cardcreator/ui/ConfirmationFragmentTest.kt
```diff
@@ -90,6 +93,31 @@ class ConfirmationFragmentTest {
...
 
+  @After
+  fun tearDown() {
+    /*
+    * There are some issues with activityscenario state when an Activity's finish() is called
+    * at certain lifecycle points so we only call close on non-finishing activities to avoid this.
+    * Not closing the scenarios invoking finish does introduce potential waste, but as of right now
+    * there is only this one test involving a finish() call.
+    *
+    * Reportedly this issue might be fixed by upgrading AndroidX Test Core to 1.5+
```
MalcolmMcFly, line 104: Would upgrading upgrading AndroidX Test Core to 1.5+ be alot of overhead?

MalcolmMcFly (approved):
I see alot of `relaxed = true` and `scenario.close()`. Any idea of how much time this ended up saving us?

simplified-cardcreator/src/test/kotlin/org/nypl/simplified/cardcreator/ui/ConfirmationFragmentTest.kt
robbyice: I don't think the `relaxed=true` or `scenario.close()` saved us much time if any. They're moreso housekeeping additions. Ultimately I wasn't able to figure out any obvious way to make things measurably quicker.

The `relaxed=true` was to address some random exceptions that were being thrown for certain tests where methods that weren't stubbed were being called indirectly by the code under test. 

The `scenario.close()` is actually a change that could result in a performance increase, but we just dont have that many scenario tests yet. I think you'd see a performance improvement with a large suite of tests (like hundreds).  For us the improvement is likely very minimal and this more just to have proper tests that clean up after themselves.

As for AndroidX Test Core 1.5+ I'm not so sure, I haven't looked into if there's any significant or breaking changes but I imagine it would be an easy upgrade. Currently however the 1.5+ builds are in an early alpha stage, so I figured we could wait a bit and just use this janky fix for the time being given it's only for a single test.

PR #1006, opened 2022-09-27
SMA-395: Spike - Investigate a Github Actions-Only CI/CD pipeline

MalcolmMcFly:
**What's this do?**

- [x] Adds new manual trigger for distributing builds to Firebase.
- [x] Separates build process for Open eBooks and SimplyE

[SMA-395: Investigate a Github Actions-Only CI/CD pipeline](https://jira.nypl.org/browse/SMA-395)

Trigger manual build for Github.

n/a

n/a

### DIFF
--- a/.github/workflows/simplye.yml
+++ b/.github/workflows/simplye.yml
@@ -61,16 +61,14 @@ jobs:
           java-version: 11
 
       - name: Run tests
-        run: ./gradlew clean testDebug
+        run: ./gradlew clean testRelease
 
   distribute:
     needs: [ unit_tests ]
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout code
-        uses: actions/checkout@v2
-
-      - uses: actions/download-artifact@main
+      - name: Download APK
+        uses: actions/download-artifact@main
         with:
           name: simplye-release.apk
 
--- a/.github/workflows/simplye.yml
+++ b/.github/workflows/simplye.yml
@@ -1,0 +1,84 @@
+name: SimplyE Build, Test, & Distribute
+
+on: workflow_dispatch
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Reposistory
+        uses: actions/checkout@v3
+        with:
+          submodules: 'recursive'
+
+      - name: Set up JDK 11
+        uses: actions/setup-java@v1
+        with:
+          distribution: 'temurin'
+          java-version: 11
+
+      - name: Set Up Credentials
+        env:
+          NYPL_S3_ACCESS_KEY: ${{ secrets.NYPL_S3_ACCESS_KEY }}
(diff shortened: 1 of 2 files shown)


.github/workflows/simplye.yml
```diff
@@ -0,0 +1,84 @@
...
+
+      - name: Run tests
+        run: ./gradlew clean testDebug
+
+  distribute:
+    needs: [ unit_tests ]
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
```
robbyice: Do we need to checkout here? If so we should use v3 w/ recursion.

.github/workflows/simplye.yml
```diff
@@ -0,0 +1,84 @@
...
+          submodules: 'recursive'
+
+      - name: Set up JDK 11
+        uses: actions/setup-java@v1
+        with:
+          distribution: 'temurin'
+          java-version: 11
+
+      - name: Run tests
+        run: ./gradlew clean testDebug
```
robbyice: I would probably do testRelease here. We don't really have any differences in debug vs release flavors but it is more proper to test the flavor we end up distributing.

.github/workflows/simplye.yml
```diff
@@ -0,0 +1,84 @@
...
+      - name: Build
+        env:
+          NYPL_S3_ACCESS_KEY:               ${{ secrets.NYPL_S3_ACCESS_KEY }}
+          NYPL_S3_SECRET_ACCESS_KEY:        ${{ secrets.NYPL_S3_SECRET_ACCESS_KEY }}
+          MAVEN_CENTRAL_USERNAME:           ${{ secrets.MAVEN_CENTRAL_USERNAME }}
+          MAVEN_CENTRAL_PASSWORD:           ${{ secrets.MAVEN_CENTRAL_PASSWORD }}
+          MAVEN_CENTRAL_STAGING_PROFILE_ID: 'af061f5afba777'
+          MAVEN_CENTRAL_SIGNING_KEY_ID:     'Library Simplified'
+          NYPL_GITHUB_ACCESS_TOKEN:         ${{ secrets.NYPL_GITHUB_ACCESS_TOKEN }}
+        run: .ci/ci-build.sh normal
```
robbyice, line 40: As part of this PR I would update this ci-build.sh script in the CI module to not do all the things it's doing right now.

Currently its doing `assemble test verifySemanticVersioning` - it could probably just do `assembleRelease` which will cut the runtime down significantly

MalcolmMcFly: Good idea!

robbyice: Oh also you can probably adjust the parameters for the Gradle invocation to be similar to the one for pull-requests - i.e. `-Dorg.gradle.parallel=true`. Though eventually I plan on centralizing that setup

robbyice (approved):

PR #1010, opened 2023-08-22
Update target sdk and error screen

russellcullen:
**What's this do?**
Update target and compile sdk to prep for Google policy changes. In addition, this also updates the error reporting screen since the current "Report Issue" button sends the report to a generic, untracked email, we're replacing that button with some text to direct the user to their local library help email.

### DIFF
--- a/simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/ui/LocationFragment.kt
+++ b/simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/ui/LocationFragment.kt
@@ -107,7 +107,7 @@ class LocationFragment : Fragment(), LocationListener {
       if (isNewYork || locationMock != null) {
         locationMock?.let { location ->
           activityModel.userLocationAddress =
-            geocoder.getFromLocation(location.latitude, location.longitude, 1)[0]
+            geocoder.getFromLocation(location.latitude, location.longitude, 1)?.get(0)
         }
 
         logger.debug("User navigated to the next screen")
@@ -230,7 +230,7 @@ class LocationFragment : Fragment(), LocationListener {
 
     try {
       if (location != null) {
-        address = geocoder.getFromLocation(location.latitude, location.longitude, maxResults)[0]
+        address = geocoder.getFromLocation(location.latitude, location.longitude, maxResults)!![0]
         logger.debug("Region is: ${address.adminArea} ${address.countryCode} ")
         binding.regionEt.setText("${address.adminArea} ${address.countryCode}", TextView.BufferType.EDITABLE)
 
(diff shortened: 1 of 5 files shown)


samanthaandrews (approved):
Looks good to me, but an Android dev should probably give final approval 👍

samanthaandrews (approved):

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/ui/LocationFragment.kt
```diff
@@ -292,7 +292,7 @@ class LocationFragment : Fragment(), LocationListener {
     val address: Address?
 
     try {
-      address = geocoder.getFromLocation(location.latitude, location.longitude, maxResults)[0]
+      address = geocoder.getFromLocation(location.latitude, location.longitude, maxResults)!![0]
```
russellcullen, line 295: both these force unwraps are contained in generic `try {} catch (e: Exception) {}` blocks so it's 'safe' to do. (as safe as if we had a null address here, the exception is caught either way)

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/ui/LocationFragment.kt
```diff
@@ -230,7 +230,7 @@ class LocationFragment : Fragment(), LocationListener {
 
     try {
       if (location != null) {
-        address = geocoder.getFromLocation(location.latitude, location.longitude, maxResults)[0]
+        address = geocoder.getFromLocation(location.latitude, location.longitude, maxResults)!![0]
```
MalcolmMcFly, line 233: Might be more readable like `address = geocoder.getFromLocation(location.latitude, location.longitude, maxResults).first()`

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/ui/LocationFragment.kt
MalcolmMcFly, line 295: Same as above, might be more readable like `address = geocoder.getFromLocation(location.latitude, location.longitude, maxResults).first()`

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/ui/LocationFragment.kt
```diff
@@ -107,7 +107,7 @@ class LocationFragment : Fragment(), LocationListener {
       if (isNewYork || locationMock != null) {
         locationMock?.let { location ->
           activityModel.userLocationAddress =
-            geocoder.getFromLocation(location.latitude, location.longitude, 1)[0]
+            geocoder.getFromLocation(location.latitude, location.longitude, 1)?.get(0)
```
MalcolmMcFly, line 110: Might be more readable like: `geocoder.getFromLocation(location.latitude, location.longitude, 1).first()`

MalcolmMcFly (approved):
Very minor non-blocking comments. This code probably won't be supported long term so take my suggestions here with a grain of salt.

simplified-cardcreator/src/main/java/org/nypl/simplified/cardcreator/ui/LocationFragment.kt
MalcolmMcFly: No biggie, just aiming for readability. Good to merge 👏

PR #1011, opened 2023-11-14
Add expiration for open access books

russellcullen:
**What's this do?**
Adds an optional expiration date from the `until` field in the OPDS availability feed for Open Access publications. 
This change also updates the `BookSyncTask` to check for expired open access books and remove them from the device, as there is no loan feed to sync with. 

Yup!

### DIFF
--- a/simplified-books-controller/src/main/java/org/nypl/simplified/books/controller/BookSyncTask.kt
+++ b/simplified-books-controller/src/main/java/org/nypl/simplified/books/controller/BookSyncTask.kt
@@ -1,6 +1,7 @@
 package org.nypl.simplified.books.controller
 
 import com.io7m.jfunctional.Some
+import org.joda.time.DateTime
 import org.librarysimplified.http.api.LSHTTPClientType
 import org.librarysimplified.http.api.LSHTTPResponseStatus
 import org.nypl.simplified.accounts.api.AccountAuthenticationCredentials
@@ -23,6 +24,7 @@ import org.nypl.simplified.feeds.api.FeedLoading
 import org.nypl.simplified.opds.core.OPDSAvailabilityRevoked
 import org.nypl.simplified.opds.core.OPDSFeedParserType
 import org.nypl.simplified.opds.core.OPDSParseException
+import org.nypl.simplified.opds.core.getOrNull
 import org.nypl.simplified.patron.api.PatronUserProfile
 import org.nypl.simplified.patron.api.PatronUserProfileParsersType
 import org.nypl.simplified.profiles.api.ProfileID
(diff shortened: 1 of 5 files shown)


simplified-opds-core/src/main/java/org/nypl/simplified/opds/core/OPDSAcquisitionFeedEntryParser.java
```diff
@@ -164,7 +165,16 @@ private OPDSAcquisitionFeedEntry parseAcquisitionEntry(
...
+
+    String distribution = findDistribution(element);
+    entry_builder.setDistribution(distribution);
+    if ("Axis 360".equals(distribution)) {
+      DateTime end = DateTime.now().plusMonths(2);
+      entry_builder.setAvailability(
+        OPDSAvailabilityOpenAccess.get(revoke, Option.some(end))
+      );
+    }
+
```
russellcullen: This is the code that can be removed if we get the date in the OPDS feed

simplified-books-controller/src/main/java/org/nypl/simplified/books/controller/BookSyncTask.kt
```diff
@@ -64,12 +66,14 @@ class BookSyncTask(
...
     if (providerAuth == AccountProviderAuthenticationDescription.Anonymous) {
       this.logger.debug("account does not support syncing")
+      this.removeExpiredBooks(account)
       return this.taskRecorder.finishSuccess(Unit)
     }
 
     val credentials = account.loginState.credentials
     if (credentials == null) {
       this.logger.debug("no credentials, aborting!")
+      this.removeExpiredBooks(account)
```
russellcullen, line 76: This case and the one above at L69 are the branches for no-auth accounts that do not support remote syncing. We check for expired books on these accounts only.

simplified-opds-core/src/main/java/org/nypl/simplified/opds/core/OPDSAcquisitionFeedEntryParser.java
```diff
@@ -164,7 +165,16 @@ private OPDSAcquisitionFeedEntry parseAcquisitionEntry(
 
     findAcquisitionAuthors(element, entry_builder);
     entry_builder.setPublisherOption(findPublisher(element));
-    entry_builder.setDistribution(findDistribution(element));
+
+    String distribution = findDistribution(element);
+    entry_builder.setDistribution(distribution);
+    if ("Axis 360".equals(distribution)) {
```
ExplorerNautilus: Would checking the [acquisition type](https://github.com/NYPL-Simplified/Simplified-iOS/blob/develop/Simplified/OPDS/NYPLOPDSAcquisitionPath.m#L18) better than the distributor name? The acquisition type seems to be more reliable than the distributor string literal

russellcullen: Yeah thats a good call, we should be able to use that instead!

simplified-opds-core/src/main/java/org/nypl/simplified/opds/core/OPDSAcquisitionFeedEntryParser.java
russellcullen: Removed

ExplorerNautilus (approved):

